PR #4, opened 2022-01-14
Initial commit

vladsaveliev:
Implementation of https://github.com/populationgenomics/production-pipelines/tree/design-docs proposal.

* Add `Pipeline` class along with `@stage` decorator to create Hail Batch base pipelines
* Add functions that create typical bioinformatics jobs: alignment, variant calling, pedigree checks, etc.
* Add a specific implementation of a pipeline: a `SeqrLoader` pipeline.
* Add tests and CI.


lgruen (changes requested):
Please let me know when this is ready for another look!

lgruen (changes requested):
Is the plan still to split this into smaller chunks for review?

PR #13, opened 2022-02-08
VEP job and batch

vladsaveliev:
* Add a Hail Batch job to run VEP: `cpg_pipes/jobs/vep.py`
* Add example batch: `batches/vep.py`
* Add basic `cpg_pipes` utilities that help running that VEP batch:
  * `from cpg_pipes.hb.batch import setup_batch`
  * `from cpg_pipes.hb.command import wrap_command`
  * `from cpg_pipes import images` with CPG AR images,
  * `from cpg_pipes import ref_data` with file pointers on the CPG reference bucket,
  * `from cpg_pipes import buckets` with `file_exists` and `can_reuse` helper functions that check the object existence on a bucket.

Will add more in following PRs.


cpg_pipes/images.py
```diff
@@ -0,0 +1,5 @@
+"""
+Docker images used in the pipelines.
+"""
+AR_REPO = 'australia-southeast1-docker.pkg.dev/cpg-common/images'
```
MattWellie, line 4: Would common image reference location replace the alternative cpg_pipes.resources in [this branch](https://github.com/populationgenomics/production-pipelines/blob/initial_with_vep/cpg_pipes/resources.py)? If so, can we add a bunch more of the default container names in this MR?

cpg_pipes/buckets.py
```diff
@@ -0,0 +1,69 @@
+"""
+Utility functions to interact with objects on buckets.
```
lgruen, line 2: Should these be part of `cpg_utils`?

cpg_pipes/hb/batch.py
```diff
@@ -0,0 +1,81 @@
+"""
+Extending the Hail's `Batch` class.
```
lgruen, line 2: That sounds like you're deriving a class, but this is more like a set of utility functions for creating new batches, right?

setup.py
```diff
@@ -0,0 +1,47 @@
...
+import os
+from os.path import join
+import setuptools
+
+
+setuptools.setup(
+    name='cpg-pipes',
+    version='0.2.9',
+    description=(
+        'Hail Batch pipelines for large cohort and rare deseases projects'
```
lgruen, line 12:
```suggestion
        'Hail Batch pipelines for large cohort and rare disease projects'  
```

batches/vep.py
```diff
@@ -0,0 +1,40 @@
...
+    '--vcf',
+    '--vcf-path',
+    'vcf_paths',
+    multiple=True,
+    help='Paths to VCFs (can be multiple)',
+)
+@click.option(
+    '--project',
+    'project',
+    help='Billing project for Batch'
```
lgruen, line 23: Don't we typically set that through an environment variable, e.g. in the analysis-runner?

lgruen (changes requested):

cpg_pipes/hb/command.py
```diff
@@ -0,0 +1,59 @@
...
+    Wraps a command for submission
+    If job_resource is defined, monitors output space.
+    If output_bucket_path_to_check is defined, checks if this file(s) exists,
+    and if it does, skips running the rest of the job.
+    
+    :param command: command to wrap
+    :param monitor_space: add a background process that checks the instance disk 
+        space every 5 minutes and prints it to the screen
+    :param setup_gcp: login to GCP
+    :param dedent: remove all common leading intendation from the command
```
MattWellie, line 39:
```suggestion
    :param dedent: remove all common leading indentation from the command
```

cpg_pipes/images.py
vladsaveliev: That's right, it's the same common image reference location!

What is `MR`? 

We can definitely add more images into the `cpg-common/images` repositories, and list them in the `cpg_pipes/images.py` if they are used in the pipeline (e.g. I will be adding more images while I'm moving more implemented jobs from the `initial` branch).

cpg_pipes/hb/batch.py
vladsaveliev: I dropped the class from this initial PR for simplicity.

In the `initial` branch, I derived the `Batch` class so I can override `b.new_job()` and count jobs and job types before submission, as a good sanity check for pipelines: https://github.com/populationgenomics/production-pipelines/blob/initial/cpg_pipes/hb/batch.py
It's not a critical feature, and in this PR I wanted to keep only the basic things that would make the VEP example running.

batches/vep.py
vladsaveliev: It's an optional parameter, and defaults to the env variable if it's defined:
```
billing_project = billing_project or os.getenv('HAIL_BILLING_PROJECT')
```
Removed the parameter now, reading the env var directly!

Just wanted to avoid it depending on our infra if we want the code to be reused outside of CPG, but could think about it down the road.

cpg_pipes/buckets.py
vladsaveliev: It probably shouldn't. 
`can_reuse` is something relevant for pipeline-like code - it checks whether stage's outputs can be reused to avoid unnecessary computation.
`exists` checks the object existence, but it also has special cases for `.ht` and `.mt` paths, in which case it would check for the existence of the `_SUCCESS` object. This hail-related special case might not be good for a more generic cpg_utils package.

cpg_pipes/hb/command.py
vladsaveliev: Thank you!!

cpg_pipes/images.py
MattWellie: Sorry, MR = merge request = pull request, I think that's just a terminology difference between gitlab and github

vladsaveliev: Matt, answering your question:
> If so, can we add a bunch more of the default container names in this MR?
I wanted to try to stick to making small PRs, so I removed everything but one image from this module. I'm happy to add more images in a following PR!

README.md
```diff
@@ -1 +1,25 @@
...
+  * `cpg_pipes/hb`: interact with Hail Batch and build new jobs,
+  * `cpg_pipes/buckets`: interact with objects on buckets (GCP, and Azure in the future),
+  * `cpg_pipes/images`: Docker image tags hosted by the CPG,
+  * `cpg_pipes/ref_data`: bioinformatics reference data hosted by the CPG,
+* `batches`: example Batches that use jobs from `cpg_pipes/jobs`
+
+Batches must be run with the [analysis-runner](https://github.com/populationgenomics/analysis-runner):
+
+```bash
+# make sure you've installed https://anaconda.org/cpg/analysis-runner
```
lgruen, line 16: This is now in pypi, right?

README.md
```diff
@@ -1 +1,25 @@
...
+
+Helper functions to run Hail Batch pipelines at the CPG.
+
+* `cpg_pipes` contains the codebase for the `cpg-pipes` pip package, which includes multiple utility functions:
+  * `cpg_pipes/jobs`: create Job objects for various bioinformatics purposes,
+  * `cpg_pipes/hb`: interact with Hail Batch and build new jobs,
+  * `cpg_pipes/buckets`: interact with objects on buckets (GCP, and Azure in the future),
+  * `cpg_pipes/images`: Docker image tags hosted by the CPG,
+  * `cpg_pipes/ref_data`: bioinformatics reference data hosted by the CPG,
+* `batches`: example Batches that use jobs from `cpg_pipes/jobs`
```
lgruen, line 11: If this is just meant as an example, should the directory be called `examples`?

cpg_pipes/files.py
```diff
@@ -0,0 +1,70 @@
...
+      in which case it will check for the existence of a
+      *.mt/_SUCCESS or *.ht/_SUCCESS file
+    For cloud objects, the function would log each object access, which is useful
+    to detect unnecessary accesses which can be a bottleneck when working with a lot
+    of files. The `silent` flag can be set to True to disable logging.
+    """
+    if path.startswith('gs://'):
+        bucket = path.replace('gs://', '').split('/')[0]
+        p = path.replace('gs://', '').split('/', maxsplit=1)[1]
+        p = p.rstrip('/')  # ".mt/" -> ".mt"
```
lgruen, line 31:
```suggestion
bucket, p = path.removeprefix('gs://').split('/', maxsplit=2)[:2]
```

lgruen (approved):

MattWellie (approved):

PR #26, opened 2022-03-30
Update get ped

vivbak:


cpg_pipes/providers/inputs.py
```diff
@@ -127,12 +131,21 @@ def populate_analysis(self, cohort: Cohort) -> None:
...
         cohort: Cohort,
         ped_files: list[Path] | None = None,
     ) -> None:
         """
-        Populate pedigree data
+        Populate pedigree data from file
+        """
+
+    @abstractmethod
+    def populate_pedigree_from_db(
```
vladsaveliev, line 143: The abstract class shouldn't know about the DB. DB is a detail of the SMDB implementation, and in contrast, the CSV implementation wouldn't know what to do in the `populate_pedigree_from_db` method vs. standard `populate_pedigree`.

I think there should be only one method called `populate_pedigree`.

cpg_pipes/providers/cpg/smdb.py
```diff
@@ -343,16 +350,64 @@ def process_existing_analysis(
...
+                    family['individual_id'] = pid_sid_map[family['individual_id']]
+
+                if family['maternal_id'] != '':
+                    family['maternal_id'] = pid_sid_map[family['maternal_id']]
+
+                if family['paternal_id'] != '':
+                    family['paternal_id'] = pid_sid_map[family['paternal_id']]
+
+            return pedigree
+        except ApiException:
```
vladsaveliev, line 382: Should we fail silently? In what scenarios do we expect SMDB to not work properly?

vladsaveliev (changes requested):
Thank you!

Left two comments about the abstract `populate_pedigree_from_db` and the broad exception, otherwise looks good

cpg_pipes/providers/inputs.py
vivbak: Good point! I've combined the two now.

cpg_pipes/providers/cpg/smdb.py
vivbak: I've thought a bit about this and can't think of a specific case where this would fail that we could anticipate/catch. For example, dataset being invalid would be caught far earlier. And if there aren't corresponding families or pedigree information, then that doesn't constitute a failure (it should populate it with nothing, which it does). 
I've removed the try-catch for now. Do you think there is a better solution here?

vladsaveliev: Removing it is good, I think. If we hit an error, we should look into fixing sample-metadata, instead of failing silently

cpg_pipes/providers/inputs.py
vladsaveliev: Thanks! We might even want to move the ped parsing block into the abstract class, because I imagine it should be able to parse them - if provided - for any provider implementation.

vladsaveliev (approved):

PR #27, opened 2022-04-06
Fix ped

vivbak:
Removes redundant mapping to participant in SMDB. 
Handles scenario where participants exist, but corresponding samples do not.


cpg_pipes/providers/cpg/inputs.py
```diff
@@ -106,22 +106,26 @@ def populate_pedigree(
...
-        sample_by_internal_id = dict()
-        for s in cohort.get_samples():
-            sample_by_internal_id[s.id] = s
-
         for dataset in cohort.get_datasets():
             ped_entries = self.db.get_ped_entries(dataset_name=dataset.name)
             for entry in ped_entries:
-                sam_id = entry['individual_id']
-                if sam_id not in sample_by_internal_id:
+                part_id = str(entry['individual_id'])
```
vladsaveliev, line 112: Since you set `replace_with_participant_external_ids=False`, does it mean that `individual_id` would be an internal ID, no a participant ID?

vivbak: It means that `individual_id` is the _internal_ participant ID as opposed to the _external_ participant id. It was previously a sample ID, because in SMDB I used to convert the participant IDs to sample IDs.

vladsaveliev: But wouldn't `replace_with_participant_external_ids=False` mean that `individual_id` here would be an internal ID? How does the flag `replace_with_participant_external_ids` work?

vivbak: Yup! It would be an internal participant id (as opposed to an internal sample ID). PED files are created referencing participant external IDs by default (value is by default true). When you set `replace_with_participant_external_ids=False`, then instead of returning external participant IDs, it returns internal participant IDs. I switched it to False, because the participant ID that is already already populated for each sample (`s.participant_id`), in cpg_pipes, is the internal participant ID.

vladsaveliev: Ah, I didn't realise the difference between internal IDs and internal _participant_ IDs. So many IDs. 
Okay, got it now, I'll merge.

PR #29, opened 2022-04-09
Input provider and status reporter

vladsaveliev:
* Abstract class `InputProvider` to read pipeline inputs (samples, datasets, sequencing data, pedigree etc). Two implementations of `InputProvider`: sample-metadata and CSV.
* Abstract class `StatusReporter` to report analysis status, and its `sample-metadata` implementation.

### DIFF
--- a/cpg_pipes/providers/__init__.py
+++ b/cpg_pipes/providers/__init__.py
@@ -6,7 +6,7 @@ from enum import Enum
 
 
 class InputProviderType(Enum):
-    """Available types of input prpviders"""
+    """Available types of input providers"""
 
     SMDB = 'smdb'
     CSV = 'csv'
--- a/cpg_pipes/providers/__init__.py
+++ b/cpg_pipes/providers/__init__.py
@@ -1,0 +1,18 @@
+"""
+Providers that can be overriden in specific infrastructutres.
+"""
+
+from enum import Enum
+
+
+class InputProviderType(Enum):
+    """Available types of input prpviders"""
+
+    SMDB = 'smdb'
+    CSV = 'csv'
+
+
+class StatusReporterType(Enum):
+    """Available types of status reporters"""
+
+    SMDB = 'smdb'
--- a/cpg_pipes/providers/__init__.py
+++ b/cpg_pipes/providers/__init__.py
@@ -1,5 +1,5 @@
 """
-Providers that can be overriden in specific infrastructutres.
+Providers that can be overridden in specific infrastructures.
 """
 
 from enum import Enum
--- a/cpg_pipes/providers/__init__.py
+++ b/cpg_pipes/providers/__init__.py
@@ -15,4 +15,4 @@ class InputProviderType(Enum):
 class StatusReporterType(Enum):
     """Available types of status reporters"""
 
-    SMDB = 'smdb'
+    CPG = 'cpg'
--- a/cpg_pipes/providers/cpg/inputs.py
+++ b/cpg_pipes/providers/cpg/inputs.py
@@ -32,7 +32,7 @@ class SmdbInputProvider(InputProvider):
         ped_files: list[Path] | None = None,
     ) -> Cohort:
         """
-        Overwriding the superclass method.
+        Overriding the superclass method.
         """
         if not dataset_names:
             raise InputProviderError(
--- a/cpg_pipes/providers/cpg/inputs.py
+++ b/cpg_pipes/providers/cpg/inputs.py
@@ -1,0 +1,190 @@
+"""
+InputProvider implementation that pulls data from the sample-metadata database.
+"""
+
+import logging
+
+from .smdb import SMDB, SmSequence
+from ..inputs import InputProvider, InputProviderError
+from ... import Path
+from ...targets import Cohort, Sex, PedigreeInfo, Dataset
+from ...types import SequencingType
+
+logger = logging.getLogger(__file__)
+
+
+class SmdbInputProvider(InputProvider):
+    """
+    InputProvider implementation that pulls data from the sample-metadata database.
+    """
+
+    def __init__(self, db: SMDB, check_files: bool = False):
+        super().__init__(check_files=check_files)
+        self.db = db
+
+    def populate_cohort(
+        self,
+        cohort: Cohort,
+        dataset_names: list[str] | None = None,
+        skip_samples: list[str] | None = None,
+        only_samples: list[str] | None = None,
+        skip_datasets: list[str] | None = None,
+        ped_files: list[Path] | None = None,
+    ) -> Cohort:
+        """
+        Overwriding the superclass method.
+        """
+        if not dataset_names:
+            raise InputProviderError(
+                'Dataset must be provided for SmdbInputProvider.populate_cohort()'
+            )
+        return super().populate_cohort(
+            cohort=cohort,
+            dataset_names=dataset_names,
+            skip_samples=skip_samples,
+            only_samples=only_samples,
+            skip_datasets=skip_datasets,
+            ped_files=ped_files,
+        )
+
+    def get_entries(
+        self,
+        dataset: Dataset | None = None,
+    ) -> list[dict]:
+        """
+        Return list of data entries.
+        """
+        if dataset is None:
+            raise InputProviderError(
+                'SmdbInputProvider: dataset must be provided for get_entries()'
+            )
+        entries = self.db.get_sample_entries(project_name=dataset.name)
+        # Adding "dataset" into entries, needed for `self.get_dataset_name()`:
+        for e in entries:
+            e['dataset'] = dataset.name
+        return entries
+
+    def get_dataset_name(self, entry: dict) -> str:
+        """
+        Get name of the dataset. Not relevant for SMDB because we pull
+        specific datasets by their names.
+        """
+        return entry['dataset']
+
+    def get_sample_id(self, entry: dict) -> str:
+        """
+        Get sample ID from a sample dict.
+        """
+        return entry['id'].strip()
+
+    def get_external_id(self, entry: dict) -> str | None:
+        """
+        Get external sample ID from a sample dict.
+        """
+        return entry['external_id'].strip()
+
+    def get_participant_id(self, entry: dict) -> str | None:
+        """
+        Get participant ID from a sample dict.
+        """
+        return None  # Unknown before all samples are loaded
+
+    def get_participant_sex(self, entry: dict) -> Sex | None:
+        """
+        Get participant ID from a sample dict.
+        """
+        return Sex.parse(entry['sex']) if 'sex' in entry else None
+
+    def get_sample_meta(self, entry: dict) -> dict:
+        """
+        Get sample metadata..
+        """
+        return entry.get('meta', {})
+
+    def get_sequencing_type(self, entry: dict) -> SequencingType | None:
+        """
+        Get sequencing type.
+        """
+        return None  # Unknown before all samples are loaded
+
+    def populate_alignment_inputs(self, cohort: Cohort) -> None:
+        """
+        Populate sequencing inputs for samples.
+        """
+        seq_infos: list[dict] = self.db.seqapi.get_sequences_by_sample_ids(
+            cohort.get_sample_ids()
+        )
+        seq_by_sid = {seq['sample_id']: seq for seq in seq_infos}
+        for sample in cohort.get_samples():
+            seq_info = seq_by_sid[sample.id]
+            seq = SmSequence.parse(seq_info, self.check_files)
+            sample.alignment_input = seq.alignment_input
+            sample.sequencing_type = seq.sequencing_type
+
+    def populate_analysis(self, cohort: Cohort) -> None:
+        """
+        Populate Analysis entries.
+        """
+        pass
+
+    def populate_participants(self, cohort: Cohort) -> None:
+        """
+        Populate Participant entries.
+        """
+        for dataset in cohort.get_datasets():
+            pid_sid_multi = (
+                self.db.papi.get_external_participant_id_to_internal_sample_id(
+                    dataset.name
+                )
+            )
+            participant_by_sid = {}
+            for group in pid_sid_multi:
+                pid = group[0]
+                for sid in group[1:]:
+                    participant_by_sid[sid] = pid.strip()
+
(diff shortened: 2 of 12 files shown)


cpg_pipes/providers/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""
+Providers that can be overriden in specific infrastructutres.
```
lgruen, line 2:
```suggestion
Providers that can be overridden in specific infrastructures.
```

cpg_pipes/providers/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+from enum import Enum
+
+
+class InputProviderType(Enum):
+    """Available types of input prpviders"""
```
lgruen, line 9:
```suggestion
    """Available types of input providers"""
```

cpg_pipes/providers/cpg/inputs.py
```diff
@@ -0,0 +1,190 @@
...
+        self,
+        cohort: Cohort,
+        dataset_names: list[str] | None = None,
+        skip_samples: list[str] | None = None,
+        only_samples: list[str] | None = None,
+        skip_datasets: list[str] | None = None,
+        ped_files: list[Path] | None = None,
+    ) -> Cohort:
+        """
+        Overwriding the superclass method.
```
lgruen, line 35:
```suggestion
        Overriding the superclass method.
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+from ..types import FastqPair, CramPath, AlignmentInput
+from ..targets import Cohort, Dataset, Sex, SequencingType, PedigreeInfo
+from .. import Path
+
+logger = logging.getLogger(__file__)
+
+
+class InputProviderError(Exception):
+    """
+    Exception thrown when there is something wrong happened parsing
```
lgruen, line 20:
```suggestion
    Exception thrown when there is something wrong while parsing
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
+"""
+Abstract provider if input data source.
```
lgruen, line 2:
```suggestion
Abstract provider of input data sources.
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+    Exception thrown when there is something wrong happened parsing
+    inputs.
+    """
+
+    pass
+
+
+class InputProvider(ABC):
+    """
+    Abstract class for implementing inputs source.
```
lgruen, line 29:
```suggestion
    Abstract class for implementing input sources.
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+
+        return cohort
+
+    @abstractmethod
+    def get_entries(
+        self,
+        dataset: Dataset | None = None,
+    ) -> list[dict]:
+        """
+        Overide this method to get a list of data entries (dicts).
```
lgruen, line 91:
```suggestion
        Override this method to get a list of data entries (dicts).
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+        If dataset is not missing, it should be specific to a dataset.
```
lgruen, line 92: Not sure what's meant by this?

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+    @abstractmethod
+    def get_external_id(self, entry: dict) -> str | None:
+        """
+        Get external sample ID.
+        """
+
+    @abstractmethod
+    def get_participant_id(self, entry: dict) -> str | None:
+        """
+        Get participant ID. Can be also set later.
```
lgruen, line 116: What's meant by "Can be also set later" in this context? (Also below.)

cpg_pipes/providers/cpg/inputs.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+
+    def __init__(self, db: SMDB, check_files: bool = False):
+        super().__init__(check_files=check_files)
+        self.db = db
+
+    def populate_cohort(
+        self,
+        cohort: Cohort,
+        dataset_names: list[str] | None = None,
```
lgruen, line 28: Here and below, what's preferable in terms of Python style? `type | None` or `Optional[type]`?

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+    def get_participant_id(self, entry: dict) -> str | None:
+        """
+        Get participant ID from a sample dict.
+        """
+        return entry.get(FieldMap.participant_id.value, None)
+
+    # noinspection PyMethodMayBeStatic
+    def get_participant_sex(self, entry: dict) -> Sex:
+        """
+        Get participant's sex/gender from a sample dict.
```
lgruen, line 344:
```suggestion
        Get participant's sex from a sample dict.
```

cpg_pipes/providers/status.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Raised for problems reporting status
+    """
+
+
+class AnalysisStatus(Enum):
+    """
+    Corresponds to SMDB Analysis statuses:
+    https://github.com/populationgenomics/sample-metadata/blob/dev/models/enums
+    /analysis.py#L14-L21
```
lgruen, line 28:
```suggestion
    https://github.com/populationgenomics/sample-metadata/blob/dev/models/enums/analysis.py#L14-L21
```

lgruen (approved):
It would be great to get @illusional's review as well, given how much the SMDB API is used here.

cpg_pipes/providers/cpg/inputs.py
vladsaveliev: `type | None` is the new syntax, introduced in python 3.10 https://peps.python.org/pep-0604, and should become preferable as far as I understand. The goal is to allow to avoid importing the `typing` module for most cases (that's why python 3.10 also now allows using `dict`, `list`, `tuple` as type hints instead of `Typing.List`, `Typing.Dict` etc.)

cpg_pipes/providers/inputs.py
vladsaveliev: Just to point that in the implementations, this info is not necessarily available in an `entry` (e.g. in sample-metadata, we pull participant data for entire dataset with one api request). Though it should be clear anyway, I'll remove that comment

cpg_pipes/providers/inputs.py
vladsaveliev: If `dataset` parameter is provided, the method must return entries for this dataset only. I'll rename the parameter to `limit_to_dataset` to make it clearer

cpg_pipes/providers/cpg/status.py
```diff
@@ -0,0 +1,127 @@
...
+            b,
+            analysis_id=aid,
+            status=AnalysisStatus.COMPLETED,
+            analysis_type=analysis_type,
+            job_attrs=target.get_job_attrs(),
+        )
+
+        in_progress_j.depends_on(jobs[0])
+        completed_j.depends_on(*jobs)
+        return [in_progress_j] + jobs + [completed_j]
```
illusional, line 78: This syntax is fine, but also want to point out that this exists because I like the destructuring more

```suggestion
        return [in_progress_j, *jobs, completed_j]
```

cpg_pipes/providers/cpg/status.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Create a Hail Batch job that updates status of analysis.
+        """
+        try:
+            analysis_id_int = int(analysis_id)
+        except ValueError:
+            raise SmdbError('Analysis ID for sample-metadata must be int')
+
+        job_name = ''
+        job_name += f'Update status to {status.value}'
```
illusional, line 97:
```suggestion
        job_name = f'Update status to {status.value}'
```

cpg_pipes/providers/cpg/status.py
```diff
@@ -0,0 +1,127 @@
...
+        if analysis_type:
+            job_name += f' (for {analysis_type})'
+
+        j = b.new_job(job_name, job_attrs)
+        j.image(images.SM_IMAGE)
+        cmd = dedent(
+            f"""\
+        export SM_ENVIRONMENT=PRODUCTION
```
illusional, line 105: This should be the default, were there issues in the past?

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+                if f.strip()
+            ]
+            fqs2 = [
+                f.strip()
+                for f in entry.get(FieldMap.fqs_r2.value, '').split('|')
+                if f.strip()
+            ]
+            cram = entry.get(FieldMap.cram.value, None)
+            if fqs1:
+                if len(fqs1) != len(fqs1):
```
illusional, line 400: I think mypy should catch this:
```suggestion
                if len(fqs1) != len(fqs2):
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+                    raise InputProviderError(
+                        'Numbers of fqs_r1 and fqs_r2 values (pipe-separated) '
+                        'must match.'
+                    )
+                if self.check_files:
+                    for fq in fqs1 + fqs2:
+                        if not exists(fq):
+                            raise InputProviderError(f'FQ {fq} does not exist')
```
illusional, line 408: Not a big deal, but you could check all and report all failures:

```suggestion
                    missing_fastqs = [fq for fq in fqs1 + fqs2 if not exists(fq)]:
                    if missing_fastqs:
                        raise InputProviderError(f'FQs {missing_fastqs} does not exist')
```

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+class FieldMap(Enum):
+    """
+    CSV field names.
+    """
+
+    dataset = 'dataset'
+    sample = 'sample'
+    external_id = 'external_id'
+    participant_id = 'participant_id'
+    fqs_r1 = 'fqs_r1'
```
illusional, line 265: Why pass them like this, and not fq_pairs, and just chunk by two? Is this generally easier to produce?

cpg_pipes/providers/status.py
```diff
@@ -0,0 +1,60 @@
...
+logger = logging.getLogger(__file__)
+
+
+class StatusReporterError(Exception):
+    """
+    Raised for problems reporting status
+    """
+
+
+class AnalysisStatus(Enum):
```
illusional, line 24: I'm a little concerned about this reimplementation, why not just pull it from sample-metadata and subclass it - then if we add / remove statuses it will fail before it even starts.

tests/test_input_provider.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertEqual(len(s1.alignment_input), 2)
+        self.assertEqual(s1.pedigree.sex, Sex.MALE)
+        self.assertEqual(s1.sequencing_type, SequencingType.WGS)
+        s2 = ds.get_samples()[1]
+        self.assertEqual(s2.external_id, extid2)
+        self.assertTrue(s2.alignment_input.ext == 'cram')
+        self.assertEqual(s2.pedigree.sex, Sex.UNKNOWN)
+        self.assertEqual(s2.sequencing_type, SequencingType.EXOME)
+
+    @skip('Figure out SMDB permissions from GitHub workflows')
```
illusional, line 67: Could mock these calls? Otherwise you have to create a service-account json, and pick a test dataset, might be good to use the `dev` instance too.

cpg_pipes/providers/status.py
vladsaveliev: The idea was to have the abstract `StatusReporter` independent of sample-metadata implementation, thus it can't import sample-metadata. Not sure if there is a better way to work around it.

cpg_pipes/providers/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""
+Providers that can be overridden in specific infrastructures.
+"""
+
+from enum import Enum
+
+
+class InputProviderType(Enum):
```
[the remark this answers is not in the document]
vladsaveliev: Good idea!

cpg_pipes/providers/inputs.py
```diff
@@ -0,0 +1,417 @@
...
+        """
+        Return list of data entries. Optionally, specific for a dataset.
+        """
+        entries = self.d
+        if dataset:
+            entries = [
+                e
+                for e in entries
+                if (
+                    not self.get_dataset_name(e)
```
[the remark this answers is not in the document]
vladsaveliev: E.g. a CSV file entry doesn't need to have a "dataset" field. It would default to the global "analysis_dataset" (required `--analysis-dataset` option.)

cpg_pipes/providers/cpg/status.py
vladsaveliev: Yeah, it looks better

cpg_pipes/providers/inputs.py
vladsaveliev: Basically, lack of delimiters. We should support CSV and TSV files, and we use `|` to separate FASTQs from multiple lanes. So we need yet another delimiter to separate pairs, and it would get really confusing at this point. I guess I could borrow your parser code from sample-metadata, but didn't want to invest too much at the CSV provider, as we are only going to be using the sample-metadata input provider.

cpg_pipes/providers/inputs.py
vladsaveliev: `mypy` is just a type checker, it wouldn't catch that unfortunately :( Thanks!

cpg_pipes/providers/cpg/status.py
vladsaveliev: I guess it an artefact from the past. Removed it

tests/test_input_provider.py
vladsaveliev: It's using `acute-care-test`, and only reads it, so should be fine to use prod. With a service-account with only read permissions

PR #30, opened 2022-04-09
Storage provider, CPG Google Cloud storage policy implementation

vladsaveliev:
* Abstract storage provider, and CPG implementation with cloud storage policies

Stacked on https://github.com/populationgenomics/production-pipelines/tree/input_provider


cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    https://github.com/populationgenomics/team-docs/tree/main/storage_policies#main-vs-test
+    """
+
+    MAIN = 'main'
+    TEST = 'test'
+
+
+class Cloud(Enum):
+    """
+    Cloud storage provider and correponding protocol prefix.
```
lgruen, line 33:
```suggestion
    Cloud storage provider and corresponding protocol prefix.
```

cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    GS = 'gs'
+    AZ = 'az'
```
lgruen, line 37: Is that actually typical for Azure blobs? Or do people just use the `https` URI? @illusional, do you know? How is this implemented in Hail?

cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        Bucket to write primary results.
+        @param dataset: dataset/stack name
+        @param namespace: namespace (test or main)
+        @param suffix: (optional) suffix to append to the bucket name
+        @param version: (optional) pipeline version
+        @param sample: (optional) sample name
+        """
+
+    def get_analysis_bucket(
```
lgruen, line 68: Does this refer to our `main` buckets? Why isn't this implemented in the CPG provider?

cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        return self.get_bucket(
+            dataset=dataset,
+            namespace=namespace,
+            version=version,
+            sample=sample,
+            suffix='analysis',
+        )
+
+    def get_tmp_bucket(
```
lgruen, line 86: This also isn't implemented for the CPG provider.

cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        return self.get_bucket(
+            dataset=dataset,
+            namespace=namespace,
+            version=version,
+            sample=sample,
+            suffix='tmp',
+        )
+
+    def get_web_bucket(
```
lgruen, line 104: If we want to support these extra categories, there probably needs to be more documentation that describes when to use them. I wonder if this is too CPG-specific though and we should be aiming for something more abstract, e.g. a `category` that different providers can map however they like. This would be more similar to the current implementation of `output_path` in `cpg-utils`.

lgruen (changes requested):

cpg_pipes/providers/storage.py
vladsaveliev: Hail uses `hail-az`. StorageProvider specifies where the files are written (outputs and intermediates), so we shouldn't depend on the suffix that Hail is using? But can parametrise it, of course!

cpg_pipes/providers/storage.py
vladsaveliev: Other implementations might also use some sort of an analysis bucket. Or they can disregard the `suffix` parameter. But `get_analysis_bucket` must exist in the abstract class anyway, because the stages don't have to know about the specific implementations.

cpg_pipes/providers/storage.py
vladsaveliev: Same, it's not an abstract method. It redirects to get_bucket(suffix='tmp')

cpg_pipes/providers/storage.py
lgruen: Thanks, that's really interesting. Wouldn't it be fine to depend on the suffix that Hail is using, given that this framework is built on top of the Hail Batch?

Is the `hail-az` suffix only used internally in Hail or is that something a user would specify as well (for inputs / outputs)? If it's the latter, would it be worth bringing this up with the Hail team and see whether they'd be okay to use `az` instead, if that's somewhat commonly used already (e.g. in `cloudpathlib`), to be more compatible?

cpg_pipes/providers/storage.py
lgruen: Right, but we don't have `analysis` buckets anymore?

cpg_pipes/providers/storage.py
lgruen: You could potentially just link to our storage policies even for the abstract class, as that provides a general idea of the categories.

cpg_pipes/providers/storage.py
vladsaveliev: But we do, we re-added analysis buckets at some point:
https://github.com/populationgenomics/analysis-runner/blob/main/stack/__main__.py#L236-L238
Though perhaps the goal was to allow read by human, and we allow that for main buckets already, so we perhaps don't need analysis buckets anymore? And I can see that we don't have them described in team-docs.

cpg_pipes/providers/cpg/storage.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        dataset: str,
+        namespace: Namespace,
+        version: str | None = None,
+        sample: str = None,
+    ) -> str | None:
+        """
+        URL corresponding to the WEB bucket.
+        """
+        url = f'https://{namespace.value}-web.populationgenomics.org.au/' f'{dataset}'
```
illusional, line 64: Is this a reformatting thing, I know it collapses to the same but can it just be shorted to:

```suggestion
        url = f'https://{namespace.value}-web.populationgenomics.org.au/{dataset}'
```

cpg_pipes/providers/storage.py
```diff
@@ -0,0 +1,124 @@
...
+from abc import ABC, abstractmethod
+from typing import Union
+
+from cloudpathlib import CloudPath
+from cloudpathlib.anypath import to_anypath
+import hail_az  # importing to register hail-az prefix.
+
+
+# Path can be either a cloud URL or a local posix file path.
+Path = Union[CloudPath, pathlib.Path]
```
illusional, line 16: Doesn't AnyPath cover this from cloudpathlib: https://cloudpathlib.drivendata.org/stable/api-reference/anypath/

cpg_pipes/providers/storage.py
illusional: I think we should use `hail-az` until that linked issue is resolved - I'm in favour of URLs being unique and I would be hesitant to rely on something that could change (or _should_ change IMO).

lgruen: It's super cool that you extended `cloudpathlib` that way :). Would that extension be useful as part of `cpg-utils`?

In either case I wouldn't see this blocking this PR, but switching to `hail-az` (as @illusional suggested) seems to be the safer choice?

cpg_pipes/providers/cpg/storage.py
vladsaveliev: Black artefact :( Thanks!

cpg_pipes/providers/storage.py
vladsaveliev: Unfortunately `AnyPath` does not equal `Union[CloudPath, Path]`. Calling `AnyPath(path)` would return either `CloudPath` or `Path`, but neither derives from `AnyPath`, so you can't use it in type hints :(

cpg_pipes/providers/storage.py
lgruen: Thanks for pointing that out! I had completely forgotten that we revived the `analysis` buckets, even though I wrote that PR in September myself ;).

I've checked the permissions again to understand the difference between the `analysis` and `main` buckets in the current setup. As far as I can tell, members of the access group currently only have *listing* permissions for the `main` bucket, but can *read* the contents for the `analysis` bucket:
https://github.com/populationgenomics/analysis-runner/blob/f6045469308d1445cc262b5a1ecd5cbcafc88ab0/stack/__main__.py#L544-L549

I'll now finally update the storage policies doc (which got me confused in the first place).

lgruen (changes requested):
Does this need to be rebased on `main` now? There seem to be a lot of unrelated changes in the diff.

cpg_pipes/providers/storage.py
vladsaveliev: Have we thought about the internal storage policy for Azure? Not sure how account and container would be specified (e.g. through environment variables like in Hail bootstrap), whether we would use `cpg-{dataset}-` prefix anywhere, or we can use account and container exactly for that, etc.

vladsaveliev: For example, in `CpgStorageProvider` we might have something like that:
```
        if self.cloud == Cloud.HAIL_AZ:
            base = f'hail-az://{self.account}/{self.container}/'
        elif self.cloud == Cloud.GS:
            base = f'gs://cpg-'
        path = f'{base}{dataset}-{namespace}')
```

And in the abstract `StorageProvider`,

```
        container = os.environ.get('STORAGE_CONTAINER_NAME')
        account = os.environ.get('STORAGE_ACCOUNT_NAME')
        if cloud == Cloud.HAIL_AZ:
            not_set = ', '.join(v for k, v in {
                container: '$STORAGE_CONTAINER_NAME',
                account: '$STORAGE_ACCOUNT_NAME',
            } if not k)
            if not_set:
                raise ValueError(
                    f'For Azure storage provider, these two environment variables '
                    f'must be set: $STORAGE_CONTAINER_NAME and $STORAGE_ACCOUNT_NAME. '
                    f'Not set: {not_set}'
                )
```

Or in `CpgStorageProvider` we can encode `dataset` in `container`, and hardcode `account` (like `cpg-` prefix for GS), and do not require environment variables for the `CpgStorageProvider` implementation.

vladsaveliev:
Rebased, though there are still unrelated changes from the `input_provider` branch which is not merged

cpg_pipes/providers/storage.py
lgruen: I think it would be fine to set additional environment variables (and potentially also use those within `cpg-utils`'s `output_path` if we adopt the scheme to encode the account / container in the path).

Dan has responded on the Zulip thread, by the way -- maybe we should look into `wasbs` too?
https://docs.microsoft.com/en-us/azure/hdinsight/hdinsight-hadoop-use-blob-storage

vladsaveliev: `wasbs` O_o 
what does it even stand for?

vladsaveliev: Anyway, created an issue: https://github.com/populationgenomics/production-pipelines/issues/32
I don't think Azure should any longer block the code that works on Google.

vladsaveliev:
My personal opinion that pull requests shouldn't be blocked by requests of new features, because that contradicts the idea that pull requests should be short and only do one thing. New feature requests are really welcomed, but they should be just recorded in an issue tracker. So e.g. Azure support sounds like an additional feature request, and should go as a separate PR.

lgruen (approved):

lgruen:
> My personal opinion that pull requests shouldn't be blocked by requests of new features, because that contradicts the idea of short and do-one-thing PRs. New feature requests are really welcomed, but they should be just recorded in an issue tracker. So e.g. in this case Azure support sounds like an additional feature request, and should go in a separate PR.

Certainly agree with you, but it's just tempting for reviewers (including me) to start discussing partially implemented features (like Azure storage in this case, which was already part of the original PR) :). Apologies if that ended blocking things. I found the conversation about the protocol prefix etc. really helpful, but it certainly could have been part of another PR.

vladsaveliev:
@lgruen, good point, shouldn't have submitted partially implemented features! Will take a note for the future :)

And the discussion was definitely helpful!

PR #35, opened 2022-04-21
Add slack messaging to status reporter

vladsaveliev:
Activated through `--slack-channel=` or `CPG_SLACK_CHANNEL`. Token is provided through `CPG_SLACK_TOKEN`, or `slack-seqr-loader-token` secret in `cpg-common`.

Example of usage to send a link to an analysis HTML report for a dataset:

```python
status_reporter.slack_env(job)
text = f'*[{dataset.name}]* <{html_url}|analysis report>'
job.command(dedent(status_reporter.slack_message_cmd(text=text)))
```

### DIFF
--- a/cpg_pipes/providers/cpg/status.py
+++ b/cpg_pipes/providers/cpg/status.py
@@ -25,8 +25,23 @@ class CpgStatusReporter(StatusReporter):
     database Analysis entries.
     """
 
-    def __init__(self, smdb: SMDB):
+    def __init__(self, smdb: SMDB, slack_channel: str | None = None):
+        super().__init__()
         self.smdb = smdb
+        self.slack_channel = slack_channel or os.environ.get('CPG_SLACK_CHANNEL')
+        self.slack_token = os.environ.get('CPG_SLACK_TOKEN')
+        if self.slack_channel and not self.slack_token:
+            project_id = 'cpg-common'
+            secret_name = 'slack-seqr-loader-token'
+            slack_token_secret = (
+                f'projects/{project_id}/secrets/{secret_name}/versions/latest'
+            )
+            secret_manager = secretmanager.SecretManagerServiceClient()
+            # noinspection PyTypeChecker
+            response = secret_manager.access_secret_version(
+                request={'name': slack_token_secret}
+            )
+            self.slack_token = response.payload.data.decode('UTF-8')
 
     def add_updaters_jobs(
         self,
--- a/cpg_pipes/providers/status.py
+++ b/cpg_pipes/providers/status.py
@@ -46,6 +46,10 @@ class StatusReporter(ABC):
     Status reporter
     """
 
+    def __init__(self):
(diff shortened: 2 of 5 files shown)


cpg_pipes/providers/status.py
```diff
@@ -59,3 +63,39 @@ def add_updaters_jobs(
         """
         Add Hail Batch jobs that update the analysis status.
         """
+
+    def slack_env(self, j: Job):
```
lgruen, line 67: Should this be part of the `StatusReporter` interface? I would have expected the Slack reporter to be one concrete subclass of a `StatusReporter`. Can we come up with a more generic interface that would fit e.g. both Slack and SMDB updates?

cpg_pipes/providers/status.py
```diff
@@ -59,3 +63,39 @@ def add_updaters_jobs(
...
+        """
+        Add environment variables that configure Slack reporter.
+        """
+        if self.slack_channel and self.slack_token:
```
lgruen, line 71: Should this be an assertion instead?

cpg_pipes/providers/status.py
```diff
@@ -59,3 +63,39 @@ def add_updaters_jobs(
...
+        text: str | None = None,
+        data: dict[str, str] | None = None,
+    ) -> str:
+        """
+        Make a bash command that prepares and sends a Slack message.
+        Message can be constructed from a dict `data`, or can be passed directly 
+        as text (`text`). In either case, strings can use Slack `mrkdwn` 
+        for formatting: https://api.slack.com/reference/surfaces/formatting
+        """
+        if not self.slack_channel or not self.slack_token:
```
lgruen, line 86: Should this be an assertion / raise an error instead of returning an empty string? (Same below.)

cpg_pipes/providers/cpg/status.py
```diff
@@ -25,8 +25,23 @@ class CpgStatusReporter(StatusReporter):
     database Analysis entries.
     """
 
-    def __init__(self, smdb: SMDB):
+    def __init__(self, smdb: SMDB, slack_channel: str | None = None):
+        super().__init__()
         self.smdb = smdb
+        self.slack_channel = slack_channel or os.environ.get('CPG_SLACK_CHANNEL')
+        self.slack_token = os.environ.get('CPG_SLACK_TOKEN')
+        if self.slack_channel and not self.slack_token:
```
lgruen, line 33: Do we want this logic here or should this instead be set up in the wrapper (i.e. analysis-runner in our case)? In other places we just rely on the environment variables already being set up. That would also make it more clear how this should be configured (right now there's a lot of different options how to set this up, with a mix of flags, environment variables, and secrets).

We could also consider moving the `slack_message_cmd` implementation to `cpg-utils`, which could be more generally useful to send status updates (and reusable by the `StatusReporter`).

lgruen (changes requested):

PR #37, opened 2022-04-21
Add Target, Dataset, Sample, Cohort

vladsaveliev:
Another huge PR, this time adding the abstract Target class, representing a stage target, following the pipeline's model. And three implementations: Sample, Dataset, Cohort. Dataset is somewhat tied to CPG dataset, so not sure if it needs some extra level abstraction for external users to decouple from CPG-specific stack vs. project details.

### DIFF
--- a/cpg_pipes/targets.py
+++ b/cpg_pipes/targets.py
@@ -289,7 +289,7 @@ class Dataset(Target):
         storage_provider: StorageProvider | None = None,
     ) -> 'Dataset':
         """
-        Create a dataset and add it into the cohort.
+        Create a dataset.
         """
         # Normalising the dataset's name:
         name = build_dataset_name(*parse_stack(name, namespace))
--- a/cpg_pipes/targets.py
+++ b/cpg_pipes/targets.py
@@ -383,7 +383,7 @@ class Dataset(Target):
         alignment_input: AlignmentInput | None = None,
     ) -> 'Sample':
         """
-        Create a new sample and add it into the dataset.
+        Create a new sample and add it to the dataset.
         """
         if id in self._sample_by_id:
             logger.debug(f'Sample {id} already exists in the dataset {self.name}')
--- a/cpg_pipes/targets.py
+++ b/cpg_pipes/targets.py
@@ -1,5 +1,6 @@
 """
 Pipeline stage target and subclasses: Cohort, Dataset, Sample.
+Plus Sample properties: PedigreeInfo, Sex.
 """
 import hashlib
 import logging
@@ -483,7 +484,7 @@ class Sample(Target):
         self.id = id
         self._external_id = external_id
         self.dataset = dataset
-        self.sequencing_type = sequencing_type or SequencingType.WGS
+        self.sequencing_type = sequencing_type or SequencingType.GENOME
         self._participant_id = participant_id
         self.meta: dict = meta or dict()
         self.pedigree: PedigreeInfo | None = pedigree
--- a/cpg_pipes/targets.py
+++ b/cpg_pipes/targets.py
@@ -188,7 +188,7 @@ class Cohort(Target):
         storage_provider: StorageProvider | None = None,
     ) -> 'Dataset':
         """
-        Create a dataset and add it into the cohort.
+        Create a dataset and add it to the cohort.
         """
         namespace = namespace or self.analysis_dataset.namespace
         # Normalising the dataset's name:
--- a/cpg_pipes/targets.py
+++ b/cpg_pipes/targets.py
@@ -1,0 +1,639 @@
+"""
+Pipeline stage target and subclasses: Cohort, Dataset, Sample.
+"""
+import hashlib
+import logging
+from dataclasses import dataclass
+from enum import Enum
+from typing import Optional
+import pandas as pd
+
+from . import Path
+from .types import AlignmentInput, CramPath, GvcfPath, SequencingType
+from .providers.storage import StorageProvider, Namespace
+
+logger = logging.getLogger(__file__)
+
+
+class Target:
+    """
+    Defines a target that stage can act upon. Classes like Sample, Dataset, Pipeline
+    extend this class.
+    """
+
+    def __init__(self):
+        # Whether to process even if outputs exist:
+        self.forced: bool = False
+        # If not set, exclude from the pipeline:
+        self.active: bool = True
+
+    def get_samples(self, only_active: bool = True) -> list['Sample']:
+        """
+        Get flat list of all samples corresponding to this target.
+        """
+        raise NotImplementedError
+
+    def get_sample_ids(self, only_active: bool = True) -> list[str]:
+        """
+        Get flat list of all sample IDs corresponding to this target.
+        """
+        return [s.id for s in self.get_samples(only_active=only_active)]
+
+    def alignment_inputs_hash(self) -> str:
+        """
+        Unique hash string of sample alignment inputs. Useful to decide
+        whether the analysis on the target needs to be rerun.
+        """
+        s = ' '.join(
+            sorted(
+                [
+                    str(s.alignment_input)
+                    for s in self.get_samples()
+                    if s.alignment_input is not None
+                ]
+            )
+        )
+        h = hashlib.sha256(s.encode()).hexdigest()[:38]
+        return f'{h}_{len(self.get_sample_ids())}'
+
+    @property
+    def target_id(self) -> str:
+        """
+        ID should be unique across target of all levels.
+
+        We are raising NotImplementedError instead of making it an abstract class, 
+        because mypy is not happy about binding TypeVar to abstract classes, see:
+        https://stackoverflow.com/questions/48349054/how-do-you-annotate-the-type-of
+        -an-abstract-class-with-mypy
+
+        Specifically,
+        ```
+        TypeVar('TargetT', bound=Target)
+        ```
+        Will raise:
+        ```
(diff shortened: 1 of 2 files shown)


cpg_pipes/targets.py
```diff
@@ -0,0 +1,639 @@
...
+        external_id: str | None = None,
+        participant_id: str | None = None,
+        sequencing_type: SequencingType | None = None,
+        meta: dict | None = None,
+        sex: Optional['Sex'] = None,
+        pedigree: Optional['PedigreeInfo'] = None,
+        alignment_input: AlignmentInput | None = None,
+    ) -> 'Sample':
+        """
+        Create a new sample and add it into the dataset.
```
lgruen, line 386:
```suggestion
        Create a new sample and add it to the dataset.
```

cpg_pipes/targets.py
```diff
@@ -0,0 +1,639 @@
...
+
+    def get_job_prefix(self) -> str:
+        """
+        Prefix job names.
+        """
+        return f'{self.dataset.name}/{self.id}: '
+
+
+@dataclass
+class PedigreeInfo:
```
lgruen, line 607: Should we use the `peddy` classes here instead?

cpg_pipes/targets.py
```diff
@@ -0,0 +1,639 @@
+"""
+Pipeline stage target and subclasses: Cohort, Dataset, Sample.
```
lgruen, line 2: At the moment this also includes pedigree classes. Should those be moved to a separate module?

lgruen (approved):
Only added a few nits, but it would be good to get @illusional's review as well.

cpg_pipes/targets.py
```diff
@@ -0,0 +1,639 @@
...
+
+    def alignment_inputs_hash(self) -> str:
+        """
+        Unique hash string of sample alignment inputs. Useful to decide
+        whether the analysis on the target needs to be rerun.
+        """
+        s = ' '.join(
+            sorted(
+                [
+                    str(s.alignment_input)
```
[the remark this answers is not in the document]
vladsaveliev: Good idea! 
Added an issue. https://github.com/populationgenomics/production-pipelines/issues/38
Need to write tests for that feature anyway. There might be edge cases that I haven't thought through.

cpg_pipes/targets.py
vladsaveliev: Thought about it a bit. We indeed don't need PedigreeInfo at all. Created an issue:
https://github.com/populationgenomics/production-pipelines/issues/39

cpg_pipes/targets.py
vladsaveliev: Thought about it a bit. We indeed don't need PedigreeInfo at all. Created an issue:
https://github.com/populationgenomics/production-pipelines/issues/39

cpg_pipes/targets.py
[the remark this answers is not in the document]
illusional: MD5 is usually stored in the metadata server where we are provided it - we also do check if we're provided an MD5.

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+        Prefix job names.
+        """
+        raise NotImplementedError
+
+    def get_sequencing_type(self) -> SequencingType:
+        """
+        Sequencing type of samples. Will throw an error if more than
+        one sequencing type is found.
+        """
+        sequencing_types = set(s.sequencing_type for s in self.get_samples())
```
illusional, line 98: As long as `SequencingType` is hashable, which I didn't think it was - maybe explicitly taking the string representation and then casting back to SequencingType.

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+    def target_id(self) -> str:
+        """Unique target ID"""
+        return self.name
+
+    def get_datasets(self, only_active: bool = True) -> list['Dataset']:
+        """
+        Gets list of all datasets.
+        Include only "active" datasets (unless only_active is False)
+        """
+        return [
```
illusional, line 146: What's an inactive dataset? I might be missing context so feel free to ignore if it's obvious with other context.

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+    resolved to stack="seqr" and is_test=True, unless `namespace` is provided
+    explicitly.
+
+    Returns the stack id and a corrected namespace.
+    """
+    namespace = namespace or Namespace.MAIN
+    if name.endswith('-test'):
+        stack = name[: -len('-test')]
+        if namespace == Namespace.MAIN:
+            namespace = Namespace.TEST
```
illusional, line 235: Why not just always set to TEST?

```suggestion
        namespace = Namespace.TEST
```

cpg_pipes/targets.py
```diff
@@ -0,0 +1,639 @@
...
+        self,
+        analysis_dataset_name: str,
+        namespace: Namespace,
+        name: str | None = None,
+        storage_provider: StorageProvider | None = None,
+    ):
+        """
+        @param name: name of the cohort
+        @param analysis_dataset_name: deferring creation of the analysis
+        dataset in case if it is also a part of the cohort datasets.
```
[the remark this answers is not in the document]
vladsaveliev: Sorry, that's a remnant from an older code. Cleaned that up!

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+    def parse(sex: str | None) -> 'Sex':
+        """
+        Parse a string into a Sex object.
+        """
+        if sex:
+            if sex.lower() in ('m', 'male', '1'):
+                return Sex.MALE
+            if sex.lower() in ('f', 'female', '2'):
+                return Sex.FEMALE
+        return Sex.UNKNOWN
```
illusional, line 460: I think UNKNOWN should be a specific case handled within the conditional, and unrecognisable values should be an exception:

```suggestion
            if sex.lower() in ('u', 'unknown', '0'):
                return Sex.UNKNOWN
            raise ValueError(f'Unrecognised sex value {sex}')
        return Sex.UNKNOWN
```

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+        if sex:
+            self.pedigree = PedigreeInfo(
+                sample=self,
+                fam_id=self.participant_id,
+                sex=sex,
+            )
+        self.alignment_input: AlignmentInput | None = alignment_input
+
+    def __repr__(self):
+        return (
```
illusional, line 497: Probably not a big deal, but could construct this a bit cleaner. This suggestion probably isn't formatted properly.

```suggestion
    values = {
        'participant': self._participant_id if self._participant_id else '',
        'forced': str(self.forced),
        'active': str(self.active),
        'meta': str(self.meta),
        'sequencing_type': str(self.sequencing_type.value),
        'alignment_input': self.alignment_input if self.alignment_input else ''
        'pedigree': self.pedigree if self.pedigree else ''
    }

    retval = f'Sample({self.dataset.name}/{self.id}'
    if self._external_id:
        retval += f'|{self._external_id}'
    return retval + ''.join(f', {k}={v}' for k, v in values.items())
```

cpg_pipes/targets.py
```diff
@@ -0,0 +1,640 @@
...
+        ext_id = f'|{self._external_id}' if self._external_id else ''
+        return f'Sample({self.dataset.name}/{self.id}{ext_id}{ai_tag})'
+
+    @property
+    def participant_id(self) -> str:
+        """
+        Get ID of participant corresponding to this sample,
+        or substitute it with external ID.
+        """
+        return self._participant_id or self.external_id
```
illusional, line 534: Should all samples that go through this process have a participant_id, should this be an assert?

illusional (commented):
Looks good, nothing critical from me, just a few nitpicks as well.

cpg_pipes/targets.py
vladsaveliev: Enums are hashable, seems to be working good!

```
>>> set([SequencingType.GENOME, SequencingType.EXOME])
{<SequencingType.EXOME: 'exome'>, <SequencingType.GENOME: 'genome'>}
>>> set([SequencingType.GENOME, SequencingType.GENOME])
{<SequencingType.GENOME: 'genome'>}
```

cpg_pipes/targets.py
vladsaveliev: Base `Target` class defines `.active` field, which is inherited by all subclasses, so using it for consistency.
Also, in the pipelines logic, `active=False` is used in couple with `--skip-samples-with-missing-input` to flag targets (samples and datasets) that are not to be used in the pipeline:
https://github.com/populationgenomics/production-pipelines/blob/initial/cpg_pipes/pipeline/pipeline.py#L586-L587

cpg_pipes/targets.py
vladsaveliev: Thanks!

cpg_pipes/targets.py
vladsaveliev: Agree!

cpg_pipes/targets.py
vladsaveliev: That looks much better!

cpg_pipes/targets.py
vladsaveliev: I don't want to force non-CPG users to use participant IDs. I think they should be able to use only one type of IDs if they want, and the code that expects participant IDs still should work

PR #45, opened 2022-05-17
Use cpg-utils for CPG storage provider

vladsaveliev:
Updating CpgStorageProvider to use `cpg_utils.hail_batch.dataset_path()`.

Not sure if it makes sense to merge it here, or work directly on moving this into cpg-utils. Happy to discuss more.

There are multiple stacked PR (https://github.com/populationgenomics/production-pipelines/pull/46, https://github.com/populationgenomics/production-pipelines/pull/47, https://github.com/populationgenomics/production-pipelines/pull/48). It would be ideal to quickly merge everything one after another, to avoid running into merge problems with so many branches depending on each other. I'm also okay to close them all, and move the codebase into cpg_utils.

### DIFF
--- a/cpg_pipes/pipeline/cli_opts.py
+++ b/cpg_pipes/pipeline/cli_opts.py
@@ -12,7 +12,7 @@ from ..providers import (
     StatusReporterType,
     InputProviderType,
 )
-from ..providers.storage import Namespace, Cloud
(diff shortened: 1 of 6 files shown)


lgruen (approved):
Looks great! If we could merge `cpg-pipes` and `cpg-utils` at some point, that would be nice.

PR #47, opened 2022-05-17
Add abstract/CPG images provider

vladsaveliev:
Abstract provider of docker images, and CPG implementation that makes use of `cpg_utils.image_path()`.


cpg_pipes/providers/images.py
```diff
@@ -0,0 +1,49 @@
+"""
+Docker images used in the pipelines.
+"""
+import os
+
+
+AVAILABLE = {
```
lgruen, line 7: Would it make sense to source this from a JSON file (or similar), so it can be updated programmatically whenever a new image version is built?

Should this potentially be part of the TOML options blob?

cpg_pipes/providers/images.py
```diff
@@ -0,0 +1,49 @@
...
+                f'Image {name} is not available. Known images: {AVAILABLE}'
+            )
+        return os.path.join(self._prefix, AVAILABLE[name])
+
+    @staticmethod
+    def driver_image() -> str:
+        """
+        Hail "driver" image.
+        """
+        return os.environ['CPG_DRIVER_IMAGE']
```
lgruen, line 49: This will need to be updated to use the TOML config.

lgruen (changes requested):

cpg_pipes/providers/images.py
vladsaveliev: We currently have all CPG variables essentially built-in in analysis-runner code https://github.com/populationgenomics/analysis-runner/blob/d4441f6ffaf308d97c7207507bbc9b0a1431ddeb/server/util.py#L17-L30
Would it make sense to make a TOML config be the source of those? And we could add other stuff like docker names there as well.

cpg_pipes/providers/images.py
vladsaveliev: Updated!

PR #49, opened 2022-05-18
Click config file: check potential typos

vladsaveliev:
When a config file contains parameters that are not defined by normal `click.option()` decorators, `click-config-file` would just silently ignores that. Even though it's sometimes a useful behaviour when you want to reuse the same config for different pipelines, we still want to catch potential typos in a config, e.g. like the one fixed in this commit: `only_samples` vs. `only-samples`. 

So I'm dropping the `click-config-file` dependency, and instead adding a custom callback that prints a warning if a config option doesn't match defined click decorators.

UPDATE: it's easy to overlook the warning, so changed the default behaviour to failure for now.

### DIFF
--- a/cpg_pipes/pipeline/cli_opts.py
+++ b/cpg_pipes/pipeline/cli_opts.py
@@ -253,7 +253,10 @@ def pipeline_click_options(function: Callable) -> Callable:
         'config',
         type=click.Path(exists=True, dir_okay=False, readable=True),
         help='Read configuration from a YAML FILE.',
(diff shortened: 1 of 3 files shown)


vivbak:
Do you happen to know whats causing the type error in test_pipeline? 
`TypeError: the JSON object must be str, bytes or bytearray, not NoneType`
https://github.com/populationgenomics/production-pipelines/runs/6488222074?check_suite_focus=true#step:8:62

vivbak (approved):

PR #57, opened 2022-05-29
Support multiple sequencing types

vladsaveliev:
* Handles sequencing data of multiple sequencing types. In input provider, sanity checks that there is only one "sequence" entry per type.
* Adds `--sequencing-type` to the seqr_loader pipeline to limit to specific type.
* `AlignmentInput` is not a proper class, incapsulating `CramPath` and `FastqPairs` as `data` field.

### DIFF
--- a/cpg_pipes/providers/cpg/inputs.py
+++ b/cpg_pipes/providers/cpg/inputs.py
@@ -162,15 +162,15 @@ class SmdbInputProvider(InputProvider):
             for d in found_seqs_by_sid.get(sample.id, []):
                 seq = SmSequence.parse(d, self.check_files)
                 if seq.alignment_input:
-                    seq_type = seq.alignment_input.sequencing_type
-                    if seq_type in sample.alignment_input_by_seq_type:
+                    if seq.sequencing_type in sample.alignment_input_by_seq_type:
                         raise InputProviderError(
                             f'{sample}: found more than 1 alignment input with '
-                            f'sequencing type: {seq_type.value}. Check your input '
-                            f'provider to make sure there is only one data source of '
-                            f'sequencing type per sample.'
+                            f'sequencing type: {seq.sequencing_type.value}. Check your '
+                            f'input provider to make sure there is only one data source '
+                            f'of sequencing type per sample.'
                         )
-                    sample.alignment_input_by_seq_type[seq_type] = seq.alignment_input
+                    sample.alignment_input_by_seq_type[seq.sequencing_type] = \
+                        seq.alignment_input
 
     def populate_analysis(self, cohort: Cohort) -> None:
         """
--- a/cpg_pipes/types.py
+++ b/cpg_pipes/types.py
@@ -3,6 +3,7 @@ Wrappers for bioinformatics file types (CRAM, GVCF, FASTQ, etc).
 """
 
 import logging
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
 from typing import List, Union
@@ -56,7 +57,24 @@ class SequencingType(Enum):
         return str_to_val[str_val]
 
 
-class CramPath:
+class AlignmentInput(ABC):
+    """
+    Data that works as input for alignment or realignment.
+    """
+    @abstractmethod
+    def exists(self) -> bool:
+        """
+        Check if all files exist.
+        """
+
+    @abstractmethod
+    def path_glob(self) -> str:
+        """
+        Compact representation of file paths. 
+        """
+
+
+class CramPath(AlignmentInput):
     """
     Represents alignment data on a bucket within the pipeline.
     Includes a path to a CRAM or a BAM file along with a corresponding index,
@@ -104,6 +122,13 @@ class CramPath:
                 f'{self.ext}.{self.index_ext}': str(self.index_path),
             }
         )
+    
+    def path_glob(self) -> str:
+        """
+        Compact representation of file paths. 
+        For a CRAM, it's just the CRAM file path.
+        """
+        return str(self.path)
 
 
 class GvcfPath:
@@ -169,7 +194,9 @@ class FastqPair:
         Makes a pair of ResourceFile objects for r1 and r2.
         """
         return FastqPair(*[
-            self[i] if isinstance(self[i], ResourceFile) else b.read_input(str(self[i]))
+            self[i] 
+            if isinstance(self[i], ResourceFile) 
+            else b.read_input(str(self[i]))
             for i in [0, 1]
         ])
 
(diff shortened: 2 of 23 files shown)


cpg_pipes/providers/cpg/inputs.py
```diff
@@ -135,25 +137,40 @@ def populate_alignment_inputs(self, cohort: Cohort) -> None:
...
-            if seq_info := seq_by_sid.get(sample.id):
-                seq = SmSequence.parse(seq_info, self.check_files)
-                sample.alignment_input = seq.alignment_input
-                sample.sequencing_type = seq.sequencing_type
+            for d in found_seqs_by_sid.get(sample.id, []):
+                seq = SmSequence.parse(d, self.check_files)
+                if seq.alignment_input:
+                    seq_type = seq.alignment_input.sequencing_type
+                    if seq_type in sample.alignment_input_by_seq_type:
+                        raise InputProviderError(
```
illusional, line 167: Why not allow multiple? This isn't a valid case for bulk imports, but might be for top-up sequencing?

cpg_pipes/types.py
```diff
@@ -141,36 +141,6 @@ def __str__(self):
...
-            all_fastq_paths.extend([pair.r1, pair.r2])
-        return ''.join([
-            f'{{{",".join(set(chars))}}}' if len(set(chars)) > 1 else chars[0] 
-            for chars in zip(*map(str, all_fastq_paths))
-        ])
-
-
 class SequencingType(Enum):
     """
     Type (scope) of a sequencing experiment.
```
illusional, line 146: You wouldn't consider extending the SequencingType from the SM API? I'm just about to add an ONT sequencing type, and we don't expect any refactoring of these values from now.

cpg_pipes/types.py
```diff
@@ -204,3 +174,51 @@ def parse(str_val: str) -> 'SequencingType':
                 f'Available: {list(str_to_val.keys())}'
             )
         return str_to_val[str_val]
+
+
+class AlignmentInput:
+    """
+    Alignment input can be an indexed CRAM file, or a list of FASTQ file pairs.
```
illusional, line 181: Why couldn't it be unindexed?

cpg_pipes/types.py
```diff
@@ -204,3 +174,51 @@ def parse(str_val: str) -> 'SequencingType':
...
+                utils.exists(pair.r1) and utils.exists(pair.r2) 
+                for pair in self.data
+            )
+
+    def compact_file_paths(self) -> str:
+        """
+        Compact representation of file paths. 
+        For CRAM, it's simply path to CRAM. 
+        For FASTQ pairs, it's glob string to find all FASTQ files.
+        """
```
illusional, line 212: Not a super big deal, but this would be cool as a static method, with doctests

pipelines/configs/seqr.yml
```diff
@@ -1,12 +1,17 @@
 namespace: main
 analysis_dataset: seqr
-skip_datasets: ['ohmr3-mendelian', 'ohmr4-epilepsy', 'mito-disease', 'flinders-ophthal']
+#skip_datasets: ['ohmr3-mendelian', 'ohmr4-epilepsy', 'mito-disease', 'flinders-ophthal']
 skip_samples: [CPG11783,CPG12310,CPG199679,CPG13326]  # acute-care,acute-care,acute-care,perth-neuro
```
illusional, line 4: Long term goal would be to move this to the SM somehow - I've added a card to our backlog to address this.

pipelines/configs/seqr.yml
```diff
@@ -1,12 +1,17 @@
...
+datasets: ['ohmr3-mendelian', 'ohmr4-epilepsy']  # ['heartkids', 'ravenscroft-rdstudy']  # ['circa', 'acute-care', 'perth-neuro', 'heartkids', 'ravenscroft-rdstudy']
+check_inputs: true
 keep_scratch: true
 skip_samples_with_missing_input: false
-ped_checks: false
-version: may21
+version: 'may21'
```
illusional, line 9: What does this version mean? It's may22 btw, could it be automatically generated?

tests/test_jobs.py
```diff
@@ -214,11 +217,11 @@ def test_align_cram(self):
         self.assertEqual(sid, _read_file(cram_details_paths['sample_name']))
         cram_details = {k: _read_file(v) for k, v in cram_details_paths.items()}
         print(cram_details_paths)
-        self.assertAlmostEqual(223007, int(cram_details['reads_num']), delta=50)
+        self.assertAlmostEqual(223007, int(cram_details['reads_num']), delta=100)
```
illusional, line 220: I love the name of this method

illusional (approved):
Only minor feedback so approving, but please resolve my comments before merging

cpg_pipes/providers/cpg/inputs.py
vladsaveliev: I wasn't sure how should the `align` stage handle that. Should we assume it's always going to be a top-up, so it should align all available data together? Do we distinguish top-ups from new data that should replace previous? I'm cautious of the scenario when there are multiple sequences because of a parser or metadata bug, and the pipeline just assumes a top-up.

illusional:
Just noting there are linting issues that look related

cpg_pipes/types.py
vladsaveliev: It would imply a hard dependency on the `sample-metadata` package. I guess why not though

cpg_pipes/types.py
vladsaveliev: I could add an extra check in the job and index CRAM automatically if index is not available. Just didn't need so far.

pipelines/configs/seqr.yml
vladsaveliev: Oh yeah. I shouldn't probably track this config under version control for now at all, just polluting commit contents.

pipelines/configs/seqr.yml
vladsaveliev: I meant "21 May", as in day, not year, haha.
New version means that the workflow wouldn't attempt to reuse intermediate results from previous runs (except for CRAMs and GVCFs). With the same version, intermediate joint-calling, VQSR, matrix tables, etc would be reused if exist. I should probably move into the direction of checking if the inputs changed (i.e. the set of input samples and input FASTQ changed) instead of these strict versions, but that's for another PR.

tests/test_jobs.py
vladsaveliev: Haha. We also need such metric for the CPG: "pipeline almost worked", "seqr load is almost done"

cpg_pipes/types.py
vladsaveliev: Ah, so doctests only work with static method? I can't get it working with a class method, it throws `Module [...] has no class "AlignmentInput"` :(

cpg_pipes/types.py
vladsaveliev: Ah, actually the enum is not a part of the `sample-metadata` package. It's in the `models` folder and `models` is not packaged.

PR #58, opened 2022-05-30
Wrappers for bioinformatics data types

vladsaveliev:
Add classes for basic bioinformatics data types used in the pipelines, to help validating inputs for Hail Batch bioinformatics jobs. 
* CramPath
* GvcfPath
* FastqPath/FastqPair/FastqPairs
* AlignmentInput
* Also adds an enum SequencingType, matching the one of sample-metadata (see this PR against `initial` https://github.com/populationgenomics/production-pipelines/pull/57 for more context)

### DIFF
--- a/cpg_pipes/types.py
+++ b/cpg_pipes/types.py
@@ -1,0 +1,188 @@
+"""
+Basic bioinformatics status types.
+"""
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Union
+
+from . import Path, to_path
+from . import utils
+from hailtop.batch import ResourceGroup, ResourceFile, Batch
+
+logger = logging.getLogger(__file__)
+
+
+class CramPath:
+    """
+    Represents alignment data on a bucket within the pipeline.
+    Includes a path to a CRAM or a BAM file along with a corresponding index,
+    and a corresponding fingerprint path.
+    """
+
+    def __init__(self, path: str | Path, index_path: Path | str | None = None):
+        self.path = to_path(path)
+        self.is_bam = self.path.suffix == '.bam'
+        self.ext = 'cram' if not self.is_bam else 'bam'
+        self.index_ext = 'crai' if not self.is_bam else 'bai'
+        self._index_path = index_path
+        self.somalier_path = to_path(f'{self.path}.somalier')
+
+    def __str__(self) -> str:
+        return str(self.path)
+
+    def __repr__(self) -> str:
+        return f'CRAM({self.path})'
+
+    def exists(self) -> bool:
+        """
+        CRAM file exists.
+        """
+        return self.path.exists()
+
+    @property
+    def index_path(self) -> Path:
+        """
+        Path to the corresponding index
+        """
+        return (
+            to_path(self._index_path)
+            if self._index_path
+            else to_path(f'{self.path}.{self.index_ext}')
+        )
+
+    def resource_group(self, b: Batch) -> ResourceGroup:
+        """
+        Create a Hail Batch resource group
+        """
+        return b.read_input_group(
+            **{
+                self.ext: str(self.path),
+                f'{self.ext}.{self.index_ext}': str(self.index_path),
+            }
+        )
+
+
+class GvcfPath:
+    """
+    Represents GVCF data on a bucket within the pipeline.
+    Includes a path to a GVCF file along with a corresponding TBI index,
+    and a corresponding fingerprint path.
+    """
+
+    def __init__(self, path: Path | str):
+        self.path = to_path(path)
+        self.somalier_path = to_path(f'{self.path}.somalier')
+
+    def __str__(self) -> str:
+        return str(self.path)
+
+    def __repr__(self) -> str:
+        return f'GVCF({self.path})'
+
+    def exists(self) -> bool:
+        """
+        GVCF file exists.
+        """
+        return self.path.exists()
+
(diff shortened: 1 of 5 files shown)


cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+from . import Path, to_path
+from . import utils
+from hailtop.batch import ResourceGroup, ResourceFile, Batch
+
+logger = logging.getLogger(__file__)
+
+
+class CramPath:
+    """
+    Represents alignment data on a bucket within the pipeline.
```
illusional, line 19: Do all BAM / CRAMs have indexes? Is this a procedural guarantee? Are there ever cases where this might not be the case?

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+                'g.vcf.gz.tbi': str(self.tbi_path),
+            }
+        )
+
+
+FastqPath = Union[str, Path, ResourceFile]
+
+
+@dataclass
+class FastqPair:
```
illusional, line 113: Do we have a Single FASTQ case - this is relevant for ONT sequencing (RNA seq too?)

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+        r1 = (
+            self.r1 if isinstance(self.r1, ResourceFile) else b.read_input(str(self.r1))
+        )
+        r2 = (
+            self.r2 if isinstance(self.r2, ResourceFile) else b.read_input(str(self.r2))
+        )
+        return FastqPair(r1, r2)
+
+    def __str__(self):
+        return f'{self.r1}|{self.r2}'
```
illusional, line 138: Consider using `os.path.commonprefix` here, then you could do something like:

Not that this is directly correct either though:
```python
common_prefix = os.path.commonprefix(self.r1, self.r2)
return f'{common_prefix}*.fastq.gz'
```

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+class SequencingType(Enum):
+    """
+    Type (scope) of a sequencing experiment.
+    """
+
+    GENOME = 'genome'
+    EXOME = 'exome'
+    SINGLE_CELL = 'single_cell'
+    MTSEQ = 'mtseq'
+
```
illusional, line 153: I feel like I commented on a similar implementation, but have you considered extending the SequencingType from sample_metadata?

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+        str_val = str_val.lower()
+        if str_val not in str_to_val:
+            raise ValueError(
+                f'Unrecognised sequence type {str_val}. '
+                f'Available: {list(str_to_val.keys())}'
+            )
+        return str_to_val[str_val]
+
+
+class AlignmentInput:
```
illusional, line 179: This is for WGS/WES only though.

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        if isinstance(self.data, CramPath):
+            return utils.exists(self.data.path)
+        else:
+            return all(
+                utils.exists(pair.r1) and utils.exists(pair.r2) 
+                for pair in self.data
+            )
+
+    def compact_file_paths(self) -> str:
```
illusional, line 207: Could we refactor this out to it's own method, looks useful :D
A set of doctests for the future would be useful here too.

illusional (approved):
This PR as it stands is good to me, but I think we should consider more explicitly the difference on processing between WGS/WES/MTSeq, ONT, and RNA-Seq, because alignment inputs are not the same.

cpg_pipes/types.py
vladsaveliev: Yeah, I responded there. `models` is not a part of the `sample_metadata` package, so I can't import it here. Would need to refactor `sample_metadata` first

cpg_pipes/types.py
vladsaveliev: Oh, that's a handy function!

cpg_pipes/types.py
vladsaveliev: I haven't explored ONT or RNA yet. Another PR. I really want to avoid the PR branches to live for longer than a day, my brain just can't handle so many diverging codebases.

cpg_pipes/types.py
vladsaveliev: Responded in the other PR, but in the current pipelines, that's guaranteed. Maybe in the future I'll support other cases

cpg_pipes/types.py
vladsaveliev: I refactored it a bit in the other PR after your suggestions. Going to update it here now as well :)

cpg_pipes/types.py
vladsaveliev: Yeah, for now

cpg_pipes/types.py
```diff
@@ -0,0 +1,224 @@
+"""
+Wrappers for bioinformatics file types (CRAM, GVCF, FASTQ, etc).
+"""
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Union
+
+from . import Path, to_path
```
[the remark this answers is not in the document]
vladsaveliev: It works though. I might revert to normal imports in the future if it gives me any trouble

PR #65, opened 2022-06-03
Determine input datasets in input provider

vladsaveliev:
If datasets are not provided explicitly in the config, and the input provider type is CPG, the list of datasets would be pulled from seqr pulumi config. Essentially moving a chunk from seqr-loader into the CpgInputProvider to activate this for other pipelines as well (like QC, CRAM, etc)


cpg_pipes/providers/cpg/inputs.py
```diff
@@ -241,3 +238,21 @@ def populate_pedigree(self, cohort: Cohort) -> None:
                 f'{dataset.name}: found pedigree info for {len(samples_with_ped)} '
                 f'samples out of {len(dataset.get_samples())}'
             )
+
+    def requested_dataset_names(self) -> list[str] | None:
+        """
+        List of dataset names to process. Provides defaults for the
+        populate_cohort(datsets=[]) parameter.
```
lgruen, line 245:
```suggestion
        populate_cohort(datasets=[]) parameter.
```

cpg_pipes/providers/cpg/inputs.py
```diff
@@ -241,3 +238,21 @@ def populate_pedigree(self, cohort: Cohort) -> None:
...
+        """
+        # Parsing dataset names from the analysis-runner Seqr stack:
+        from urllib import request
+
+        seqr_stack_url = (
+            'https://raw.githubusercontent.com/populationgenomics/analysis-runner/main'
+            '/stack/Pulumi.seqr.yaml'
+        )
```
lgruen, line 253: @illusional Could we get this from the SM server instead? That might be a bit cleaner, as our deployment config isn't really meant to be consumed by pipelines.

lgruen (changes requested):

cpg_pipes/providers/cpg/inputs.py
illusional: This would be a good thing to address part of https://github.com/populationgenomics/sample-metadata/issues/220, we currently do not have a mechanism for determining this.

PR #67, opened 2022-06-09
Messaging to slack: second attempt

vladsaveliev:
Decoupling slack messaging from status reporter. Configuring through the global config.

Expects `slack` section like this:

```
[slack]
channel = 'seqr-loader-test'
token_secret_id = 'slack-seqr-loader-token'
token_project_id = 'seqr-308602'
```

If `channel` is specified, would check the secret provided for a token. 

Adds `slack_message_cmd(j, text)` function that adds a command to a job that sends `text` to the requested channel.

### DIFF
--- a/cpg_pipes/utils.py
+++ b/cpg_pipes/utils.py
@@ -97,18 +97,3 @@ def can_reuse(
 
     logger.debug(f'Reusing existing {path}. Use --overwrite to overwrite')
     return True
-
-
-def read_secret(project_id: str, secret_name: str) -> str:
-    """
-    Reads the latest version of a GCP Secret Manager secret.
-
-    Unlike cpg_utils.cloud.read_secret, crashes if the secret can not be read.
-    """
-
-    secret_manager = secretmanager.SecretManagerServiceClient()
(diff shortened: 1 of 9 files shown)


cpg_pipes/utils.py
```diff
@@ -97,3 +97,18 @@ def can_reuse(
 
     logger.debug(f'Reusing existing {path}. Use --overwrite to overwrite')
     return True
+
+
+def read_secret(project_id: str, secret_name: str) -> str:
+    """
+    Reads the latest version of a GCP Secret Manager secret.
+
+    Unlike cpg_utils.cloud.read_secret, crashes if the secret can not be read.
```
lgruen, line 106: Should we make `cpg_utils.cloud.read_secret` more configurable instead in terms of error handling, to avoid this duplication?

lgruen (approved):

cpg_pipes/utils.py
vladsaveliev: I'll do that!

vladsaveliev: https://github.com/populationgenomics/cpg-utils/pull/52 @lgruen

PR #69, opened 2022-06-20
Add additional filters

vivbak:


vivbak:
Hey @vladsaveliev , I know we discussed a different approach for handling https://github.com/populationgenomics/production-pipelines/issues/70, but I quickly put together this skeleton for a more generic filtering option. I think there may be other uses for filtering on a range of criteria (not just batch numbers) down the track, and this may be more extensible. 

Very keen to hear your thoughts. Also happy to throw it out if you don't like it.

cpg_pipes/pipeline/pipeline.py
```diff
@@ -815,11 +815,19 @@ def __init__(
             input_provider = CsvInputProvider(to_path(csv_path).open())
 
         if input_provider is not None:
+            # Pulls a set of samples to include based on a dict of filters. 
+            filters = get_config()['workflow'].get('sequence_filters')
```
vladsaveliev, line 819: Maybe call it `metadata` filters and filter based on anything in the `meta` dictionary? This way it can be generalised.

PR #71, opened 2022-06-26
Add 'reference_assembly' to bazam from SMDB

illusional:
I think it should look something like this, still need to test but easier to get thoughts like this first.

Based on: https://github.com/populationgenomics/sample-metadata/pull/242


cpg_pipes/types.py
```diff
@@ -93,9 +93,11 @@ class CramPath(AlignmentInput):
     """
 
     def __init__(
-        self, 
-        path: str | Path, 
+        self,
+        path: str | Path,
         index_path: Path | str | None = None,
+        # TODO: should this be required?
```
lgruen, line 99: I think it would be cleaner if we always require this, so we don't assume that the assembly used for realignment is identical to the one required for decoding the CRAM.

Does that mean we'd need to backfill the corresponding fields in the sample metadata server for our existing CRAMs though?

cpg_pipes/jobs/align.py
```diff
@@ -357,13 +375,36 @@ def _align_one(
...
+            then
+                echo -e "Background processes finished successfully"
+            else
+                # Background processes failed
+                trap 'error1' ERR
+            fi
+        """).strip()
+
+        # Now prepare command
+        cmd = '\n'.join([*fifo_pre, dedent(cmd).strip(), fifo_post])
```
vladsaveliev, line 404: That's awesome!!

cpg_pipes/providers/cpg/inputs.py
```diff
@@ -162,22 +164,23 @@ def populate_alignment_inputs(self, cohort: Cohort) -> None:
...
-                    f'\t{ds}, {len(list(samples))} samples: '
+                    f'\t{ds}, {len(list(samples))}/{ds_sample_count.get(ds)} samples: '
                     f'{", ".join([s.id for s in samples])}\n'
                 )
-            if get_config()['workflow'].get('smdb_errors_are_fatal', True):
-                raise InputProviderError(msg)
-            else:
-                msg += '\nContinuing without sequencing data because lenient=true'
-                logger.warning(msg)
+            # if get_config()['workflow'].get('smdb_errors_are_fatal', True):
```
vladsavelyev, line 179: Should we remove this flag and remove the commendet lines?

vladsavelyev:
Looks good to me, happy to merge!

cpg_pipes/providers/cpg/inputs.py
illusional: I've converted this to a `logger.info` call, removed the commented out lines, rephrased the message to also include the sequencing type + added a comment

vladsavelyev (approved):

PR #72, opened 2022-07-04
Add project name to populate analysis

illusional:


scripts/populate_sm_analysis.py
```diff
@@ -116,9 +141,11 @@ def _populate_qc_analysis_entries(multiqc_json_path: Path):
             analysis_type='joint-calling',
             analysis_status='completed',
             target=cohort,
-            meta=cohort.get_job_attrs() | dict(
+            meta=cohort.get_job_attrs()
+            | dict(
                 sequencing_type=cohort.sequencing_type.value,
             ),
+            project_name='seqr',
```
vivbak, line 148: Is this a remnant from testing?

illusional: Intentionally left here, joint-calling results should go into the `seqr` project (we joint-call across datasets)

cpg_pipes/providers/cpg/status.py
```diff
@@ -88,7 +88,8 @@ def create_analysis(
         analysis_type: str,
         analysis_status: str,
         target: Target,
-        meta: dict | None = None,        
+        meta: dict | None = None,
+        project_name: str=None,
```
vivbak, line 92: I think the linter is complaining here 
```suggestion
        project_name: str = None,
```

scripts/populate_sm_analysis.py
vivbak: ah, that makes sense :)

cpg_pipes/providers/cpg/status.py
illusional: Addressed

vivbak (approved):

PR #74, opened 2022-07-27
Fix loop-param-value-bind error

illusional:
Untested - how do we assess the impact of the incorrect loop value here?


vladsaveliev:
Great catch and solution!! 
Wondering how have you noticed that?

vladsaveliev (approved):

illusional:
It popped up in my flake8 linter :)

vladsaveliev:
Upstream is also affected: https://github.com/broadinstitute/seqr-loading-pipelines/blob/14bdeef02989d904087a537cf22dbbeb15e6594e/luigi_pipeline/lib/model/seqr_mt_schema.py#L241-L264 
I wonder if that's the reason the GQ filter was broken. Though doesn't explain why it was actually fixed in our fork 🤔  I'll investigate :)

vladsavelyev:
JFYI, I tested the previous code and worked correctly as well. Wonder if it depends on python version? Still, good to make the linter happy!

PR #75, opened 2022-07-28
Fix bug in alignment storage setting

vladsavelyev:


illusional (changes requested):
Can we add asserts + type annotations to the storage_for_cram_job definition to easier catch these?

vladsavelyev:
Passing named arguments should catch this. Will add type annotations too, yes

illusional (approved):
Awesome, thanks!

PR #78, opened 2022-07-28
Bulk updates from initial

vladsavelyev:
Propagating recent changes in `initial`:
* Moving storage provider to cpg-utils
* Use TOML configuration, remove CLI
* Add sequencing type
* Refactor process substitution
* Creating ES-index SMDB analysis
* Filter samples by sequencing type
* Various fixes


cpg_pipes/pipeline/pipeline.py
```diff
@@ -618,25 +620,34 @@ def _outputs_are_reusable(self, expected_out: ExpectedResultT) -> bool:
...
+                    _find_paths(el)
+            if isinstance(val, dict):
+                for el in val.values():
+                    _find_paths(el)
+
+        _find_paths(expected_out)
+
+        if paths and get_config()['workflow'].get('check_expected_outputs'):
+            # Checking that all paths in the expected output exist:
+            return all(exists(path) for path in paths)
```
lgruen, line 643: Not blocking, but just checking in terms of design: is the plan to move away from checking the existence of files and instead rely on metamist to determine the state of everything?
(@illusional FYI)

cpg_pipes/pipeline/pipeline.py
```diff
@@ -768,104 +775,73 @@ class Pipeline:
...
     ):
-        self.storage_provider = storage_provider
+        self._stages = stages
+        self.run_id = get_config()['workflow'].get('run_id', timestamp())
+
+        analysis_dataset = get_config()['workflow']['dataset']
+        name = get_config()['workflow'].get('name') or name
+        description = get_config()['workflow'].get('description') or description
+        name = name or description or analysis_dataset
+        self.name = slugify(name)
```
lgruen, line 789: This is nice, as it'll make the job searchable in the UI.

cpg_pipes/pipeline/pipeline.py
```diff
@@ -944,7 +916,9 @@ def set_stages(
         # implicit_stages_d: dict[str, Stage] = dict()  # If there are required
         # dependency stages that are not requested explicitly, we are putting them
         # into this dict as `skipped`.
+        depth = 0
         while True:  # Might need several rounds to resolve dependencies recursively.
```
lgruen, line 920: Should this include a check for cyclic dependencies?

lgruen (approved):
Very nice, thanks!

@illusional also had a few fixes recently (e.g. regarding process substitution and CRAM reference paths). Would be great to merge them into `main` soon as well, if it hasn't already happened.

cpg_pipes/pipeline/pipeline.py
vladsavelyev: I didn't plan to make metamist a hard dependency, but I'm open to this suggestion!

cpg_pipes/pipeline/pipeline.py
vladsavelyev: Not here. This code got has grown out of hand a bit and became too complicated. Essentially it's just topological sort here, and I have plans to switch to a more robust algorithm, but that's after I got everything in main.

cpg_pipes/pipeline/pipeline.py
vladsavelyev: I'm more open to relying on metamist more, since now it's very stable and has a GUI

PR #80, opened 2022-08-02
User highmem 2-core jobs for picard

vladsavelyev:
Tested resource allocation, but haven't run an actual job to completion.

Hail would provide 14G mem for 2 cores jobs, hope it should be enough!


illusional (approved):
Looks good for picard :)

vladsavelyev:
I'll run a test and then add then work on haplotype caller and others

PR #84, opened 2022-08-29
Add CRAM QC

vladsavelyev:
Adding CRAM QC batch jobs, largely replicating WARP exome and genome pipelines:
* `picard_collect_metrics`
* (for exomes) `picard_hs_metrics`
* (for genomes) `picard_wgs_metrics`
* `verifybamid` for contamination checks


MattWellie:
Not exactly part of this PR, but these imports are not grouped
https://github.com/populationgenomics/production-pipelines/blob/35e32870695f614c232b13069c5acda52fa5af36/jobs/picard.py#L10-L11

jobs/picard.py
```diff
@@ -65,3 +65,178 @@ def markdup(
...
+            out_quality_yield_metrics_path,
+        ],
+        overwrite,
+    ):
+        return None
+
+    job_attrs = (job_attrs or {}) | {'tool': 'picard_CollectMultipleMetrics'}
+    j = b.new_job('Picard CollectMultipleMetrics', job_attrs)
+    j.image(image_path('picard'))
+    res = STANDARD.request_resources(mem_gb=7)
```
MattWellie, line 100: No functional comment here, but 7gb seems like an odd reservation!

jobs/picard.py
```diff
@@ -65,3 +65,178 @@ def markdup(
...
+    )
+    b.write_output(j.out_insert_size_metrics, str(out_insert_size_metrics_path))
+    b.write_output(
+        j.out_quality_by_cycle_metrics, str(out_quality_by_cycle_metrics_path)
+    )
+    b.write_output(
+        j.out_base_distribution_by_cycle_metrics,
+        str(out_base_distribution_by_cycle_metrics_path),
+    )
+    b.write_output(j.out_quality_yield_metrics, str(out_quality_yield_metrics_path))
```
MattWellie, line 142: This seems less clean than creating a single resource group for the outputs, but I guess it's necessary if all the outputs need to end up on different paths

MattWellie (approved):
couple of questions, but looks good

jobs/picard.py
MattWellie: e.g. https://github.com/populationgenomics/rare-disease/blob/main/validation/validation_runner.py#L171-L194

jobs/picard.py
vladsavelyev: That's just directly borrowed from WARP :)

vladsavelyev: But good point, to make it clearer, I'll modify to request the minimal instance chunk instead with `res = STANDARD.request_resources(ncpu=2)`, it would correspond to ~7G

jobs/picard.py
vladsavelyev: Yeah, I thought about changing it to a resource group, but I wanted flexibility of writing to different paths, plus I find this a bit more clear personally because I always forget how resource groups actually work :)

MattWellie: Same 😆

PR #85, opened 2022-08-29
Add somalier and pedigree checks

vladsavelyev:
* Add job to extract [somalier](https://github.com/brentp/somalier) fingerprints from CRAM or GVCF files.
* Add job to infer pedigree and sex with somalier, and corresponding stage that generates an HTML report.
* Add script `check_pedigree.py` to check inferred pedigree against provided.
* Extract fingerprints as part of the `Align` stage.


jobs/python_scripts/check_pedigree.py
```diff
@@ -0,0 +1,335 @@
...
+
+            if (
+                expected_rel == 'unknown'
+                and inferred_rel != 'unknown'
+                or expected_rel == 'unrelated'
+                and inferred_rel != 'unrelated'
+            ):
+                mismatching_unrelated_to_related.append(line)
+            # elif inferred_rel in ['related at unknown level', 'unknown']:
+            #     pass
```
lgruen, line 263:
```suggestion
```

jobs/somalier.py
```diff
@@ -0,0 +1,364 @@
...
+from cpg_utils.workflows.resources import STANDARD
+from cpg_utils.workflows.slack import slack_message_cmd
+from cpg_utils.workflows.targets import Dataset, Sample
+from cpg_utils.workflows.filetypes import CramPath, GvcfPath
+from cpg_utils.workflows.utils import can_reuse, rich_sample_id_seds
+
+from .python_scripts import check_pedigree
+
+
+# We exclude contaminated samples from relatedness checks
```
lgruen, line 29: Is that because contaminated samples would show up as related to a number of other samples? Would be great to expand the comment a bit.

Should this be called `MAX_FREEMIX` instead, as it's an upper bound?

jobs/somalier.py
```diff
@@ -0,0 +1,364 @@
...
+
+    j.command(command(cmd))
+    # Copy somalier outputs to final destination.
+    b.write_output(j.output_samples, str(out_samples_path))
+    b.write_output(j.output_pairs, str(out_pairs_path))
+    b.write_output(j.output_html, str(out_html_path))
+    return j
+
+
+def extact(
```
lgruen, line 310:
```suggestion
def extract(
```

lgruen (approved):

jobs/somalier.py
vladsavelyev: Yeah, a contaminated sample can negatively affect relatedness estimation results for the entire dataset, so we need to make sure we don't pass those to `somalier relate`.

 I'll expand that in a comment!

PR #86, opened 2022-08-29
Add haplotype caller and `Genotype` stage

vladsavelyev:
Genotyping samples to generate per-sample GVCFs:

* Add job that generates intervals for variant calling parallelisation,
* Add job that runs GATK HaplotypeCaller in parallel based on intervals,
* Add job that post-processes GVCFs with [GATK ReblockGVCFs](https://gatk.broadinstitute.org/hc/en-us/articles/4405443600667-ReblockGVCF),
* Add corresponding `Genotype` stage.


lgruen (approved):
Thanks, this is nicely structured! Let's get this submitted as a snapshot of a state that's robustly working. However, I'd love to revisit this in a follow-up to see whether we can reduce the cost of running HaplotypeCaller significantly, e.g. by using fewer cores, making streaming from GCS work, etc. @illusional has a lot of ideas (and experience) with this, but it'll require a bunch of benchmarking and experiments.

PR #87, opened 2022-08-29
Add GVCF QC

vladsavelyev:
* Add `CollectVariantCallingMetrics` to call variant calling stats,
* Add `hap.py` to evaluate GVCFs for validation samples.
* Run QC as part of the `Genotype` workflow stage.


MattWellie:
I had a look through this before and it all looks good to me (and already works in the pipeline, so that's good!)
The main gripe I have (and it's shared with a lot of this workflow code being ported to cpg_utils) is the repetition of file names. I get that it may not be relevant very often, but could be clarify file names to better describe their function? e.g. with 2 or more somalier.py's:

```python
from cpg_utils.stages import somalier
from cpg_utils.jobs import somalier # :(
```

instead of having to do `from cpg_utils.jobs import somalier as somalier_jobs`, could we re-title `stage` files as `XXXX_stage.py` etc., and have the embedded classes self identified as SomalierStage vs. SomalierJob etc.?

I don't know if this is a useful observation based on how they'll be imported at runtime, but even going through code review I have to double check _which_ `somalier.py` I'm looking at 😆

jobs/happy.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+import logging
+
+from cpg_utils import Path
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import reference_path, image_path, fasta_res_group
+from hailtop.batch import ResourceGroup
+
+from cpg_utils.hail_batch import command
```
MattWellie, line 12: un-grouped import from cpg_utils.hail_batch

MattWellie (approved):

vladsavelyev:
@MattWellie, I replied to you on that question in other PR, e.g.: https://github.com/populationgenomics/production-pipelines/pull/92#issuecomment-1231179719

Basically, structuring a project always involves trade-offs, so I'm a bit hesitant to change it. The parent folder name shows which `somalier.py` it is, and wrt importing, you don't usually need to import a somalier "stage" and a somalier "job" in the same module. 

But I'd be happy for you to use the project with fewer friction, so if you think it would help, I can change the file names :-)

MattWellie:
@vladsavelyev sorry, I didn't see that comment! 

If the usage of these components doesn't really overlap/is always in context then it's not a problem, I was more thrown by the image of a PR adding 3 scripts with the same name 😆 

Screenshot 2022-09-02 at 1 19 35 pm

I don't think my criticism is necessarily a good one, and within the project structure it isn't going to be confusing, so I'm happy to approve

vladsavelyev:
@MattWellie, I'll think about changing the structure in a follow-up PR

PR #90, opened 2022-08-29
Add VEP

vladsavelyev:
VEP functions, pretty much the same code as in AIP.

* Add jobs to run VEP in parallel, 
* Add Job to gather VEP results into a hail `Table` using Query Batch backend, 
* Add `Vep` pipeline stage.

I'm thinking of removing that here and just import functions from AIP (WIP https://github.com/populationgenomics/automated-interpretation-pipeline/pull/107), but before that I want this merged into main.


MattWellie:
Without thinking about it too much, I quite like the idea of removing the annotation code from AIP and importing from `cpg-utils` 🙃

jobs/vep.py
```diff
@@ -0,0 +1,217 @@
...
+    b: Batch,
+    vcf_path: Path,
+    tmp_prefix: Path,
+    out_path: Path | None = None,
+    overwrite: bool = False,
+    scatter_count: int = 50,
+    job_attrs: dict | None = None,
+) -> list[Job]:
+    """
+    Runs VEP on provided VCF. Whites a VCF into `out_path` by default,
```
lgruen, line 40:
```suggestion
    Runs VEP on provided VCF. Writes a VCF into `out_path` by default,
```

jobs/picard.py
```diff
@@ -37,42 +42,47 @@ def get_intervals(
     that, but Hail Batch is not dynamic and have to expect certain number of output
```
lgruen, line 42: In line 41 (GitHub doesn't let make a suggestion there), "unpredicted" should probably be "unpredictable".

jobs/vep.py
```diff
@@ -0,0 +1,217 @@
...
+        'conservation_file': f'{vep_dir}/loftee.sql',
+    }
+
+    authenticate_cloud_credentials_in_job(j)
+    cmd = f"""\
+    ls {vep_dir}
+    ls {vep_dir}/vep
+
+    LOFTEE_PLUGIN_PATH=/root/micromamba/share/ensembl-vep-105.0-1
+    FASTA={vep_dir}/vep/homo_sapiens/105_GRCh38/Homo_sapiens.GRCh38.dna.toplevel.fa.gz
```
lgruen, line 189: Should the VEP version be parameterized?

lgruen (approved):

PR #91, opened 2022-08-29
Add MultiQC

vladsavelyev:
* Add jobs that gather CRAM QC and GVCF QC outputs into MultiQC reports.
* Add script that checks MultiQC stats and sends Slack messages for failed metrics.
* Add `MultiQC` stage.


jobs/multiqc.py
```diff
@@ -0,0 +1,197 @@
...
+
+    report_filename = 'report'
+    cmd = f"""\
+    mkdir inputs
+    cat {file_list} | gsutil -m cp -I inputs/
+    
+    # Temporary fix for Somalier module before https://github.com/ewels/MultiQC/pull/1670
+    # is merged and released:
+    git clone --single-branch --branch fix-somalier https://github.com/vladsaveliev/MultiQC
+    pip3 install -e MultiQC
```
lgruen, line 96: Is this still necessary? I've noticed that this PR got merged, but not sure whether it also got released.

lgruen (approved):

jobs/multiqc.py
vladsavelyev: It's not released yet

PR #92, opened 2022-08-29
Add matrix table annotation

vladsavelyev:
Add jobs based on Hail Query with a Batch backend, to:
* load joint-called VCF into matrix table,
* annotate with VQSR and VEP results,
* Split resulting matrix table per dataset,
* Annotate datasets with dataset-specific stats,
* Add corresponding stages `AnnotateCohort` and `AnnotateDataset`.


MattWellie:
can we rename these files to differentiate them? e.g. seqr_loader_stages/seqr_loader_methods/seqr_loader_jobs

jobs/query_scripts/seqr_loader.py
```diff
@@ -0,0 +1,429 @@
...
+    logging.info('Adding VQSR annotations into the Matrix Table')
+    mt = mt.annotate_globals(**vqsr_ht.index_globals())
+    mt = mt.annotate_rows(
+        # vqsr_ht has info annotation split by allele, plus the new AS-VQSR annotations
+        info=vqsr_ht[mt.row_key].info,
+        filters=mt.filters.union(vqsr_ht[mt.row_key].filters).filter(
+            lambda val: val != 'PASS'
+        ),
+    )
+    mt = _checkpoint(mt, 'mt-vep-split-vqsr.mt')
```
MattWellie, line 196: how do you feel about making the VQSR load/annotation an optional action, so that the same code can be shared by AIP and cpg_utils?

I think this could be very simple if `site_only_vqsr_vcf_path` is optional, and [this code](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/reanalysis/annotation.py#L112-L113) is added here (for non-GATK VCFs to be annotated)

MattWellie: Happy to have a go at this if you think it's possible, it would allow removal of redundancy between this and AIP

MattWellie (dismissed):

vladsavelyev:
Currently the structure looks like this: 

```
jobs/
jobs/seq_loader.py
jobs/vep.py
...
stages/
stages/seq_loader.py
stages/vep.py
...
```

Do you think it's not clear? It's a bit tricky to structure it because it's a sacrifice no matter how you do that. I could rename it like this, but it looks a bit repetitive:

```
jobs/
jobs/seq_loader_jobs.py
jobs/vep_jobs.py
...
stages/
stages/seq_loader_stage.py
stages/vep_stage.py
...
```

Alternatively I could make the structure completely flat.

jobs/query_scripts/seqr_loader.py
vladsavelyev: Yes, I'm all for sharing code with AIP, just with all refactoring always kept pushing this back. I'll do that in this PR :)

lgruen (approved):

jobs/query_scripts/seqr_loader.py
```diff
@@ -0,0 +1,429 @@
...
+    #   gs://cpg-seqr-main-tmp/seqr_loader/v0/AnnotateCohort/seqr_loader/checkpoints/vqsr.ht
+    #   ht = hl.filter_intervals(ht, [hl.parse_locus_interval('chrX:52729395-52729396')])
+    # Hail's hl.float() correctly parses this value, however, seqr loader doesn't
+    # recognise "Inf" and "Nan" values, so we need to replace it with zero.
+    # hl.is_finite() returns False for "Inf" and "NaN".
+    # Example of NaN: https://batch.hail.populationgenomics.org
+    # .au/batches/6973/jobs/12
+
+    def _fix_inf(x):
+        return hl.if_else(hl.is_finite(x), x, 0.0)
```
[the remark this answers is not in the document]
vladsavelyev: https://github.com/populationgenomics/production-pipelines/issues/98

PR #93, opened 2022-08-29
Load MT into ES

vladsavelyev:
* Add dataproc script to load matrix table into ElasticSearch for seqr. Depends on the [elasticsearch config section](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config-template.toml#L107-L114) and corresponding secret.
* Add `MtToEs` stage that drives that script.


jobs/dataproc_scripts/mt_to_es.py
```diff
@@ -0,0 +1,695 @@
+#!/usr/bin/env python3
+
+"""
+Hail script to submit on a dataproc cluster. 
+
+Loads the matrix table into an ElasticSearch index.
```
lgruen, line 6:
```suggestion
Loads the matrix table into an Elasticsearch index.
```

jobs/dataproc_scripts/mt_to_es.py
```diff
@@ -0,0 +1,695 @@
...
+        secret_name=secret_name,
+        fail_gracefully=False,
+    )
+    assert password
+
+    print(
+        f'Connecting to ElasticSearch: host="{host}", port="{port}", user="{username}"'
+    )
+    print(f'Reading passport from secret "{secret_name}" in project "{project_id}"')
+    # def __init__(self, host='localhost', port='9200', es_username='pipeline', es_password=None, es_use_ssl=False):
```
lgruen, line 88:
```suggestion
```

jobs/dataproc_scripts/mt_to_es.py
```diff
@@ -0,0 +1,695 @@
...
+    es = HailElasticsearchClient(
+        host=host,
+        port=port,
+        es_username=username,
+        es_password=password,
+        es_use_ssl=(host != 'localhost'),
+    )
+
+    mt = hl.read_matrix_table(mt_path)
+    # Annotate GRCh37 coordinates here, as they are not supported by Batch Backend
```
lgruen, line 98: Do you mean that _liftover_ is not supported by Batch?

stages/seqr_loader.py
```diff
@@ -113,3 +113,74 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
...
+
+        j = dataproc.hail_dataproc_job(
+            self.b,
+            f'jobs/dataproc_scripts/mt_to_es.py '
+            f'--mt-path {dataset_mt_path} '
+            f'--es-index {index_name} '
+            f'--es-password {es_password()} '
+            f'--liftover-path {reference_path("liftover_38_to_37")} '
+            f'--use-spark ',  # es export doesn't work with the service backend
+            max_age='24h',
```
lgruen, line 167: Are we getting close to this limit?

stages/seqr_loader.py
```diff
@@ -113,3 +113,74 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
...
+            packages=[
+                'cpg_utils',
+                'elasticsearch==8.*',
+                'google',
+                'fsspec',
+                'sklearn',
+                'gcloud',
+                'selenium',
```
lgruen, line 175: Do we need `sklearn` / `selenium` here?

jobs/dataproc_scripts/mt_to_es.py
```diff
@@ -0,0 +1,695 @@
...
+            logging.warning(
+                'Waiting for {} shards to transfer off the es-data-loading nodes: \n{}'.format(
+                    len(shards.strip().split('\n')), shards
+                )
+            )
+            time.sleep(5)
+
+        raise Exception('Shards did not transfer off loading nodes')
+
+    def export_table_to_elasticsearch(
```
lgruen, line 372: Is this mostly a copy from the `hail-elasticsearch-pipelines` code? Would it be too painful to import that instead of duplicating it here? Is there any risk this might diverge from upstream?

lgruen (approved):

stages/seqr_loader.py
vladsavelyev: Not there yet, but loading acute-care is ~11 hours.

stages/seqr_loader.py
vladsavelyev: I can't recall, let's try without and add a comment if it's needed after all

jobs/dataproc_scripts/mt_to_es.py
vladsavelyev: Yeah, and I used to import it, but it felt that maintaining a fork of https://github.com/populationgenomics/hail-elasticsearch-pipelines with just added `setup.py` that builds a package with modules named `hail_scripts` and `lib`, and maintaining that package, is not ideal either. But agree that it would make it harder if we diverge, I'll think if I can maybe get away with a git submodule

PR #94, opened 2022-09-06
Add `main.py` and toml configs

vladsavelyev:
Putting stages together by adding `main.py`, and toml configs for different scenarios.


configs/seqr-main.toml
```diff
@@ -0,0 +1,38 @@
+[workflow]
+access_level = 'full'
+datasets = [
```
lgruen, line 3: @illusional I wonder if we could also reflect these dependent datasets in metamist?

configs/seqr-main.toml
```diff
@@ -0,0 +1,38 @@
+[workflow]
+access_level = 'full'
```
lgruen, line 2: This should typically only be set through the analysis-runner, right?

```suggestion
```

main.py
```diff
@@ -0,0 +1,32 @@
...
+@click.command()
+@click.argument('config_paths', nargs=-1)
+def main(config_paths: list[str]):
+    """
+    Run a workflow, using CONFIG_PATHS in the order specified, overriding
+    $CPG_CONFIG_PATH if specified.
+    """
+    if _cpg_config_path_env_var := os.environ.get('CPG_CONFIG_PATH'):
+        config_paths = _cpg_config_path_env_var.split(',') + config_paths
+    set_config_paths(list(config_paths))
```
lgruen, line 27: Is this meant for tests only? I'm asking because I thought we'd typically want users to rely on the `--config` parameter in the analysis-runner to invoke workflows.

lgruen (approved):

PR #99, opened 2022-09-14
Add unit tests and CI

vladsavelyev:
* Add a dry test of the seqr loader workflow: `test_workflow_dry.py`
* Add test of parsing and checking QC: `test_check_multiqc.py`
* Add test of parsing and checking pedigree: `test_check_pedigree.py`
* Add Github CI

TODO in new PRs:
* Add individual jobs tests
* Add query code tests (annotation, vep)


test/test_check_multiqc.py
```diff
@@ -0,0 +1,53 @@
...
+"PERCENT_DUPLICATION" = 25
+"""
+
+
+@pytest.fixture()
+def tmp_dir() -> Path:
+    dir_path = to_path('results') / timestamp()
+    dir_path.mkdir(parents=True, exist_ok=True)
+    yield dir_path
+    shutil.rmtree(dir_path)
```
MattWellie, line 32: isn't this just replicating pytest's `tmpdir` fixture?

vladsavelyev: Oh, didn't know about it! Thanks a lot @MattWellie  :D

MattWellie: It's my favourite default fixture ❤️

vladsavelyev: The only problem is that Hail's local backend doesn't work well with `/tmp`

vladsavelyev: @MattWellie while you are here, would you please click "approve"?

MattWellie: sample-metadata was removed from the requirements, which is killing the test runs.. should that be added back in?

vladsavelyev: Oh, it actually should be `cpg-utils[workflows]` that installs `sample-metadata` now

MattWellie (approved):

test/test_check_multiqc.py
vladsavelyev: `cpg-utils` needs to be bumped :) https://github.com/populationgenomics/cpg-utils/pull/76

vladsavelyev: Alright, tests are passing now, and the standard `tmpdir` fixture works here! Thanks @MattWellie :)

PR #100, opened 2022-09-17
Add the Broad's seqr-loading-pilelines as submodule

vladsavelyev:
Some re-work of how the Broad seqr-loader query code is run.

1. Adding the Broad's [seqr-loading-pilelines](https://github.com/broadinstitute/seqr-loading-pipelines) as submodule and send to dataproc with `--pyfiles` (instead of building a package or keeping a copy of code). Adjust for Elasticsearch v8. Note that for IDE to resolve the imports from `seqr-loading-pilelines` (like `hail_scripts` imported in `mt_to_es.py`, as well as within the `hail_scripts` python modules themselves), `seqr-loading-pilelines` has to be marked as a source root (PyCharm: `Mark as: Sources`, or PYTHON_PATH should be set explicitly, depending on the IDE).
2. Moving `query_scripts`, `python_scripts`, `dataproc_scripts` from `jobs` into the repo root level. Rename `query_scripts`->`query_module`. The reason is to be able to pass `query_modules` to dataproc with `--pyfiles` without passing entire `jobs` module, and avoiding setting extra source root levels in PyCharm (like for `seqr-loading-pilelines`).
3. Adding dataproc scripts `annotate_cohort.py` and `annotate_dataset.py`, and use them by default instead of running `query_modules` functions using Batch Backend. Batch Backend is harder to debug, but they are there to be swapped back quickly. Dataproc scripts are still importing from `query_modules`.
4. Annotation: drop `AS-*` annotations (causing troubles when splitting multiallelics, and they are used only to run AS-VQSR anyway, and not needed after FILTERS are populated from AS-VQSR).


MattWellie (approved):
LGTM, but I'm happy to check for other opinions

vladsavelyev:
Thanks for approving @MattWellie ^_^

PR #101, opened 2022-09-22
Exome validation

vladsavelyev:
Exome validation: share truth data between genomes and exomes, but pass intervals with `--target-regions`.

Note that the metric won't be accurate because the exome evaluation regions are generic, and do not represent actual capture regions used for the experiment. In the future, we could pass actual regions and baits as well.

### DIFF
--- a/jobs/happy.py
+++ b/jobs/happy.py
@@ -25,70 +25,81 @@ def happy(
     """
     Run hap.py validation/concordance stats on a sample GVCF or joint VCF.
     """
-    if 'validation' not in get_config()['workflow']:
+    if 'validation' not in get_config():
         logging.warning(
             'workflow/validation section is not defined in config, skip running hap.py'
         )
         return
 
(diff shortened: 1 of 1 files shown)


jobs/happy.py
```diff
@@ -25,70 +25,81 @@ def happy(
...
-    sequencing_type = get_config()['workflow']['sequencing_type']
-    if sequencing_type not in get_config()['workflow']['validation']:
-        logging.warning(
-            f'workflow/validation/{sequencing_type} section is not defined in config, skip running hap.py'
-        )
-        return
-
-    truth_sample_id = get_config()['workflow']['validation'].get(sample.external_id)
+    truth_sample_id = (
+        get_config().get('validation', {}).get('sample_map').get(sample.participant_id)
```
MattWellie, line 35:
```suggestion
        get_config().get('validation', {}).get('sample_map', {}).get(sample.participant_id)
```

MattWellie (approved):

PR #103, opened 2022-09-22
Add exome capture QC

vladsavelyev:
Prepare exome evaluation intervals and call `picard CollectHsMetrics`. Note that the metric won't be accurate because the exome evaluation regions are generic, and do not represent actual capture regions used for the experiment. In the future, we could pass actual regions and baits as well.


MattWellie (approved):
looks tricky!

PR #106, opened 2022-09-28
Validation stage: return `None` for samples with no truth data

vladsavelyev:
Make the variant validation stage (that drives `hap.py`) return `None` as the `expected_output` for samples with no validation data. So the `MultiQC` stage can use that to filter down to the samples for which the validation data is available.


vivbak (approved):

vladsavelyev:
Sorry, forgot to merge with `main` again after previous PRs. The PR should be cleaner now.

PR #120, opened 2022-10-03
Set logging level explicitly in executable python scripts

vladsavelyev:
Fixes unit tests.

### DIFF
--- a/python_scripts/check_multiqc.py
+++ b/python_scripts/check_multiqc.py
@@ -22,6 +22,10 @@ from cpg_utils import to_path
 from cpg_utils.config import get_config
 
 
+logging.basicConfig()
+logging.getLogger().setLevel(logging.DEBUG)
+
+
 @click.command()
 @click.option(
     '--multiqc-json',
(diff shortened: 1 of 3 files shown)


python_scripts/check_multiqc.py
```diff
@@ -22,6 +22,10 @@
 from cpg_utils.config import get_config
 
 
+logging.basicConfig()
+logging.getLogger().setLevel(logging.DEBUG)
```
lgruen, line 26: You can combine these statements:

```py
logging.basicConfig(level=logging.DEBUG)
```

Do you think it's worth instantiating a logger here instead of using the default one? Often, `logger = logging.getLogger(__name__)` or similar is used at the top level (see https://docs.python.org/3/library/logging.html#logger-objects).

lgruen (approved):

python_scripts/check_multiqc.py
vladsavelyev: > You can combine these statements:

Unfortunately, this is not an equivalent statement.`logging.basicConfig(level=logging.DEBUG)` sets the global level, and doesn't alter the root logger instance that would be instantiated by this point. I know that it's super confusing. See also https://stackoverflow.com/a/38537983

vladsavelyev: > Do you think it's worth instantiating a logger here instead of using the default one?

I think using the root logger is just simpler, since I don't need multiple loggers, and also it's a small isolated script that is always run within a Batch job.

PR #129, opened 2022-10-09
Use `driver_image` instead of `images/hail` and `images/sm-api`

vladsavelyev:
Follow up to https://github.com/populationgenomics/cpg-utils/pull/96

Also, version bump didn't work last time for some reason.


lgruen (approved):
I actually kind of like the `hail` image path, as presumably this would also be run outside of the CPG, where the concept of a "driver image" doesn't make that much sense? So should we instead set a `hail` image tag in the analysis-runner instead?

PR #136, opened 2022-10-13
Use `cpg_utils` image instead of driver image for python/hail jobs

vladsavelyev:


MattWellie (approved):
This will now be dependent on cpg-utils>=4.8.16

PR #141, opened 2022-10-18
Add bioheart config

KatalinaBobowik:
Add a config file for bioheart-test in order for the analysis team to test and have help walking through the large cohort pipeline workflow.

### DIFF
--- a/configs/bioheart-test.toml
+++ b/configs/bioheart-test.toml
@@ -1,0 +1,16 @@
+[workflow]
+dataset = 'bioheart'
+input_datasets = ['bioheart']
+dataset_gcp_project = 'bioheart'
+sequencing_type = 'genome'
+access_level = 'test'
+
+check_inputs = true
+check_expected_outputs = true
+# leave this to true, since one sample doesn't have sequencing data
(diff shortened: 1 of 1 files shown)


configs/bioheart-test.toml
```diff
@@ -0,0 +1,16 @@
+[workflow]
+dataset = 'bioheart'
+input_datasets = ['bioheart']
+dataset_gcp_project = 'bioheart'
+sequencing_type = 'genome'
+access_level = 'test'
```
vladsavelyev, line 6: Looks nice! Just a note that the following options are not necessary if you are running with the analysis-runner: `dataset`, `dataset_gcp_project`, `access_level`, `billing_project`. Also, `check_inputs` and `check_expected_outputs` are true by default, so you can leave them out, or keep if you want :)

vladsavelyev (approved):

configs/bioheart-test.toml
KatalinaBobowik: Ok excellent, thanks! That's good to know the analysis runner populates all of that data for you (I've deleted them now).

PR #145, opened 2022-10-20
Move default configs

vladsavelyev:
Move the TOML configs that loaded by default into `configs/defauts`.


vivbak (approved):
Oh thats cool I didnt realise the docs on the config existed!

PR #155, opened 2022-10-26
Refactor into `cpg_workflows` installable package

vladsavelyev:
Refactor the codebase into a installable/importable package called `cpg_workflows`, so jobs can be imported from e.g. AIP. Doesn't change anything functionally, just moving files around.

If we make it a package, perhaps I could also move the `cpg_utils/workflows` module into this repo as well.


configs/thousand-genomes.toml
```diff
@@ -6,6 +6,7 @@ scatter_count = 20
 first_stages = ['Combiner']
 last_stages = ['LoadVqsr', 'Frequencies']
 output_version = '0.1'
+skip_qc = true
```
lgruen, line 9: Was this intentional?

setup.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+
+from setuptools import find_packages, setup
+
+setup(
+    name='cpg-workflows',
+    # This tag is automatically updated by bumpversion
+    version='0.1.0',
+    description='CPG workflows for Hail batch',
```
lgruen, line 9:
```suggestion
    description='CPG workflows for Hail Batch',
```

lgruen (approved):
Moving the `cpg_utils/workflows` module into this repo as well sounds great to me, if it's not too much work!

configs/thousand-genomes.toml
vladsavelyev: fixed

setup.py
vladsavelyev: fixed

MattWellie (approved):
exciting!

PR #156, opened 2022-10-27
Make markdup-metrics optional, but use `samtools stats` for dup rate

vladsavelyev:
Currently, we run CRAM QC as part of the `Align` stage. Ideally, we want to make QC a separate stage, so some missing QC files don't force the alignment stage. Some older CRAMs (e.g. thousand-genomes, acute-care) are missing some QC files.

The reason we made QC part of the `Align` stage originally is that we use markdup metrics that are generated by the same tool that runs the deduplication itself (namely `picard MarkDuplicate`, or optionally `biobambam`).

As a solution, we can instead use the dup-rate metic generated by `samtools stats`, and we can make `picard MarkDuplicate` stats an optional outputs for more details if we ever need them (but in common scenarios we never look beyond simple dup-rate ratio).

Also, as part of this refactoring, moving all CRAM QC functions into `stages/cram_qc.py`, and similarly all GVCF QC into `stages/gvcf_qc.py`.


cassimons (commented):
LGTM. After looking through these stages yesterday I really like the simplicity that pulling the QC out to its own stages brings thanks.

cassimons (approved):

PR #158, opened 2022-10-27
Allow all stages as allowed values for `last_stages` and `first_stages`

vladsavelyev:
Allow all stages as allowed values for `last_stages` and `first_stages`.

`last_stages` and `first_stages` options are a bit tricky because the stages graph is not linear anymore: for example, it's not obvious whether to run `JointGenotyping` when `last_stages = ['CramQC']`, because `JointGenotyping` is not a direct descendant of `CramQC`. I decided to go with the logic of skipping `JointGenotyping` in this case. So if `last_stages = ['CramQC']` is set, then only `Align` and `CramQC` would be triggered.

Down the road, we need a better test to cover different combinations of `first_stages`, `only_stages`, etc. Or maybe rework the whole logic.


lgruen (changes requested):
> last_stages and first_stages options are a bit tricky because the stages graph is not linear anymore

Yes, that's an interesting problem. Really we'd want a graph cut specified in the options, but that's not easy to explain either :).

How about a simpler model, where one explicitly lists the complete subset of stages to run? (And by default, everything is run.) If some of those specified stages have unmet dependencies, that should ideally trigger an error (unless we want to be able to "force-run" anyway).

As long as our pipelines aren't huge, I think that's conceptually easier to understand than a somewhat ill-defined "first" and "last" in a DAG.

vladsavelyev:
Good suggestion, there are some considerations though. If we add new stages into a workflow, people will forget to add them to that "list". Should the workflow fail immediately if some intermediate stages are not listed?
But I'm happy to think about a simpler system, I was planning to re-work it anyway.

That aside, did you mean to block this PR? Should such requests go as side issues?

cassimons:
Listing all stages sounds painful.

Rather than list all stages, why follow the overall logic and only list the endpoints that you want to run. So if `last_stages = ['CramQC']` then only `CramQC` and its dependencies will run. If  `last_stages = ['CramQC', 'JointGenotyping']`  then `JointGenotyping` and its dependencies will also run.

It would also be helpful if we could locally run the script with a `--endpoints` flag or similar to see the available list for this workflow to select from.

lgruen:
> That aside, did you mean to block this PR? Should such requests go as side issues?

Sorry, certainly didn't mean to block this -- your PR description just sounded like you were looking for input on what the desired semantics should be like and I assumed that was in scope for discussion before merging. That's why I used the "request changes" review option, as that'll remove the PR from my queue until you re-request a review. If you want to merge this first and then change the semantics afterwards, that's of course fine too. Looks like some recent changes conflict with this branch currently though.

vladsavelyev:
> It would also be helpful if we could locally run the script with a --endpoints flag or similar to see the available list for this workflow to select from.

Yeah, that would be helpful! I'll add something like that

vladsavelyev:
@lgruen, sorry, I wasn't clear in the description indeed. But for the future, let's just decide that for discussions, we want to start an issue or a draft PRs; but if it's a ready-to-review PR, it is assumed that the author wants the change to go in.

I think that a most common response to a small pull request should be to either approve, or close it. It's getting painful quickly to navigate in multiple unmerged PRs. As you pointed, there are merge conflicts already. If we want to encourage our policy of tiny and targeted pull requests, we should try to use the "request changes" button less often

vladsavelyev:
> Looks like some recent changes conflict with this branch currently though.

@lgruen, I resolved conflicts now! (though I thought you should be able to click "approve" even if there are conflicts?)

vladsavelyev:
@lgruen closing to avoid stale PRs

MattWellie (approved):
Makes sense, could/should this have a directionality check to make sure that every last stage is a descendant of at least one first stage if both are supplied? Please correct me if this is wrong... 

1. First stages and Last stages are always kept
2. Everything after `last` and before `first` is skipped
3. If this was configured wrongly and first stage/last stage were swapped, every middle stage of the pipeline would be marked to skip and only the listed first/last stages would run

vladsavelyev:
Thank @MattWellie  for taking a look at it and for approving! :)

To answer your questions, 1 and 2 are correct and as intended, but 3 is not ideal. I'll add some test case and an error message for this scenario, thank you! Will also add some tests.

vladsavelyev:
Update: added tests for different config scenarios in https://github.com/populationgenomics/production-pipelines/pull/226 Scenario 3 in your example actually works correctly, raising a `WorkflowError("no stages to run")` :)

lgruen:
Looks like these changes have been merged already? Should we close this PR?

PR #161, opened 2022-10-28
Use bcftools==1.16 that supports `--tmp-dir`

vladsavelyev:
Gathering VCFs [doesn't work](https://batch.hail.populationgenomics.org.au/batches/378600/jobs/302) for large datasets because `bcftools` attempts to write to `/tmp`, which doesn't use the attached disk in the Batch job environment. [Newer](https://github.com/samtools/bcftools/issues/1642) `bcftools` version does have `--tmp-dir` option so we can redirect tmp location to `$BATCH_TMPDIR`.

Test: https://batch.hail.populationgenomics.org.au/batches/378615


vivbak (approved):
Looks like the relevant jobs were successful :)

PR #168, opened 2022-11-02
add rdp-kidney to seqr-main  config

cassimons:


configs/seqr-main.toml
```diff
@@ -20,6 +20,7 @@ input_datasets = [
     'ag-hidden',
     'flinders-ophthal',
     'udn-aus-training',
+    'rdp-kidney'
```
vladsavelyev, line 23:
```suggestion
    'rdp-kidney',
```

vladsavelyev (approved):

configs/seqr-main.toml
cassimons: good catch thanks.

PR #169, opened 2022-11-02
Improve dry run

vladsavelyev:
Making `dry_run` more consistent:

* `hail/dry_run = true` would only be directly propagated as `batch.run(dry_run=True)`, in consistency with other `[hail]` options. So when it's set, we will run all the workflows code, but `batch.run()` will only print commands instead of actually creating any batches and jobs.
* `workflow/dry_run = true` however is more limiting: it should allow the code to be run without permissions, so it will skip any communication with Metamist, Hail Batch, and Cloud Storage, and will only print the final merged config and a list of stages to be submitted.

To support it, a few changes added:
* dropping `Stage.b` and `Stage.cohort` fields in favour of `get_batch()` and `get_cohort()` (so we don't need to initialise Cohort and Batch when only initialising a Workflow), 
* mocking `get_config()` in tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,7 +493,7 @@ For example, out "Align" stage that runs BWA might look like as following:
 ```python
 from cpg_utils import Path
 from cpg_workflows.workflow import stage, SampleStage, Sample, StageInput, StageOutput
-
+from cpg_workflows.batch import get_batch
 
 @stage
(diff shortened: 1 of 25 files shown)


vivbak (approved):
I like this :)

PR #170, opened 2022-11-02
Add cmdline options to main.py: `--list-workflows`, `--list-last-stages`, `--dry-run`, `--help`

vladsavelyev:
Useful to quickly testing your config before submitting with analysis-runner.

```
$ python main.py
Usage: main.py [OPTIONS] [WORKFLOW]

  Run a Hail Batch workflow specified as a positional command line argument
  [WORKFLOW]

Options:
  --config PATH       Add configuration files to the files specified
                      $CPG_CONFIG_PATH.Configs are merged left to right,
                      meaning the rightmost file has thehighest priority.
  --list-workflows    Only list possible values for WORKFLOW (and available
                      last stages)
  --list-last-stages  Only list possible end stages for a workflow, that can
                      be specified with `workflow/last_stages` in config
  --dry-run           Dry run: do not actually communicate with Metamist or
                      Hail Batch, instead only print a final config and stages
                      to be run
  --help              Show this message and exit.
```

```
$ python main.py --list-workflows
Available values for WORKFLOW (and corresponding last stages):
	pre_alignment (FastQCMultiQC)
	seqr_loader (AnnotateDataset, MtToEs, GvcfMultiQC, CramMultiQC)
	large_cohort (LoadVqsr, Frequencies, GvcfMultiQC, CramMultiQC)
	gatk_sv (GatherBatchEvidence)
```

```
$ python main.py seqr_loader --list-last-stages
Available last stages that can be listed with workflow/last_stages for the workflow "seqr_loader":
AnnotateDataset, MtToEs, GvcfMultiQC, CramMultiQC
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 """
-Entry point to run the workflow.
+Entry point to run workflows.
 """
 import os
 
@@ -15,7 +15,7 @@ from cpg_workflows.stages.large_cohort import LoadVqsr, Frequencies
 from cpg_workflows.stages.cram_qc import CramMultiQC
 from cpg_workflows.stages.gvcf_qc import GvcfMultiQC
 from cpg_workflows.stages.fastqc import FastQCMultiQC
-from cpg_workflows.stages.seqr_loader import MtToEs
+from cpg_workflows.stages.seqr_loader import MtToEs, AnnotateDataset
 from cpg_workflows.stages.gatk_sv import GatherBatchEvidence
 
 fmt = '%(asctime)s %(levelname)s (%(name)s %(lineno)s): %(message)s'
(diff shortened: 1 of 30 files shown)


main.py
```diff
@@ -24,20 +24,71 @@
...
+    'list_workflows',
+    is_flag=True,
+    help='Only list possible values for WORKFLOW (and available last stages)',
+)
+@click.option(
+    '--list-last-stages',
+    'list_last_stages',
+    is_flag=True,
+    help='Only list possible end stages for a workflow, that can be specified '
+    'with `workflow/last_stages` in config',
```
cassimons, line 55: Is this correct that we can now only select from these as 'last_stages'? (I was not clear where the logic ended up in that previous PR).

vladsavelyev: Yeah. The previous PR would have allowed to select any stages for `last_stages` (including `Align`, etc), but it didn't go in

vivbak (approved):

PR #172, opened 2022-11-03
Add seqr pool label seqr config

cassimons:


lgruen (approved):
Thanks!

vladsavelyev:
@lgruen @cassimons hold on, it should be already set in `https://github.com/populationgenomics/production-pipelines/blob/79afc22c36c5159f89e7ac1b472ad81946f853d9/configs/defaults/seqr_loader.toml#L41` loaded by default for seqr loader

lgruen:
@vladsavelyev Hmm, https://batch.hail.populationgenomics.org.au/batches/378996 didn't seem to use that label though...

vladsavelyev:
@lgruen Must be something else, because it's in the merged config in that job
Screen Shot 2022-11-03 at 11 06 21
Looking

vladsavelyev:
@lgruen I suspect your change in Hail that adds label https://github.com/populationgenomics/hail/pull/197/files was broken when merging with the upstream?

vladsavelyev:
@lgruen I think it happened because the pool labels change wasn't merged into the upstream:
https://github.com/hail-is/hail/pull/11879
https://github.com/hail-is/hail/pull/11944

And the `cpg-workflows` image forces re-install of the `hail` pip package because of the pin ([`hail>=0.2.104`](https://github.com/populationgenomics/production-pipelines/blob/main/setup.py#L17)), which pulls the official package which is built from the upstream codebase instead of our fork.

lgruen:
Awesome find. I've commented on https://github.com/hail-is/hail/pull/11944, as it would be really nice to get this merged upstream. But your fix to remove the pin will hopefully work in the meantime.

PR #174, opened 2022-11-03
Remove hail package pin

vladsavelyev:
So we keep the fork-based package from the driver image instead of the upstream-based official package. Context: https://github.com/populationgenomics/production-pipelines/pull/172#issuecomment-1301575780

### DIFF
--- a/.github/workflows/test_batch.yaml
+++ b/.github/workflows/test_batch.yaml
@@ -22,7 +22,7 @@ jobs:
           python-version: '3.10'
 
       - name: Install
-        run: pip install .[test]
+        run: pip install .[test,full]
 
       - name: Run unit tests
         run: python -m pytest test/test_batch.py
--- a/.github/workflows/test_batch.yaml
+++ b/.github/workflows/test_batch.yaml
@@ -22,7 +22,7 @@ jobs:
           python-version: '3.10'
 
       - name: Install
-        run: pip install .[test,full]
+        run: pip install .[test]
 
       - name: Run unit tests
         run: python -m pytest test/test_batch.py
--- a/.github/workflows/test_cohort.yaml
+++ b/.github/workflows/test_cohort.yaml
@@ -22,7 +22,7 @@ jobs:
           python-version: '3.10'
 
       - name: Install
-        run: pip install .[test]
+        run: pip install .[test,full]
 
       - name: Run unit tests
         run: python -m pytest test/test_cohort.py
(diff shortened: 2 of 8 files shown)


lgruen (approved):
Great find!

lgruen (approved):
Do you need to bump the version, too?

vladsavelyev:
Doesn't work, I think I need to remove `hail` from reqs altogether.

vladsavelyev:
@lgruen can you check again?

vladsavelyev:
@leo still doesn't work: we have `analysis-runner` in deps, and `analysis-runner` [depends on `hail`](https://github.com/populationgenomics/analysis-runner/blob/main/setup.py#L35), which forces re-install. I'm a bit confused because I though pip would re-install only if `--ignore-installed` is set. 

Any suggestions?

lgruen:
> Doesn't work, I think I need to remove `hail` from reqs altogether.

That's so weird. Do you understand why? If I run `pip3 install hail` in the driver image, it doesn't overwrite anything.

lgruen:
Let me try building this directly in the driver image...

vladsavelyev:
Okay, after moving both `hail` and `analysis-runner` into extras, it works!
(doesn't happen with `gnomad` package because it [excludes](https://github.com/broadinstitute/gnomad_methods/blob/main/setup.py#L12) `hail` from `install_requires`)

lgruen:
That's good to know, but I'm still trying to understand what's happening. `pip` is actually collecting `hail` 0.2.103, 0.2.104, and 0.2.105, even when removing the pin. I'm trying to get a better output from `pip` that would explain the dependency chains here, as something seems really odd. Must be `analysis-runner`, I guess, if moving it to `extras` works?

lgruen:
But also `analysis-runner` should already be included in the driver image...

vladsavelyev:
> pip is actually collecting hail 0.2.103, 0.2.104, and 0.2.105, even when removing the pin.

I actually saw this behaviour when updating `analysis-runner` locally: pip started pulling tons of different versions of `analysis-runner` for no reason. I wonder if there was some update in pip that caused this weird behaviour.

I'm off for a couple of hours, let me know how it goes and happy to pick up from you and continue digging into this issue.

vladsavelyev:
After discussion in Slack, the solution was to remove requirements from `analysis-runner` that it doesn't use directly (we had `hail`'s dependencies duplicated for reason we couldn't figure out)

PR #179, opened 2022-11-05
Update verifybamid path

cassimons:
Path to verifybamid has changed in latest image.  Is there a better way to manage changes to container internal paths or do we just need to update them like this when they break?


lgruen:
> Path to verifybamid has changed in latest image. Is there a better way to manage changes to container internal paths or do we just need to update them like this when they break?

The Docker image paths are already part of the config, but maybe we should build the Docker image itself in a way that the executable path doesn't contain the version, or just add `VerifyBamID` to the `$PATH`. @vladsavelyev, thoughts?

vladsavelyev:
Yes, that's going to be an easy addition to the Docker file.

And thank you for this fix!

vladsavelyev (approved):

cassimons:
Great thanks :-)

PR #180, opened 2022-11-07
Joint genotyping: extract genomicsdb tar into $BATCH_TMPDIR

vladsavelyev:
GenomicsDB tar became large enough so it doesn't fit the default disk. Extracting instead into `$BATCH_TMPDIR` which is mounted to the attached volume.


cassimons (approved):
Thanks you, that is great. I thought that might be the fix but had not understood the filesystem structure yet.

PR #185, opened 2022-11-08
Remove sort after gatherVCFs

cassimons:
GatherVcfsCloud will produce a sorted vcf provided our intervals are non overlaping and listed in genomic order. This sort is therefore not needed and requried a huge abount of temp space with big vcfs so should be removed.

### DIFF
--- a/cpg_workflows/jobs/vcf.py
+++ b/cpg_workflows/jobs/vcf.py
@@ -113,13 +113,8 @@ def gather_vcfs(
     --ignore-safety-checks \\
     --gather-type BLOCK \\
     {input_cmdl} \\
-    --output $BATCH_TMPDIR/gathered.vcf.gz
+    --output {j.output_vcf['vcf.gz']}
 
-    bcftools sort --temp-dir $BATCH_TMPDIR \
-    $BATCH_TMPDIR/gathered.vcf.gz -Oz \
-    -o {j.output_vcf['vcf.gz']}
-    
-    tabix -p vcf {j.output_vcf['vcf.gz']}
     """
     j.command(command(cmd, monitor_space=True))
     if out_vcf_path:


vladsavelyev:
I had to add sort at some point: https://github.com/populationgenomics/production-pipelines/commit/1b16381152716e435de5695aead731bd5f645705

The reason for that was the following: to run VQSR, we have to split the gathered VCF again by interval. And even though the intervals were correctly ordered, if there is an indel at a boundary of two intervals, it would end up duplicated in two distinct partitions, and after merging a VCF again, it would cause a sequence of variants slightly out of order.

Besides adding sort, I ended up changing logic of VQSR slightly to avoid the duplication, namely calling it on indels separately from SNPs. SNPs can't span two intervals, and indels won't be partitioned at all, so the problem is mitigated. Sorting though is still needed because now we have to merge SNPs and indels, which are not in order: https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/jobs/vqsr.py#L278

Now that I wrote that, I think we can use a better workaround. We can delay full VCF merge, and pass partitioned site-only VCFs down directly to VQSR, to avoid secondary duplication. Still need to think what to do with the model creation step of VQSR that needs a full unpartitioned VCFs. I'll follow up with a PR, or push into this one.

cpg_workflows/jobs/vcf.py
```diff
@@ -113,13 +113,8 @@ def gather_vcfs(
...
     --gather-type BLOCK \\
     {input_cmdl} \\
-    --output $BATCH_TMPDIR/gathered.vcf.gz
+    --output {j.output_vcf['vcf.gz']}
 
-    bcftools sort --temp-dir $BATCH_TMPDIR \
-    $BATCH_TMPDIR/gathered.vcf.gz -Oz \
-    -o {j.output_vcf['vcf.gz']}
-    
-    tabix -p vcf {j.output_vcf['vcf.gz']}
```
vladsavelyev, line 122: We still need to index VCF.

vladsavelyev:
De facto merged inside a large refactoring PR, so closing :) Thanks Cas for the suggestion that led to a more efficient code!

PR #186, opened 2022-11-09
Refactor split/gather VCF to avoid need to sort

vladsavelyev:
2G times the number of samples is not sufficient for sort in merging joint-calling VCFs:
https://batch.hail.populationgenomics.org.au/batches/379527/jobs/15
Attempting with twice that storage: 
https://batch.hail.populationgenomics.org.au/batches/379658/jobs/1

UPD: changing the logic to avoid need to sort at all. See conversation at https://github.com/populationgenomics/production-pipelines/pull/185


illusional:
Is this because we need to decide storage up-front?
Could we use some simple heuristic like 4GB for VCFs we don't know, and look it up if they're already processed (ie: exist at the output location at time of batch creation)?

cassimons:
It is only running out space during the post merge sort. Is there a scenario where this sort is required if the part Vcfs are correctly ordered?

vladsavelyev:
> It is only running out space during the post merge sort. Is there a scenario where this sort is required if the part Vcfs are correctly ordered?

It would be ideal if sort was not needed, however I had to add it at some point. The intervals are correctly ordered, and the resulting gathered VCF is sorted most of the time. However, once we had an issue with an indel at a boundary of two intervals, that caused a sequence of variants slightly out of order after merging.

But good spot, I'll think if we can make some work around for that, to avoid extra very consuming sort step for an almost sorted callset.

vladsavelyev:
Let's try without `sort`. I think the only place it was relevant is for gathering VQSR results (where we have indels and snps separately), and in that case VCFs are site-only so shouldn't require so much disk.

cassimons (approved):
Sounds like a good idea to me.  If we wanted a sanity check, is there a cheap "check vcf is sorted" validation we could add after the gather?

vladsavelyev:
> "check vcf is sorted" validation we could add after the gather?

That's just `tabix`. It would fail if the input is not sorted :)

vladsavelyev:
The update seems to be working, I'll merge now!

PR #196, opened 2022-11-13
Double maxlife of ES dataproc

cassimons:
Large cohorts timeout after 24h. Trying 48h.


vladsavelyev (approved):
Thanks!

I think running multiple index creation jobs slow them down individually; in isolation, they wouldn't run over. Something to think about how we want to balance that and if we want to run them in bunches.

lgruen:
The whole ES upload (and what the bottleneck really is) has always been a bit of a mystery to me. Hopefully we won't have to deal with it that much longer 🤞 .

PR #199, opened 2022-11-15
GATK-SV: add GATKSVPipelineBatch

vladsavelyev:
Test: https://batch.hail.populationgenomics.org.au/batches/380232


vivbak:
Moved: https://github.com/populationgenomics/production-pipelines/pull/345

PR #201, opened 2022-11-17
Mock up abstract state provider

vladsavelyev:


lgruen:
@illusional, how do you feel about this vs enabling a lightweight Metamist deployment with an easy way to set up the Metamist DB through e.g. a JSON file? That would keep the dependency from the pipeline side uniform, but would instead require more integration on Metamist.

PR #202, opened 2022-11-22
Add stripy

cassimons:
Adds a non-default stage `Stripy` to the`seqr_loader` pipeline that runs the tool stripy on each cram.

[Stripy](https://gitlab.com/andreassh/stripy-pipeline/-/tree/main) is STR calling and visualisation tool used to identify STR expansions at known disease-associated STR loci (ie this tool is not useful for STR discovery). Stripy:
- Uses expansion hunter to perform the STR genotyping at each target loci 
- Compares the repeat number to a provided database of population controls and known pathogenic ranges per loci
- Produces an easy-to-read HTML report including a read-level viewer per loci. 

The current implementation does have two issues we will need to look at in the future if we keep using it:
1. It is very inefficient at reading crams as it does not pass a reference genome path to the cram parsers. This makes a job that takes ~3min to run on a bam take close to 3h on an equivalent cram. This could be easily addressed in stripy and (presumably) be upstreamed.
2. This implementation currently fails on a subset of our crams (2/10 tested). Unfortunately, even with the most verbose level of logging no meaningful error or stack trace is provided other than a message saying "run failed". Again, adding some logging to stripy should not be too hard.

Despite these current limitations, I am keen to use this initial implementation 'as is' for wider evaluation of the results in production datasets. Eventually, this is likely to be superseded by the STR calling in GATK-SV, but that will also require interfaces for reviewing the variant calls so this will not be fast even once the GATK-SV is running. 

Running this stage costs 15c per genome using the default setting or 30c per genome using the `write_to_bam` option which is ~3x faster wall time (handy for testing).

### DIFF
--- a/cpg_workflows/jobs/stripy.py
+++ b/cpg_workflows/jobs/stripy.py
@@ -46,26 +46,28 @@ def stripy(
     j = b.new_job('STRipy', job_attrs)
     j.image(image_path('stripy'))
     
+    reference = fasta_res_group(b)
+
+    assert cram_path.index_path
+
     if write_to_bam:
         res = STANDARD.request_resources(storage_gb=100)
-        write_bam_cmd = '''
+        res.set_to_job(j)
+        write_bam_cmd = f"""
             BAM=$BATCH_TMPDIR/{cram_path.path.stem}.bam
             samtools view -b -@ 6 -T {reference.base}  $CRAM > $BAM
             samtools index $BAM
             ALIGNMENT=$BAM
-        '''
+        """
     else:
         res = STANDARD.request_resources(ncpu=2)
-        write_bam_cmd = '''ALIGNMENT=$CRAM'''
+        res.set_to_job(j)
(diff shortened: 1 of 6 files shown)


vladsavelyev (approved):
That's really awesome, love to see new stages plugged in :)

cpg_workflows/jobs/stripy.py
```diff
@@ -0,0 +1,114 @@
...
+        write_bam_cmd = """ALIGNMENT=$CRAM"""
+    
+    if sample.pedigree.sex:
+        sex_argument = f'--sex {str(sample.pedigree.sex).lower()}'
+    else:
+        sex_argument = ''
+
+    cmd = f"""\
+    ## Dodgy re-write of stripy config file
+    sed 's/"log_flag_threshold": 1/"log_flag_threshold": -1/' /usr/local/bin/stripy-pipeline/config.json \
```
vladsavelyev, line 78: Maybe some comment why it's needed

PR #204, opened 2022-11-23
Fix dataproc helper for large-cohort combiner

vladsavelyev:
* ~`hailctl`'s`--pyfiles` flag didn't work properly: it only loads `.py` files so `defaults.toml` gets missing. Hardcoding the defaults config instead, most of it will be moved into images and references repos anyway.~ Alternative solution: https://github.com/populationgenomics/hail/pull/249
* We still don't guarantee that metamist will have records for all GVCFs, so falling back to checking the file system for combiner.
* Fix getting package paths for dataproc helper's `pyfiles`.


cpg_workflows/__init__.py
```diff
@@ -7,6 +7,217 @@
...
+[validation.sample_map]
+HG001_NA12878 = 'HG001'
+SYNDIP = 'syndip'
+
+[hail]
+delete_scratch_on_exit = false
+"""
+
+
+f = tempfile.NamedTemporaryFile(delete=False, mode='w', suffix='.toml')
```
vivbak, line 221: At this stage could we read the file specified at cpg_workflows/defaults.toml and then write it to this temporary file as a work-around without needing to hardcode the .toml in this init file?

cpg_workflows/defaults.toml
```diff
@@ -1,207 +0,0 @@
-[workflow]
```
vivbak, line 1: This file is currently used as a reference by the analysis teams to help them understand the different params that can be overridden in a config. If we remove it in this PR do you think we should replace it with some other form of documentation?

vivbak (approved):
Thanks for the fix Vlad! Left 2 comments that might be worth considering :)

vivbak:
Ah I didnt see before -- also note the tests are failing

cpg_workflows/defaults.toml
vladsavelyev: The entire contents of that file is in `__init__.py`. Doesn't that work?

cpg_workflows/__init__.py
vladsavelyev: That's the problem I'm trying to workaround with this PR - we can't read it because it wouldn't be uploaded to dataproc along with python files. 

This would probably solve this though: https://github.com/populationgenomics/hail/pull/249

PR #206, opened 2022-11-23
Add script to fix alt contig MD5 sums in CRAM header

vladsavelyev:
Makes cpg_workflows' CRAMs compatible with WARP CRAMs.

Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1669180382708679

TODO:
- [x] Add a script that applies `samtools reheader` to each CRAM in a dataset (success: https://batch.hail.populationgenomics.org.au/batches/381398/jobs/1)
- [x] Pass unmasked reference to the `samtools merge` in the Align workflow's stage
- [ ] Reheader all post-NAGIM datasets


lgruen:
>  Run that script on all NAGIM datasets

Did you mean all **non**-NAGIM datasets, as the NAGIM CRAMs already point to the unmasked reference?

> Add samtools reheader into the Align workflow stage.

Should we instead change the `samtools merge` call (https://cs.github.com/populationgenomics/production-pipelines/blob/a823b6e1aea91818fbb86d0c3642525c9ca435a4/cpg_workflows/jobs/align.py) to pass the unmasked reference (`--reference`), similar to the WARP pipeline (https://github.com/broadinstitute/warp/blob/a3b438422691f447032038be52c74639dafe914f/tasks/broad/DragmapAlignment.wdl#L70)?

vladsavelyev:
@lgruen, you are correct, I meant post-NAGIM.

Re `samtools merge` - I haven't tested if that actually would work, was going to :)

lgruen (approved):
This is great!

vivbak:
Relevant Discussion: https://github.com/populationgenomics/production-pipelines/issues/208

PR #207, opened 2022-11-24
Stripy: remove write_to_bam option, use cloud fuse and cleanup. 

cassimons:
Following upstreaming a couple of bugfixes to stripy-pipeline this PR removes the workarounds that were needed to read from crams. Also includes some basic cleanup edits.

* Remove code to convert cram to bam before running stripy. Reading directly from crams now works
* Use cloud fuse to access cram. Stripy only reads from a small number of small loci so fuse is faster than localization.
* Move target loci list to config 
* Export json version of results
*  Remove unused imports


MattWellie (approved):
Is it using the specific loci that forces it to process CRAMs correctly, or is that an unrelated version update?

cassimons:
No, I just moved the loci to the config so we can change them per project if needed (eg some projects will not want to include the HD STR). The cram processing fix needed the genome reference to be passed to the pysam cram reader inside stripy. I rebuilt the stripy image to pull the update.

PR #211, opened 2022-11-28
Adjust for new generated reference and images config templates

vladsavelyev:
Depends on https://github.com/populationgenomics/cpg-utils/pull/117


configs/defaults/seqr_loader.toml
```diff
@@ -63,4 +63,7 @@ target_loci = """AFF2,AR,ARX_1,ARX_2,ATN1,ATXN1,ATXN10,ATXN2,ATXN3,ATXN7,ATXN8OS
 BEAN1,C9ORF72,CACNA1A,CBL,CNBP,COMP,DAB1,DIP2B,DMD,DMPK,FMR1,FOXL2,FXN,GIPC1,GLS,\
 HOXA13_1,HOXA13_2,HOXA13_3,HOXD13,HTT,JPH3,LRP12,MARCHF6,NIPA1,NOP56,NOTCH2NLC,\
 NUTM2B-AS1,PABPN1,PHOX2B,PPP2R2B,PRDM12,RAPGEF2,RFC1,RUNX2,SAMD12,SOX3,STARD7,TBP,\
-TBX1,TCF4,TNRC6A,XYLT1,YEATS2,ZIC2,ZIC3"""
\ No newline at end of file
+TBX1,TCF4,TNRC6A,XYLT1,YEATS2,ZIC2,ZIC3"""
+
+[images]
+stripy = 'australia-southeast1-docker.pkg.dev/cpg-common/images/stripy:v2.2'
```
lgruen, line 69: Is the idea of having an `images` section in some of these default configs to override the newest versions that would come from the image config template?

cpg_workflows/defaults.toml
```diff
@@ -1,14 +1,8 @@
...
-
-# Prefix to find reference files referenced in `refdata_yaml_path`
-reference_prefix = 'gs://cpg-reference'
-
-# Template to build HTTP URLs matching the dataset_path of category
-# "web". Should be parametrised by namespace and dataset in Jinja format:
-web_url_template = 'https://{namespace}-web.populationgenomics.org.au/{dataset}'
+[images]
+dragmap = 'australia-southeast1-docker.pkg.dev/cpg-common/images/dragmap:1.3.0'
+cpg_workflows = 'australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:latest'
```
lgruen, line 3: In terms of reproducibility, is using the `latest` tag a bit dangerous here? Wouldn't the image config template contain an explicit tag for the newest version?

lgruen (approved):
Looks good, but I'm not yet understanding fully why we still need the `images` sections here.

configs/defaults/seqr_loader.toml
vladsavelyev: That's because we don't manage `stripy` through images because wasn't clearly versioned: https://github.com/populationgenomics/images/pull/50#issuecomment-1310104799
(though I checked again in the gitlab and there is v2.2 tag now, so can add it).

cpg_workflows/defaults.toml
vladsavelyev: We don't maintain `cpg_workflows` images through the images repo (it's build through the `production_pipelines` CI). Could move it there of course

configs/defaults/seqr_loader.toml
vladsavelyev: Adding stripy: https://github.com/populationgenomics/images/pull/62

cpg_workflows/defaults.toml
vladsavelyev: It would be too complicated though. I think I will just manage the tag here with bumpverison :)

lgruen: That's such an elegant solution! Thanks!

PR #214, opened 2022-11-29
Changes from Miah (fix use Batch backend for seqr annotation)

MattWellie:
A couple more fixes from @jeremiahwander

- the docker image doesn't contain the hail-scripts submodule (@vladsavelyev this will probably also be an issue with the way cpg-workflows is being installed into the driver image? https://github.com/populationgenomics/analysis-runner/pull/561
  - e.g. https://batch.hail.populationgenomics.org.au/batches/381164/jobs/1 
- possibility of `siteonly_vqsr_vcf_path` being None and cast as a String, leading to `to_path('None')` downstream
- extending the array casting to apply to both AC and AF

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.1.10
+current_version = 1.1.11
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


vladsavelyev (approved):
Awesome, thank you! I will make it work with the Batch backend :)

PR #216, opened 2022-12-01
Make GQ bands for ReblockGVCF configurable

lgruen:
No change to the actual values (i.e. a functional no-op), but could be used to add more bands for RD samples going forward (@cassimons FYI).


cassimons:
Great thanks!  I assume if we wanted to increase the block number we should probably re-call all existing samples?

vladsavelyev (approved):

PR #219, opened 2022-12-06
Fixes for large-cohort sample_qc module

vladsavelyev:
* Enable status reporter by default to make sure metamist GVCF entries are populated;
* Fall back to checking objects on buckets for GVCFs


vivbak (approved):
🙏 Thanks!

PR #226, opened 2022-12-12
Test for stages configuration options

vladsavelyev:
Add some unit tests for `skip_stages`, `first_stages`, `last_stages`, `force_stages` options. New tests for different edge scenarios can be easily added on top of the existing framework.

(Also addresses questions in https://github.com/populationgenomics/production-pipelines/pull/158)


lgruen (approved):
Thanks, this is helpful! Can we specify these semantics (which led to the questions in #158) as clearly as possible in `defaults.toml`?

PR #227, opened 2022-12-12
Make `add_make_sitesonly_job` method public, add option to override storage

vladsavelyev:


MattWellie (approved):
Nice - for AIP's purposes this will need a new step in the flow?

1. Take full VCF and make sites-only VCF
3. Split sites-only VCF into sub VCFs
4. Annotate sub-VCFs
5. Combine annotations
6. Annotate against the original VCF -> MT

vladsavelyev:
Matt, exactly!

PR #228, opened 2022-12-13
Skip SomalierPedigree for Prophecy

vivbak:
FYI @KatalinaBobowik


vladsavelyev:
Since `SomalierPedigree` is a [required stage](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/cram_qc.py#L239) for `CramMultiQC`, so I'm not sure if skipping it is going to work. 

If we don't have pedigree info in metamist, I think it should be okay to keep the stage running; it should just do nothing with empty PED data (need to double check that)

vivbak:
hmm, do you have any idea what would be causing this failure then https://batch.hail.populationgenomics.org.au/batches/393766/jobs/16

configs/prophecy.toml
```diff
@@ -9,6 +9,8 @@ skip_samples = [
   'CPG216481',  # missing GVCF
   'CPG241299',  # Replaced by CPG262907
 ]
+skip_stages =['SomalierPedigree']
```
vladsavelyev, line 12:
```suggestion
skip_stages = ['SomalierPedigree']
```

vladsavelyev (approved):
Seems to be working!

PR #231, opened 2022-12-13
switch to highmem workers 

vivbak:
Last run failed here https://batch.hail.populationgenomics.org.au/batches/396745/jobs/2
FYI @KatalinaBobowik

### DIFF
--- a/configs/prophecy.toml
+++ b/configs/prophecy.toml
@@ -11,6 +11,7 @@ skip_samples = [
 ]
 skip_stages = ['SomalierPedigree']
 skip_samples_with_missing_input = true
+highmem_workers = true
 
 scatter_count = 20
 output_version = '0.1'


configs/prophecy.toml
```diff
@@ -11,6 +11,7 @@ skip_samples = [
 ]
 skip_stages = ['SomalierPedigree']
 skip_samples_with_missing_input = true
+highmem_workers = true
```
lgruen, line 14: Is there a way to restrict that to only the failing stage (SampleQC), instead of applying this to all stages?
Should we maybe change this generally for the SampleQC stage in the code instead of the config?

lgruen (approved):

configs/prophecy.toml
vivbak: Good point! I think I can see what change needs to be made for this -- will address in a later PR :)

PR #233, opened 2022-12-14
Don't depend on a specific VEP version for LOFTEE

lgruen:
Depends on images being built after https://github.com/populationgenomics/images/pull/68.

Using a similar wildcard as here: https://github.com/populationgenomics/references/blob/9aeb22804da19193b679087269459d9eeaafb2ea/vep/dataproc-init.sh#L46


MattWellie (approved):
Thanks 😆 I was just about to make a branch for this 🎉

PR #236, opened 2022-12-16
Large-cohort sex imputation: use variant data

vladsavelyev:
Fixes the issue with the dataproc cluster running out of memory (even on `highmem` workers) during the sex imputation part of the `SampleQC` stage. Slack thread for context and links: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1670911574944789

The problem was introduced when the [`annotate_sex` function](https://github.com/broadinstitute/gnomad_methods/blob/7fb3582f359ed7e1f96e370cf60c0f6f8e8c3fb2/gnomad/sample_qc/pipeline.py#L527-L560) from the `gnomad` package was updated for VDS input to estimate coverage from reference blocks and intervals, instead of the variant data (which is just the `mt.DP` field). The gnomAD team probably didn't hit this issue on the real data because they are using a pre-calculate coverage for ploidy estimation (`coverage_mt` is defined), so never going this code branch.

In the meantime, addressing the problem by reverting back to using the variant data for coverage information (like it was used in pre-VDS workflow versions). Longer term, we should pre-calculate coverage using the gnomAD model (draft PR for that: https://github.com/populationgenomics/production-pipelines/pull/159)


vivbak (approved):
🙏 🙌 Thanks so much Vlad!

PR #241, opened 2022-12-16
Fix gathering site-only VQSR VCFs

vladsavelyev:
Fixes https://batch.hail.populationgenomics.org.au/batches/396779\?last_job_id\=1180

To merge per-interval VCFs, we currently use GATK GatherVcfsCloud with `--gather-type BLOCK` which is very fast to merge VCFs without uncompressing, but it behaves weirdly around interval boundaries. For example, given the intervals:

```
$ tail -n2 128.interval_list
...
chr12	132240356	132240878	+	ice_target_140111|protein_coding
chr12	132247290	132247735	-	retained_intron|protein_coding|ice_target_140112

$ head -n2 129.interval_list
chr12	132247736	132248177	-	retained_intron|protein_coding|ice_target_140112
chr12	132257538	132258036	+	ice_target_140113|protein_coding|processed_transcript
...
```

The resulting VCFs are:

```
128.vcf.gz
...
chr12  132247673 ... # long indel that with end at 132247746, which is beyond the interval boundary at 132247735
...
chr12  132247713 ... # SNP
chr12  132247717 ... # SNP
chr12  132247722 ... # SNP

129.vcf.gz
chr12  132247739 ... # SNP
chr12  132247745 ... # SNP
...
```

The resulting merged VCF ends up having that long indel positioned after the SNPs that start downstream of it:

```
...
chr12  132247713  # SNP from 128.vcf.gz
chr12  132247717  # SNP from 128.vcf.gz
chr12  132247722  # SNP from 128.vcf.gz
chr12  132247673  # the long indel from 128.vcf.gz that ends at 132247746
chr12  132247739  # SNP from 129.vcf.gz
...
```

To address this, I'm switching away from the native block approach, and instead indexing VCF at each partition, and passing those to `bcftools merge` which makes use of VCF indices and produces a correct sorted output. 

I also added a lot of checkpointing between the VQSR jobs that would help quicker debug problems in the future.

Test: https://batch.hail.populationgenomics.org.au/batches/419841?last_job_id=200

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.6.1
+current_version = 1.7.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  VERSION: 1.6.1
+  VERSION: 1.7.0
 
 jobs:
   docker:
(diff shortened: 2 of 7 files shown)


cassimons (approved):
Thanks Vlad!  I am still a bit confused how we end up with an out-of-order boundary in this instance, but I agree this seems like a good general fix for this issue.

PR #246, opened 2022-12-20
adds an analysis prefix generator

MattWellie:
Proposal - adds a Dataset method for generating `-analysis` bucket paths


cassimons:
LGTM but I will leave to @vladsavelyev

vladsavelyev (approved):

PR #250, opened 2022-12-21
`overwrite` logic in `can_reuse`

vladsavelyev:
Change type of `overwrite` in `can_reuse` to simple `bool`, and make explicit `True` to have precedence over the `workflow.check_intermediates` config option (so a stage can pass `overwrite=True` for forced samples and stages). 
 
Example: 

```toml
[workflow]
forced_stages = ['Align']
```

Results in the `Align` stage defined in `stages/align.py` passing `overwrite=sample.forced` as `True` to the `align` function in `jobs/align.py`, making that job ignore any existing CRAMs or intermediate results.

However, if otherwise `check_intermediates` is set to `False`:

```toml
[workflow]
check_intermediates = false
```

`overwrite=sample.forced` would have `False`, however `can_reuse` would check `check_intermediates` and still indicate that the results can be overwritten.


cassimons (approved):
LGTM thank you!

Would the intent be clearer if we changed the argument`overwrite` to something `force_to_rerun`?

When I read "overwrite" in the function, I was not clear to me if this argument was referring to forcing a job to run, or more about protecting some outputs from being overwritten. I think this would particularly add clarity when this argument is passed in job declarations.

PR #252, opened 2022-12-22
Use CPG IDs for stripy report path.

cassimons:
- Changes stripy output files to use CPG Ids
- Adds External ID into stripy HTML report

### DIFF
--- a/cpg_workflows/jobs/stripy.py
+++ b/cpg_workflows/jobs/stripy.py
@@ -78,6 +78,9 @@ def stripy(
         --locus {target_loci}
 
     ls $BATCH_TMPDIR/
+
+    # Add external ID into html report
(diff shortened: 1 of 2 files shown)


cpg_workflows/jobs/stripy.py
```diff
@@ -23,7 +23,7 @@ def stripy(
     json_path: Path,
     analysis_type: str = 'standard',
     job_attrs: dict | None = None,
-    overwrite: bool = False,
+    overwrite: bool = False
```
vladsavelyev, line 26: Is removal of comma intentional?

cassimons: No. just an artefact of merges that I did not bother to revert.  I can add it back in.

MattWellie (approved):

vladsavelyev (approved):

PR #256, opened 2023-01-03
Mechanics to pass a custom function to update `Analysis.meta`

vladsavelyev:
Supporting passing a custom python function to a stage that is evaluated after the stage is successfully finished. The function takes the output path as input, and returns a dictionary to merge into the `Analysis` entry's metadata.

Example usage is in the `test.test_status.py`:

```python
def _custom_update_meta(output_path: str) -> dict[str, Any]:
    from cloudpathlib import CloudPath

    with CloudPath(output_path).open() as f:
        return {'result': f.read().strip()}

@stage(analysis_type='custom', update_analysis_meta=_custom_update_meta)
class MyStage(SampleStage):
    def expected_outputs(self, sample: Sample) -> Path:
        return to_path(dataset_path('my_output.txt'))

    def queue_jobs(self, sample: Sample, inputs: StageInput) -> StageOutput | None:
        j = get_batch().new_job('My job', self.get_job_attrs(sample))
        j.command(f'echo 42 >> {j.output}')
        get_batch().write_output(j.output, str(self.expected_outputs(sample)))
        return self.make_outputs(sample, self.expected_outputs(sample), [j])

run_workflow(stages=[MyStage])
```

Successful job will produce an `Analysis` entry with the `result` field in `meta`:

```yaml
id = 1
output = 'gs://.../my_output.txt'
meta:
    result = 42
``` 

Context and inspiration for this feature: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1671765903071339?thread_ts=1671672065.519439&cid=C030X7WGFCL

### DIFF
--- a/cpg_workflows/stages/align.py
+++ b/cpg_workflows/stages/align.py
@@ -2,6 +2,7 @@
 Stage that generates a CRAM file.
 """
 import logging
+from cloudpathlib import CloudPath
 
 from cpg_utils import Path
 from cpg_utils.config import get_config
--- a/cpg_workflows/status.py
+++ b/cpg_workflows/status.py
@@ -1,17 +1,23 @@
 """
 Metamist wrapper to report analysis progress.
 """
-
-from textwrap import dedent
+import inspect
 from abc import ABC, abstractmethod
+from typing import Callable, Any
 
 from cpg_utils import Path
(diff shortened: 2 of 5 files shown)


cassimons (approved):
LGTM thanks!

PR #258, opened 2023-01-06
Allow passing remote path to `main.py --config <path>`

vladsavelyev:
Just handy for debugging.

Normally you would pass your config file via the analysis runner CLI, e.g.:

```sh
analysis-runner --dataset seqr --access-level test --description "Test" --output-dir output \
--image australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:latest \
--config ~/tmp/myconfig.toml \
main.py seqr_loader
```

However, you might want to quicker iterate on debugging issues when running the main workflow code, and run `main.py seqr_loader` directly. In this case, it's handy to dig out the final merged config remote path as it's generated by the analysis runner by checking out job's `CPG_CONFIG_PATH` env variable, e.g.:

Screen Shot 2023-01-06 at 14 31 53

And then pass it directly to `main.py` as follows:

```sh
python main.py seqr_loader --config gs://cpg-seqr-test-tmp/batch-tmp/config/4c1c812c-1907-43b5-9491-eb41b802d723.toml
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -38,7 +38,6 @@ WORKFLOWS: dict[str, list[StageDecorator]] = {
     '--config',
     'config_paths',
     multiple=True,
-    type=click.Path(exists=True),
     help='Add configuration files to the files specified $CPG_CONFIG_PATH.'
     'Configs are merged left to right, meaning the rightmost file has the'
     'highest priority.',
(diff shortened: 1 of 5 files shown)


main.py
```diff
@@ -38,7 +38,6 @@
     '--config',
     'config_paths',
     multiple=True,
-    type=click.Path(exists=True),
```
MattWellie, line 41: Is this being removed because the click.exists doesn't support cloud paths?

MattWellie (approved):

main.py
vladsavelyev: Yep, exactly!

PR #260, opened 2023-01-09
Script to remove arbitrary workflow output using suffix

vladsavelyev:
Usage:

```python
python scripts/remove_file.py Vqsr/tmp/gathered.vcf.gz
```

The prefix is determined from the config, assuming same config as used for the original workflow.

E.g. for the recent exome seqr load, it will remove `gs://cpg-seqr-main/exome/seqr_loader/14433af228078f9910041fff009835c8c43709_862/Vqsr/tmp/gathered.vcf.gz`

I'm not sure though if it's too much flexible and unsafe, so leaving for Leo to review

cc @cassimons

### DIFF
--- a/scripts/remove_file.py
+++ b/scripts/remove_file.py
@@ -13,6 +13,7 @@ def main(suffix: str):
     Remove arbitrary file from a workflow run output prefix.
     """
 
+    print('Loading workflow...')
     wfl = get_workflow()
     path = wfl.prefix / suffix
     print(f'Using workflow prefix {wfl.prefix}, full path would be {path}')
@@ -21,3 +22,7 @@ def main(suffix: str):
     else:
         path.unlink()
         print(f'Removed object {path}')
+
+
+if __name__ == '__main__':
+    main()


lgruen (commented):
I think this is a question of how much risk we want to take for data manager tasks like these. From my perspective, the deletion of files is inherently dangerous (despite our short-time undelete functionality on most buckets). I'd argue for the filename pattern to be the crucial part that should get reviewed carefully before executing a destructive operation like deletion. Since this would be moved to a parameter for this script, it circumvents the safeguards that a review provides.

cassimons:
Thanks @lgruen, I can see that POV, although I personally I am not convinced that the level of systematic controls needs to be as rigid to cover this risk. If we are not going to move forward with this approach this conversation probably needs to move to another venue, but a few points:
- As currently implemented, this is not a _data manager_ task, but rather a functionality needed by anyone in my team running the RD pipeline. Data is stale and needs to be regenerated, currently the only way to do this is via removing the existing result file.
- In terms of functionality we specifically need in this instance, this could instead be achieved by a workflow config forcing a specific stage to re-run. This is also fine with me, either way, it just needs to be something we can achieve with low friction.
- If we do want specific scripts with hard coded paths for removal to be added by PR, it would be great to have some specific advice on where these should live (I know it could be anywhere, but too much choice is hard).
- If we stick with PRs with hardcoded paths then I think it is important this is used by all teams at all times.

lgruen:
> * As currently implemented, this is not a _data manager_ task, but rather a functionality needed by anyone in my team running the RD pipeline. Data is stale and needs to be regenerated, currently the only way to do this is via removing the existing result file.

Right, but I think it's important to distinguish between a general approach here and the quick fix that we need to unblock this particular seqr load. A general deletion script is fairly dangerous IMO and as you point out below isn't the ideal way to configure this (medium term) anyway. If we just need a quick fix, I'd suggest to just commit that one-liner script as a separate PR and link the associated seqr issue, so it's all tracked. 

> * In terms of functionality we specifically need in this instance, this could instead be achieved by a workflow config forcing a specific stage to re-run. This is also fine with me, either way, it just needs to be something we can achieve with low friction.

I thought we already had the functionality to force a rerun of specific stages? Curious why this doesn't work here and whether this instead needs a bug fix in the pipeline itself?

> * If we do want specific scripts with hard coded paths for removal to be added by PR, it would be great to have some specific advice on where these should live (I know it could be anywhere, but too much choice is hard).

Maybe `seqr-private`, as presumably that's where the corresponding issues would be filed and those scripts could also potentially contain sample IDs etc. that we might want to keep private?

vladsavelyev:
> I thought we already had the functionality to force a rerun of specific stages? Curious why this doesn't work here and whether this instead needs a bug fix in the pipeline itself?

Sorry for chiming in. There is already a functionality to rerun specific stages (`force_stages = []`). However, in this case we didn't want to rerun the entire stage (i.e. including VQSR jobs for each intervals, etc), but rather just the VCF-gather job. This is a one-off scenario: before the holidays, we [fixed](https://github.com/populationgenomics/production-pipelines/pull/241) an issue happening on the interval edges when gathering VCFs, and we want the workflow to only rerun the gathering.

However, in my experience, similar scenarios can happen from time to time in unexpected places when running workflows on the real data, and an ability to remove specific files when debugging issues with workflows can be quite handy.

PR #261, opened 2023-01-09
fix annotate_cohort.py invocation typo

cassimons:
The invocation of annotate_cohort.py is mis formated leading to this error: https://batch.hail.populationgenomics.org.au/batches/420104/jobs/206


vladsavelyev (approved):
Thank you for spotting and fixing!

PR #262, opened 2023-01-09
Enable QC reporting to Slack. Minor touches to Slack messages

vladsavelyev:
Config's `[slack]` section - if added - activates QC reports to a Slack channel. For example:

```toml
[slack]
channel = 'workflows-qc'
token_secret_id = 'slack-seqr-loader-token'
token_project_id = 'seqr-308602'
```

Would send messages on QC completion to `workflows-qc`, flagging failed samples.

Here, re-enabling this functionality. I also renamed the existing `seqr-loader-test` channel to `workflows-qc`, as it should be relevant for any workflows. However, I'm only keeping it for the seqr loader now, as the secret `slack-seqr-loader-token` is maintained manually under the `seqr-308602` project, and other datasets can't read it, so we would need to move it under Pulumi to add read access for other datasets and enable it for large-cohort as well.

Two scripts send Slack messages: `check_multiqc.py` and `check_pedigree.py` (pedigree/sex checks with Somalier). Note that the MultiQC checks only work when the `qc_thresholds` section is also specified, for example:

```toml
[qc_thresholds.genome.min]
"MEDIAN_COVERAGE" = 10
"PCT_PF_READS_ALIGNED" = 0.80
[qc_thresholds.genome.max]
"FREEMIX" = 0.04
"PERCENT_DUPLICATION" = 25
```

Thresholds can be adjusted as needed, per workflow or dataset. Note that the metrics names here should match the ones in the MultiQC JSON; e.g. gs://cpg-validation-main/qc/cram/f5705fb1bb4bb1a966c4e970029d9072460d5e_6/multiqc_data.json

Additionally:
* abstracted sending Slack messages into a function in the `slack` module;
* reading the token from a gcloud secret directly inside the job, instead of passing it as an environment variable, to avoid expositing it;
* fixing the % of unalignment reads thresholds;
* removed backticks to get rid of misleading error logging in Batch jobs;
* added documentation for the Slack section;
* more emojis ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,3 +698,30 @@ CramQC = ['CPG13409']
 ### Available stages
 
 To see available workflow stages, explore the `cpg_workflows/stages` folder. A stage wrappers for Hail Batch jobs that allow to glue them together into genomics workflows. `cpg_workflows` also provides helper functions that create Hail Batch jobs for different domain-specific applications, e.g. alignment, deduplication, QC, variant calling, creating calling intervals, running VQSR, creating ES index, etc. Those jobs can be called from the `queue_jobs` method of a stage, or in isolation, as they don't need to know about the stage they are called from, but only need a `Batch` object.
+
+
+### Slack QC notifications
+
+Config's `[slack]` section activates QC reports to a Slack channel. For example:
+
+```toml
+[slack]
+channel = 'workflows-qc'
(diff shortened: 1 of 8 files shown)


cassimons (approved):
Looks great thanks. Will be good to have this active again.

PR #263, opened 2023-01-10
Do not "force" the QC stages

vladsavelyev:
We do not need to "force" the QC stages (MultiQC and SomalierPedigree), as their non-HTML outputs (like the MultiQC JSON and the Somliaer TSV files) are now parametrised by the samples hash (i.e. `h = dataset.alignment_inputs_hash()`). Having those outputs would make the stage rerun when the set of samples has changed.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.11.0
+current_version = 1.11.1
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  VERSION: 1.11.0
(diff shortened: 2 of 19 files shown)


cassimons (commented):
Hey @vladsavelyev, this PR seems to also have all of the changes from the last PR as well. Is that expected?

vladsavelyev:
@cassimons it was stacked on the last PR. After you merged last PR, the base branch automatically switched to `main`. However, the actual commits from the last PR are not main as they were squashed into single merge commit. That's why GitHub sees a mismatch in commit history, and shows all changes from both PRs. I will need to merge the branch with main in order the diff to become clean.

vladsavelyev:
@cassimons and merging with main often as easy as clicking the "update branch" button here in the GitHub UI (unless there are conflicts), which I did now, so the history should be clean.

MattWellie (approved):

PR #270, opened 2023-01-13
Specify background datasets for PCA analysis

vladsavelyev:
Support specifying background datasets for PCA analysis and ancestry inference.

This draft supports the following section in the config:

```toml
[large_cohort.pca_background]
datasets = ['gs://cpg-hgdp-main/mt/oceania_eur.mt']
pop_field = 'continental_pop'
```

Where datasets must be a list paths to matrix tables or VDS, and `pop_field` if defined should be the column-level field in the datasets to extract the population tag for RF training for ancestry inference.

Testing should be done, and the following work to include:

- [ ] It could take dataset names as input so we don't hardcode any paths in the config;
- [ ] Currently it needs permissions to the background dataset (e.g. if you run with prophecy, it would need to read from hgdp and it would fail because hgdp is not a dependency for prophecy). So need to either copy the mt to reference bucket e.g. into `gs://cpg-common-main/references/ancestry`, or to set dependencies for stacks.
- [ ] Might make sense to explore if it makes sense run the vcf-combiner on input datasets together instead of running the union; though perhaps because we work on desne mts that are subset to QC sites and are not huge as result, the union is more efficient.
- [ ] Keep pre-subset and and dense background mts in the reference bucket, add pre-processing script that produces such mts.


configs/prophecy.toml
```diff
@@ -1,14 +1,22 @@
 [workflow]
 input_datasets = ['prophecy']
 sequencing_type = 'genome'
+access_level= 'test'
```
vivbak, line 4: I think I actually added this, but we don't need it if running via analysis-runner :) 
```suggestion
```

cpg_workflows/large_cohort/ancestry_pca.py
```diff
@@ -3,17 +3,64 @@
...
+            mt = hl.read_matrix_table(str(path))
+            mt = hl.split_multi(mt, filter_changed_loci=True)
+            mt = mt.semi_join_rows(qc_variants_ht)
+            mt = mt.densify()
+        elif to_path(path).suffix == '.vds':
+            vds = hl.vds.read_vds(str(path))
+            vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+            vds = hl.vds.filter_variants(vds, qc_variants_ht)
+            mt = hl.vds.to_dense_mt(vds)
+            # annotate mt with metadata info
```
vivbak, line 44: Could we expand this comment a little more, something like 
```suggestion
            # annotate background mt with metadata info derived from SampleQC stage 
```

cpg_workflows/large_cohort/ancestry_pca.py
```diff
@@ -3,17 +3,64 @@
...
+    qc_variants_ht = hl.read_table(
+        str(reference_path('gnomad/predetermined_qc_variants'))
+    )
+    dense_mt = (
+        dense_mt.select_cols().select_rows().select_entries('GT', 'GQ', 'DP', 'AD')
+    )
+    for path in pca_background['datasets']:
+        logging.info(f'Adding background dataset {path}')
+        if to_path(path).suffix == '.mt':
+            mt = hl.read_matrix_table(str(path))
```
vivbak, line 35: Would we be able to rename this to `background_mt` or something similar so it's clearer

vivbak (approved):
Wonderful 🙌  Just a couple of small things to improve readability.

cpg_workflows/large_cohort/ancestry_pca.py
KatalinaBobowik: agreed, I;ve changed this to `background_mt` (and `background_vds`) now

cpg_workflows/large_cohort/ancestry_pca.py
KatalinaBobowik: now updated :)

configs/prophecy.toml
KatalinaBobowik: oh neat, so this is redundant? i.e., the analysis runner will automatically know the access level and run the large cohort pipeline accordingly?

vivbak: Yes :)

vivbak (approved):
Looks great 🙌 !! Just a few linting things to fix :)

configs/prophecy.toml
KatalinaBobowik: perfect, removing then

Issue #282, opened 2023-02-05
Joint genotyping: MakeSitesOnlyVcf jobs failing in seqr load pipeline.


cassimons:
All jobs are failing with the following error message (eg [job](https://batch.hail.populationgenomics.org.au/batches/421086/jobs/5) [batch](https://batch.hail.populationgenomics.org.au/batches/421086?q=failed)). I have not investigated if the issue is here or in the upstream ExcessHet_filter job.

```+ gatk --java-options -Xms6500m MakeSitesOnlyVcf -I /io/batch/7c9719/ExcessHet_filter-U4Q7I/output_vcf -O /io/batch/7c9719/MakeSitesOnlyVcf-tOO8D/output_vcf.vcf.gz
Using GATK jar /root/micromamba/share/gatk4-4.2.6.1-1/gatk-package-4.2.6.1-local.jar
Running:
    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=2 -Xms6500m -jar /root/micromamba/share/gatk4-4.2.6.1-1/gatk-package-4.2.6.1-local.jar MakeSitesOnlyVcf -I /io/batch/7c9719/ExcessHet_filter-U4Q7I/output_vcf -O /io/batch/7c9719/MakeSitesOnlyVcf-tOO8D/output_vcf.vcf.gz
09:24:06.546 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/root/micromamba/share/gatk4-4.2.6.1-1/gatk-package-4.2.6.1-local.jar!/com/intel/gkl/native/libgkl_compression.so
[Sun Feb 05 09:24:06 UTC 2023] MakeSitesOnlyVcf --INPUT /io/batch/7c9719/ExcessHet_filter-U4Q7I/output_vcf --OUTPUT /io/batch/7c9719/MakeSitesOnlyVcf-tOO8D/output_vcf.vcf.gz --VERBOSITY INFO --QUIET false --VALIDATION_STRINGENCY STRICT --COMPRESSION_LEVEL 2 --MAX_RECORDS_IN_RAM 500000 --CREATE_INDEX true --CREATE_MD5_FILE false --GA4GH_CLIENT_SECRETS client_secrets.json --help false --version false --showHidden false --USE_JDK_DEFLATER false --USE_JDK_INFLATER false
[...]

PR #284, opened 2023-02-06
Fix MakeSitesOnlyVcf

vivbak:
Addresses https://github.com/populationgenomics/production-pipelines/issues/282 

I also did my best to check all other instances of ResourceGroups and it doesn't look like it happens again. I did find one other error in the tabix_vcf function.


illusional (approved):
Does this need a bumpversion and release?

PR #289, opened 2023-02-08
Fix path for tabix output file

cassimons:
The tabix job is currently trying to output a resource group rather than just the index file resulting in errors like this:
https://batch.hail.populationgenomics.org.au/batches/421210/jobs/3

This PR scopes the output to just the index file.


illusional (approved):
I have no issues with the change directly, except the resource group created likely wouldn't work, and this feels like it's going to disconnect the original file from it's resource.

Instead of j.output_tbi being a resource group, why not just make it a file?

PR #291, opened 2023-02-13
Dataset stage reporting correct dataset

illusional:
Broken foundation, mostly for commentary.


cassimons:
This is resulting in all analysis objects from the seqr pipeline since ~december being labelled with the incorrect dataset. It would be great to get this fix in if/when we can.

vivbak (approved):
Thanks @illusional! It was very not broken actually :D 
I just added an extra case to handle when the target is a Cohort.

Issue #292, opened 2023-02-17
Force re-use not working? 


vivbak:
For example: 
https://batch.hail.populationgenomics.org.au/batches/421746/jobs/1

The logs indicate the jobs should be queued 

```
2023-02-17 03:56:10 INFO (root 1056): Stage #7: CramQC [forced] <- [Align]
2023-02-17 03:56:10 INFO (root 1137): Dataset thousand-genomes (19 samples):
2023-02-17 03:56:10 INFO (root 601): CramQC: Sample(thousand-genomes/CPG261107|HG03943) [QUEUE] (can reuse, but forcing the stage to rerun)
2023-02-17 03:56:10 INFO (root 601): CramQC: Sample(thousand-genomes/CPG261115|HG03944) [QUEUE] (can reuse, but forcing the stage to rerun)
```

Yet
`2023-02-17 03:56:10 ERROR (root 153): No jobs to submit`

vivbak:
The approach here should be slightly re-thought in line with https://github.com/populationgenomics/production-pipelines/issues/294

daniaki:
@vivbak I can't remember, but I have a feeling we had a chat about this in person - was this fixed in a different issue?

MattWellie:
I think there's a second layer to this that's not been captured above...
In CramQC (and some other stages) a helper method actually plans the jobs for the workflow. Inside that helper method there's a `can_reuse` check [e.g. here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/jobs/picard.py#L277-L287)

If you force the stage to repeat, you force `queue_jobs` to run, but you can't force the individual jobs to be planned without overriding the `can_reuse` check. Doing so risks a range of other workflow stages being repeated, which you might not want. 

Not sure what the solution here is - maybe redesigning the abstract Stages so that when queue_jobs runs it's aware of whether or not the Stage is being forced, and incorporate that into the logic which plans other jobs. AFAIK this is accessible as `self.forced`, set at runtime in workflow.py.

More broadly splitting stages like this into single-tool Stages, so that there's no need to use can_reuse inside queue_jobs at all:

- The workflow DAG is established
- We evaluate the expected_outputs of each Stage, and use `can_reuse` to decide if it runs, reused, or skipped
- If it runs, we use queue_jobs to plan the individual Jobs
  - currently this is where we're using `can_reuse` to second guess the workflow planning behaviour, making it pretty inaccessible to truly force stages to repeat

PR #293, opened 2023-02-21
Fix force re-run 

vivbak:
Addresses #292 
This PR handles 2 issues. 
1. If the analysis_type is not set, then forcing re-running stages does not currently work. 
2. If sample.force is not set for each individual sample in CramQC, forcing re-running does not work. (see cram_qc.py)

Tested Here: 
https://batch.hail.populationgenomics.org.au/batches/421937 

Note: When this change is accepted there will be another accompanying PR to address adding `analysis_keys` and `analysis_type` in all other stages as well as other areas where sample.force and stage.force conflicts.

### DIFF
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -512,8 +512,8 @@ class Stage(Generic[TargetT], ABC):
                 if not all(key in outputs.data for key in self.analysis_keys):
                     raise WorkflowError(
                         f'Cannot create Analysis for stage {self.name}: `analysis_keys` '
-                        f'"{self.analysis_keys}" is not found in the expected_outputs '
-                        f'dict {outputs.data}'
+                        f'"{self.analysis_keys}" is not a subset of the expected_outputs '
+                        f'keys {outputs.data.keys()}'
                     )
 
                 for analysis_key in self.analysis_keys:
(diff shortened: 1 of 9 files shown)


cpg_workflows/workflow.py
```diff
@@ -499,36 +499,46 @@ def _queue_jobs_with_checks(
...
-                job_attrs=self.get_job_attrs(target),
-                update_analysis_meta=self.update_analysis_meta,
+            else:
+                analysis_outputs.append(outputs.data)
+
+            for analysis_output in analysis_outputs:
+                assert isinstance(
+                    analysis_output, (str, Path)
+                ), f'{analysis_output} should be a str or Path object'
+                self.status_reporter.add_updaters_jobs(
```
lgruen, line 528: I'm assuming that `add_updaters_job` will talk to Metamist? Instead of looping over all `analysis_outputs` here, should we allow writing multiple entries to Metamist in a batch? @illusional, any thoughts?

lgruen (approved):
This looks good to me, but you might want to get @illusional's feedback on the best handling of writing to Metamist (see the comment I've added).

cpg_workflows/workflow.py
vivbak: @illusional, here's a link to that add_updates_job function https://github.com/populationgenomics/production-pipelines/blob/c0197bdaa0c644fa930351e525f4a84ad72326ab/cpg_workflows/status.py#L116

illusional: Makes sense to action soon, can we raise it as an issue?

Issue #294, opened 2023-02-25
CramQC stage can not run on exomes


cassimons:
In the recent update to analysis_keys, the CramQC stage [now hard codes](https://github.com/populationgenomics/production-pipelines/blame/fa45bbb30c2c200241000699dec4b5259df76058/cpg_workflows/stages/cram_qc.py#L127) `picard_wgs_metrics` as an expected analysis_key. However, when running on exome data this stage will produce `picard_hs_metrics` not `picard_wgs_metrics`. 

In the current state, if the pipeline is run on exome data you get the following [error](https://batch.hail.populationgenomics.org.au/batches/422027/jobs/1) (edited for clarity):
```
cpg_workflows.workflow.WorkflowError: Cannot create Analysis for stage CramQC: `analysis_keys` "
    ['somalier', 'verify_bamid', ... 'quality_yield_metrics', 'picard_wgs_metrics']" 
is not a subset of the expected_outputs keys dict_keys(
    ['somalier', 'verify_bamid', ... 'quality_yield_metrics', 'picard_hs_metrics'])
```
I am not sure where this target dependent behaviour should be handled?

vivbak:
This is annoying, sorry @cassimons! Is it blocking you? If so, I'll remove these keys quickly now while I think about the best way to handle this.

cassimons:
Thanks @vivbak, that fix has got the pipeline running again for the moment.

PR #295, opened 2023-02-25
Revert analysis_keys on CramQC

vivbak:
As described https://github.com/populationgenomics/production-pipelines/issues/294

The latest change does not handle the exome case. We need to think about this more carefully, but this quick fix is proposed to minimise any blocks.


cassimons (approved):
Thanks @vivbak :-)

PR #307, opened 2023-03-08
Debugging Locally Docs 

vivbak:
Adds better better instructions for running production pipelines locally, including the requirement of an additional .toml file to mimic the analysis runner, and required environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,20 +458,74 @@ The latter script will print a section for the `gatk-sv.toml` `[images]` section
 
 ### Dev installation
 
-Assuming `python3` points to `python3.10`, set up a virtual environment:
+0. _Optional_ Assuming `python3` points to `python3.10`, set up a virtual environment:
 
-```sh
-python3 -m pip install virtualenv
-virtualenv venv
-source venv/bin/activate
-```
+    ```sh
+    python3 -m pip install virtualenv
+    virtualenv venv
+    source venv/bin/activate
+    ```
 
-Install the package, and the development environment:
+1. Install the package, and the development environment:
 
-```bash
-pip install -e .
-pip install -r requirements-dev.txt
-```
+    ```bash
+    pip install -e .
+    pip install -r requirements-dev.txt
+    ```
+
+
+2. Generate a `dev.toml` config file to mimic the analysis-runner environment. 
+
+- You can generate this with the analysis runner for example 
+
+    ```bash
+    analysis-runner config --dataset thousand-genomes --access-level test --output-dir "1kg" --config configs/genome.toml --config configs/thousand-genomes.toml
+    ``` 
+- Alternatively, you can create your own `dev.toml` file with the config that’s logged in Hail Batch. 
+
+3. In `dev.toml`, ensure the following parameter is set 
+
+    ```toml
+    [hail]
+    dry_run = True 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -458,20 +458,74 @@ The latter script will print a section for the `gatk-sv.toml` `[images]` section
...
+    ```bash
+    analysis-runner config --dataset thousand-genomes --access-level test --output-dir "1kg" --config configs/genome.toml --config configs/thousand-genomes.toml
+    ``` 
+- Alternatively, you can create your own `dev.toml` file with the config that’s logged in Hail Batch. 
+
+3. In `dev.toml`, ensure the following parameter is set 
+
+    ```toml
+    [hail]
+    dry_run = True
```
illusional, line 490:
```suggestion
    dry_run = true
```

README.md
```diff
@@ -458,20 +458,74 @@ The latter script will print a section for the `gatk-sv.toml` `[images]` section
...
+    dry_run = True 
+    ```
+
+4. Set the environment variable
+
+    `export SM_ENVIRONMENT=PRODUCTION` 
+
+5. Set the google cloud project environment variable appropriately for example 
+
+    `export GOOGLE_CLOUD_PROJECT=thousand-genomes`
```
illusional, line 499: Why does GOOGLE_CLOUD_PROJECT need to be set?

README.md
```diff
@@ -458,20 +458,74 @@ The latter script will print a section for the `gatk-sv.toml` `[images]` section
...
+            "program":"main.py",
+            "console":"integratedTerminal",
+            "python":"/Users/user/path/to/python",
+            "justMyCode": false,
+            "env": {
+                "SM_ENVIRONMENT":"PRODUCTION",
+                "GOOGLE_CLOUD_PROJECT": "thousand-genomes"
+            },
+            "args": [
+                "--config", "configs/dev.toml",
```
illusional, line 518: The other two configs are already included when you generate the above? So you could omit from here (or omit the config generation). The risk of including them in both places, is if someone specifies it in the incorrect order, it can overwrite config options (giving you a different result).

illusional (approved):
Nitpicks :)

README.md
vivbak: I know that it breaks for me otherwise, I'll dig into the exact error. My immediate suspicion is perhaps due to the existence checks?

README.md
vivbak: Great point! Think I'll exclude from the analysis-runner command.

README.md
```diff
@@ -458,20 +458,74 @@ The latter script will print a section for the `gatk-sv.toml` `[images]` section
...
+
+    ```bash
+    analysis-runner config --dataset thousand-genomes --access-level test --output-dir "1kg" --config configs/genome.toml --config configs/thousand-genomes.toml
+    ``` 
+- Alternatively, you can create your own `dev.toml` file with the config that’s logged in Hail Batch. 
+
+3. In `dev.toml`, ensure the following parameter is set 
+
+    ```toml
+    [hail]
```
illusional, line 489: Per https://centrepopgen.slack.com/archives/C030X7WGFCL/p1678745923726729, does this need to be updated?

PR #318, opened 2023-03-13
Skip stages 

vivbak:
You need to explicitly set MakeSiteOnlyVcf as a last_stage, and then skip it, in order to skip all subsequent stages. It is not enough to set 'AncestryPlots' as a last_stage otherwise all the VQSR steps will still run. 

FYI @KatalinaBobowik


danielreti (approved):
Fab!

PR #321, opened 2023-03-14
Remove colour palette which breaks AncestryPlots stage

KatalinaBobowik:
See error here: https://batch.hail.populationgenomics.org.au/batches/422961/jobs/5


vivbak (approved):
I would just love it if this worked I gotta say

PR #322, opened 2023-03-14
Rerun ancestry pca

KatalinaBobowik:
Force rerunning the Ancestry stage to regenerate PCA scores, loadings, and eigenvalues. Additionally, create a prophecy config for test to avoid constant modification of the one for main.

@vivbak could you make sure the `force_stages = ['Ancestry', 'AncestryPlots']` looks ok? It seems to have worked previously for the `AncestryPlots` stage, but not sure of that's also the case for `Ancestry`?


vivbak (approved):
LGTM!

PR #323, opened 2023-03-14
Overwrite Outputs 

vivbak:
At present, a stage can be forced, run, and then have none of the outputs overwritten, because of file existence checks. 

These existence checks should be removed completely (and will be handled in a later PR r.e. Metamist source of truth). Logically, they don't make sense. These stages are 'forced' to run, yet do not produce new outputs. 

For now, we're setting overwrite=True, so that we can finish these plots promptly!


vivbak:
FYI @violetbrina, the culprit!

KatalinaBobowik:
@vivbak this works in `test`! 👌

KatalinaBobowik (approved):
Looks great

PR #328, opened 2023-03-19
Run PCA at DenseSubset stage

KatalinaBobowik:
Run PCA at DenseSUbset stage, rather than PCA stage, as this is the point at which the sites table is used.


illusional (approved):
Symbolic approval because it looks fine but I don't really know

PR #330, opened 2023-03-22
Overwrite dense_subset.mt if it exists and force_stages is true

KatalinaBobowik:
Currently, the output from the `DenseSubset` stage is not being overwritten when using `force_stages`. 

@violetbrina @lgruen I'm not sure if this is the only thing to change but I believe setting `overwrite=True` in the `can_reuse()` function will fix it?

[Here's](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1679455741667319) a slack message with the issue for more context.

### DIFF
--- a/cpg_workflows/large_cohort/dense_subset.py
+++ b/cpg_workflows/large_cohort/dense_subset.py
@@ -15,7 +15,7 @@ def run(
     and return a dense MatrixTable with split multiallelics.
     @return: filtered and densified MatrixTable.
     """
-    if can_reuse(out_dense_mt_path):
+    if can_reuse(out_dense_mt_path, overwrite=True):
         return hl.read_matrix_table(str(out_dense_mt_path))
 
     vds = hl.vds.read_vds(str(vds_path))


cpg_workflows/large_cohort/dense_subset.py
```diff
@@ -15,7 +15,7 @@ def run(
     and return a dense MatrixTable with split multiallelics.
     @return: filtered and densified MatrixTable.
     """
-    if can_reuse(out_dense_mt_path):
+    if can_reuse(out_dense_mt_path, overwrite=True):
```
lgruen, line 18: @violetbrina Why does `can_reuse` have these weird semantics?

https://github.com/populationgenomics/production-pipelines/blob/f25a8086a4f1756495faa64c1ace37df694e4851/cpg_workflows/utils.py#L70-L81

It seems odd not having to specify an `overwrite` parameter (as that will _always_ return `False`), so why is it an optional parameter? And should the value of `overwrite` within the context of `production-pipelines` come from a config value instead of hardcoding it (like setting it to `True` here)?

violetbrina: An excellent question. Yeah probably should be configurable. I've had a look through the code and there doesn't seem to be a clear answer. I guess it's worth working out if we should consider this scenario for the metamist/production pipelines integration. Instead of doing hard file checks like this I guess we should ask metamist if a file should be reused or not and control the logic on that end.

Happy to approve for now in order for Kat to get this done.

violetbrina (approved):

PR #331, opened 2023-03-23
Expose parameters

KatalinaBobowik:
This PR addresses a few hard-coded parameters that should instead be configurable input (exposing the number of PCs for the VerifyBamID job; allowing for assorted input in the Picard job; exposing the reference fasta; and exposing the scatter_count parameter in the genotyping job).

### DIFF
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -1,7 +1,9 @@
 [workflow]
 name = 'large_cohort'
-scatter_count = 50
+scatter_count_genotype = 50
 status_reporter = 'metamist'
+# define reference fasta
+ref_fasta='gs://gcp-public-data--broad-references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.fasta'
 
 # Realign CRAM when available, instead of using FASTQ.
 # The parameter value should correspond to CRAM version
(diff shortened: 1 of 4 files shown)


configs/defaults/large_cohort.toml
```diff
@@ -1,7 +1,9 @@
 [workflow]
 name = 'large_cohort'
-scatter_count = 50
+scatter_count_genotype = 50
 status_reporter = 'metamist'
+# define reference fasta
+ref_fasta='gs://gcp-public-data--broad-references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.fasta'
```
illusional, line 6: I don't mind it being overrideable, but feels awkward to constantly override it to a public source when we do have it in the bucket:

```
config['references']['broad']['ref_fasta'] = "gs://cpg-common-main/references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.fasta"
```

cpg_workflows/jobs/genotype.py
```diff
@@ -41,7 +41,7 @@ def genotype(
         output_path=hc_gvcf_path,
         cram_path=cram_path,
         tmp_prefix=tmp_prefix,
-        scatter_count=50,
+        scatter_count=get_config()['workflow'].get('scatter_count_genotype'),
```
illusional, line 44: Can this supply the default 50 here?
```suggestion
        scatter_count=get_config()['workflow'].get('scatter_count_genotype', 50),
```

illusional (approved):
Looks good, a few comments to address before merging though (but don't require additional review)

configs/defaults/large_cohort.toml
KatalinaBobowik: I've given the cpg bucket path instead here 👌

cassimons (commented):
This PR has broken the seqr pipeline as it is defining generic configs in the `large_cohort` section. 

Can this please be rolled back and the exposed configs put in a pipeline-wide config section + added to defaults?

PR #340, opened 2023-03-24
Move conifg commands to more universal location

KatalinaBobowik:
This addresses the issue caused by [this](https://github.com/populationgenomics/production-pipelines/pull/331/files) PR and discussed in [this ](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1679624807999289)slack post. 

@cassimons seeing as both commands occur within the cramqc stage, I've put both parameters in the `[cramqc]` section of the configs, but happy to change if you think they'd be placed better elsewhere.


cassimons (approved):
Thanks @KatalinaBobowik :-) Looks great.

PR #349, opened 2023-04-18
Fix test/test_large_cohort.py unit test

jmarshall:
PR #326 changed `predetermined_qc_variants` so that its path is read from an external config file. However for these unit tests, there is no config other than what is hard-coded in the `TOML` string or merged from _cpg_workflows/defaults.toml_ and _configs/defaults/large_cohort.toml_.

Revert (most of) the previous PR so that this path is hard-coded, just as the other paths in `TOML` are. Thus it refers to _test/data/large_cohort/…/pre_ld_pruning_qc_variants.ht_ within this repository.

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -250,10 +250,11 @@ def _infer_pop_labels(
 
     if training_pop_ht.count() < 2:
         logging.warning(
-            'Need at least 2 samples with known `population` label to run PCA'
+            'Need at least 2 samples with known `population` label to run PCA '
             'and assign population labels to remaining samples'
         )
         pop_ht = scores_ht.annotate(
+            training_pop=hl.missing(hl.tstr),
             pop='Other',
             is_training=False,
             pca_scores=scores_ht.scores,
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -98,7 +98,8 @@ def run(
     sample_qc_ht = hl.read_table(str(sample_qc_ht_path))
     relateds_to_drop_ht = hl.read_table(str(relateds_to_drop_ht_path))
 
-    if pca_background := get_config()['large_cohort'].get('pca_background', {}):
+    pca_background = get_config()['large_cohort'].get('pca_background', {})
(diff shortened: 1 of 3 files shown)


jmarshall:
The _large_cohort_ test still fails, but now for a more interesting reason:

```
>       for dataset in get_config()['large_cohort']['pca_background']['datasets']:
E       KeyError: 'datasets'
```

It seems that we need to either disable the pca_background stuff (?) or perhaps add some test data files (or just point to files that are already present) so that these steps can run in this test case.

vivbak:
On inspection it looks like this condition isn't great 
https://github.com/populationgenomics/production-pipelines/blob/89c68f2fea9cdf5757c61cf743582d795f8aa7a3/cpg_workflows/large_cohort/ancestry_pca.py#L101 

AFAIK, having a background dataset isn't always required. So we shouldn't attempt to hit the `add_background` function if `pca_background` isn't set in the config, and indeed never hit that subsequent key error. 

In any case, our tests should also test both cases.

jmarshall:
This is now ready for re-review. The large_cohort test case now passes, but seqr_loader_dry still times out (at least for this PR) and I haven't looked at that.

There were six separate problems with this unit test! Please review in particular the changes outwith _test_large_cohort.py_, as these affect production pipeline behaviour:

* Changed the `add_background()` conditional, as discussed. 
* Added a dummy `training_pop` column when there are too few known-population samples in `_infer_pop_labels`.

  I don't know whether it is best to use a missing value like this, `Other` or similar as is used for the `pop` column, or to improve this line of _ancestry_plots.py_ to deal with the absence of this field:
https://github.com/populationgenomics/production-pipelines/blob/ac79102af8cb00ab6e2ffe1df9f50ea8bf1d7d7b/cpg_workflows/large_cohort/ancestry_plots.py#L49

* In _ancestry_plots.py_, use `dataset` for the plot labels if `input_datasets` is omitted from the config. (This mirrors what `create_cohort()` does.)

jmarshall:
> In any case, our tests should also cover both cases.

Eventually they probably should, but I don't think I'm qualified to add the background-dataset-is-provided case :smile:
For now, this PR is just to unbreak CI. I could file an issue to record this shortcoming…

vivbak (approved):
Awesome! LGTM.

PR #352, opened 2023-04-20
Investigating test/test_seqr_loader_dry.py unit test

jmarshall:
This test started timing out on April 12th: [test run 631](https://github.com/populationgenomics/production-pipelines/actions/runs/4667615635/jobs/8263639017) is the last successful one, while [test run 632](https://github.com/populationgenomics/production-pipelines/actions/runs/4673207804/jobs/8276186524) was the first to time out.

The difference appears to be in the set of python packages selected:

```diff
--- installed-631-ok
+++ installed-632-borked

+aiodns-2.0.0
-aiohttp-session-2.12.0
+pycares-4.3.0

-boto3-1.26.110
-botocore-1.29.110
+boto3-1.26.111
+botocore-1.29.111
-google-auth-2.14.1
+google-auth-2.17.2
-hail-0.2.112
+hail-0.2.113
-plotly-5.10.0
+plotly-5.14.1
-sample-metadata-5.6.0
+sample-metadata-5.6.1
```

I suspect the cause of the job timeout is a DNS request that is being firewalled on the GitHub Actions host, coupled with code that isn't aborting its request after a sensible timeout. And I'm guessing that this has been caused by the change in DNS-related python packages at the bottom of this tower of dependencies…

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,7 +45,7 @@ jobs:
           pip show aiodns aiohttp aiohttp_session google-auth hail pycares
 
       - name: Run unit tests
-        timeout-minutes: 5
+        timeout-minutes: 3
         run: |
           PYTHONPATH=$PWD/gnomad_methods:$PWD/seqr-loading-pipelines \
           python -m pytest -s test/test_${{ matrix.suffix }}.py
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -44,10 +44,10 @@ jobs:
         run: |
           pip show aiodns aiohttp aiohttp_session google-auth hail pycares
 
-      - name: Remove aiodns
+      - name: Remove borkers
         continue-on-error: true
         run: |
-          pip uninstall -y aiodns
+          pip uninstall -y aiodns pycares
 
       - name: Run unit tests
         timeout-minutes: 5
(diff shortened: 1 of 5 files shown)


jmarshall:
That debugging reports that aiodns was pulled in directly by hail:

```
Name: aiodns
Version: 2.0.0
Summary: Simple DNS resolver for asyncio
Home-page: http://github.com/saghul/aiodns
[…]
Requires: pycares
Required-by: hail
```

So this looks like it may be more fallout from the recent hail update.

jmarshall:
Hmmm… aiodns and pycares seem to be a red herring. Really it's just hail 0.2.112 good vs 0.2.113 bad.

Fortunately hail itself is a small package, so the problem shouldn't take long to find… 🙄

jmarshall:
Confirmed from this profiling log that:

* With 0.2.112, the job calls `schedule_job` 38302 times and  terminates successfully after ~20 seconds;
* With 0.2.113, the job calls `schedule_job` 38302 times, and then runs forever repeatedly calling `add_dependents()`.

The `add_dependents()` function did not exist in 0.2.112. It was added on the way to 0.2.113 in hail-is/hail#12780 (so the problem is limited to the local backend, so may not cause a big production issue). It is a recursive function, and appears not to be able to deal with loops in its data structure's graph. (Or it's not actually infinite, but “merely” that our test set is big enough that there is a combinatorial explosion.) Whether our test case ought not to contain loops, or whether the function needs to be able to deal with loops is… unknown by me.

jmarshall:
Hail issue that caused this now fixed and should appear in 0.2.115 in due course.

PR #353, opened 2023-04-20
Add analysis record for AnnotateDataset job.

cassimons:
Adds a analyis record for annotated datasets. This will enable an easy comparison of new annotated datasets to the most recent ES index to detect if new samples have been added (suggesting and ES index build should be triggered).

This is a new analyis 'type', but as per this thread (https://centrepopgen.slack.com/archives/C020PDNH3N0/p1680127086008289), static analyis types are on the way out so in the interum this is implimneted using a custom type.

### DIFF
--- a/cpg_workflows/stages/seqr_loader.py
+++ b/cpg_workflows/stages/seqr_loader.py
@@ -3,6 +3,7 @@
 """
 Hail Query stages for the Seqr loader workflow.
 """
+from typing import Any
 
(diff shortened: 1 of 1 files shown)


illusional (commented):
No objection from me, decoupling of types is being implemented in https://github.com/populationgenomics/sample-metadata/pull/378

vivbak (approved):

PR #354, opened 2023-04-24
Use current GitHub Action versions and set test timeout

jmarshall:
Prevent "Node.js 12 actions are deprecated" warnings by updating to the current releases of Actions used. Use the current releases instead of `@main` too, as we don't want to be affected by bleeding-edge GitHub Actions bugs.

Set a 20 minute timeout instead of the default six hours. The slowest test currently takes ~5 minutes, but seqr_loader_dry has recently been affected by an infinite loop in hail — so we'd prefer not to wait for several hours to see it time out.

Seqr_loader_dry currently times out due to hail-is/hail#12915, for which a fix has been made and will likely appear in hail 0.2.115. We could perhaps pin the older hail 0.2.112 in `extras_require[test]`, but I don't think it's worth fiddling with that: we can just wait until 0.2.115 is released and pip selects that one in place of the affected 0.2.113 and 0.2.114.


vivbak (approved):
:) Thanks for the clarity on the current test failures! Changes to the GH Action LGTM.

PR #358, opened 2023-05-05
add a new stage to the seqr_loader

MattWellie:
Partial solution to https://github.com/populationgenomics/seqr-private/issues/93

Adds a new stage after `AnnotateDataset` which will ingest the per-dataset MatrixTable, and write out to 

`<bucket>/vcf/<massive_hash>-<dataset_name>.vcf.bgz`

Inserted a new config variable to control which datasets this applies to, most cohorts will just return `None` - currently we only need to export the UDN-Aus and Validation datasets.

Updates the meta for the analysis entry in Metamist to be `type:dataset-vcf`

### DIFF
--- a/configs/defaults/seqr_loader.toml
+++ b/configs/defaults/seqr_loader.toml
@@ -22,7 +22,7 @@ use_as_vqsr = true
 # create indices for all input datasets.
 #create_es_index_for_datasets = []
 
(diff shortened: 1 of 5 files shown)


cassimons (approved):
LGTM thanks.

PR #360, opened 2023-05-10
Fragment workflows

MattWellie:
First step of implementing the more complex Cohort Batching (https://github.com/broadinstitute/gatk-sv#cohort-mode)

The intention (as I understand it) is:

1. Support batching of samples from multiple different projects
2. Each run should only pick up samples which do not yet have SV data
3. The workflow will run in discrete steps:

* Step 1 - generate individual calls for all samples without SV data, run QC within this group, and obtain homologous batches
* Step 2 - run Batch clustering and filtering separately on each batch defined in step 1
* Step 3 - Combine the FilterBatch outputs from all individual batches into a single merged version of each file (see issue 530 on gatk-sv)
* Step 4 - Run GenotypeBatch in the same sample groups as Step 2, using the output from Step 3 in each
* Step 5 - feed results from each separate batch into a single workflow of MakeCohortVCF & AnnotateVCF (not yet fiddled with)

So far this adapts Steps 1 & 2:

1. Throws all the GATK-SV content into a folder within `cpg_workflows/stages`, just for tidiness
2. Moves all the single sample Stages into a separate file, `cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py`
3. Converts all the DatasetStage definitions into CohortStage definitions, which should permit running on samples across multiple projects. Data will be written to the top level `project`. Probably `seqr` in most cases...
4. Changes the combined PED file generation to be cross-project, not per-project
5. Writes outputs into `self.prefix` for all cohort stages, which includes a hash of the included sample IDs
6. Writes outputs into `<sample.dataset>/gatk_sv/...` for all Sample Stages

In the current CohortStage form I reckon this will be viable at the moment, though it will require extensive use of `only_samples` in config to split out the different parallel workflows.

The `Step 3` merging script looks very simple to re-write, I'm just not sure why it's been written the way it has.


cassimons:
Looks like a good plan @MattWellie. Unless the timeline for metamist cohort builder moves up, we are going to have to cludge this inline for the interim using something like `only_samples`. Launch one (hail) batch  per (gatk-sv) batch? 

@jmarshall will need to cover similar ground with the exome CNVs.

MattWellie:
> something like `only_samples`

That sounds about right, there would be a few cludgey steps involved... Without any funky metamist cohort selection it would probably be:

1. Find all samples without a SV analysis entry
2. Run this workflow restricted to only those samples
3. Results in a JSON file containing all the separate batched sample IDs to run, so use that to generate a bunch more sub-batches with `only_samples`
4. Draw the rest of the owl

cassimons:
Elegance in action!

MattWellie:
Superseded by #363

PR #363, opened 2023-05-16
Fragment workflows final workflow

MattWellie:
This is a draft of the changes that we'd need to run this whole process e2e, including the fancy cross-batch aggregation, and limiting samples based entirely on the `only_samples` config setting. 

Sooooo.... Off the top of my head I can remember these changes:

1. GATK-sv is 4/5 workflows instead of 1:
    * Single Sample - individual calling into sample batching (`GatherSampleEvidence` -> `CreateSampleBatches`)
    * MultiSample 1 - sub-cohort clustering and filtering (`GatherBatchEvidence` -> `FilterBatch`)
    * Sandwich - Combines FilterBatch results across all sub cohorts (`MergeBatchSites`)
    * MultiSample 2 - continues MultiSample 1, but with the Sandwich input (`GenotypeBatch`)
    * MultiSample 3 - runs as a single cohort, using the outputs of all MultiSample 2 sub-batches (`MakeCohortVCF` -> `AnnotateVCF`)

2. Adds a new MergeBatchSites stage, the WDL used to combine all sub-batch depth and pesr files into a single file
3. Adds the batch division script, modelled on the Broad Version (tweaked a bit to work with realistic CPG sample counts, runs as a PythonJob - not sure if I've implemented this correctly)
4. Adds a workflow definition and workflow config for each version of the fragmented workflow
5. Added a couple of tiny configs/templates (run until FilterBatch, only run GenotypeBatch, configs to contain a list of sub-cohort hashes)
6. Changes EVERYTHING (except individual calling) to be a CohortStage instead of DatasetStage. This requires hardcore use  of `only_samples` until fancy cohorts are built in.
7. Puts all the GATK-SV stage definitions into a separate folder
8. Creates a `sample.make_sv_evidence_path` property, which is `sample_dataset_bucket/sv_evidence`, to store the individual call data
9. Combined PED file generation is done at the cohort level instead of the dataset level
10. All other outputs are pathed out to `config.prefix`, instead of manually pulling together output paths from the `bucket / Stage / etc...`
11. Fixed the CyVCF2 installation at `cyvcf2==0.30.18` - 0.30.20 fails to install which was ruining all the CI

---

When we move to fancy cohorts:

1. Make the SampleStage a SequencingGroupStage 
2. Split these `CohortStage`s into MultiCohortStage (`MergeBatchSites`, `MakeCohortVCF`, `AnnotateVCF`) and CohortStages (everything else)
3. Remove the workflow fragmentation, and have the MultiCohortStages track down input based on the component CohortStage outputs

--- 

Test run: 

gs://cpg-broad-rgp-test/gatk_sv/b1b0be5754cfebd117b2a0e50b2d4087fe1af6_100

(I've copied the outputs from the previous execution folders in gs://cpg-broad-rgp-test/gatk_sv and checked that this workflow 'dry runs' at all stages)

- CreateSampleBatches as a PythonJob works great... This cohort used was at the minimum required size, so it only generated a single cohort... Not very interesting, but it does run to completion
- Likewise MergeBatchSites was only 'merging' one file, so that's not an interesting usage but it does work
- Providing all the different cohort hashes in config, and iteratively building paths to input files from previous unconnected batches works fine
- BASICALLY it all works as expected


cassimons (approved):
It would be great to add a bit more of a summary of functionally what the batching will do. Other than that it looks good.

configs/defaults/gatk_sv.toml
```diff
@@ -5,8 +5,7 @@ dataset = 'seqr'
 ref_fasta = 'gs://cpg-common-main/references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.fasta'
 
 [images]
-
-cpg_workflows = "australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:pr-345"
+cpg_workflows = "australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:1.14.0"
```
vivbak, line 8: Is this file used at all anymore?

configs/defaults/gatk_sv_multisample_1.toml
```diff
@@ -0,0 +1,29 @@
...
+sv_base_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv-base:v0.27.0"
+sv_pipeline_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv-pipeline:v0.27.0"
+sv_base_mini_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv-base-mini:v0.27.0"
+
+# variant callers
+manta_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv/manta:5994670"
+scramble_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv/scramble:1.0.2"
+wham_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv/wham:5994670"
+
+[sv_ref_panel]
```
vivbak, line 27: I assume params like this one are consistent across all the new gatk_sv configs. Wondering if there's a better way to not repeat this? Perhaps we can leverage that generic gatk_sv.toml if not already.

vivbak (approved):
Consistently love your work, Matt! Just wondering if there's a cleaner way to deal with these configs, they look a bit repetitive.

vivbak (approved):
Consistently love your work, Matt! Just wondering if there's a cleaner way to deal with these configs, they look a bit repetitive.

vivbak:
Not sure why my comment appeared twice, but you deserve 2x appreciation anyway.

configs/defaults/gatk_sv_multisample_1.toml
MattWellie: I've pushed something sketchy for this. It's a bummer at the moment splitting this across different 'workflows', but eventually all this will be replaced with just two workflows, at which point all the silly configs will be deleted

cpg_workflows/jobs/sample_batching.py
```diff
@@ -0,0 +1,134 @@
...
+
+def batch_samples(
+    md: pd.DataFrame,
+    batch_size: int = 200,
+    min_batch_size: int = 100,
+    max_batch_size: int = 300,
+    cov_bins: int = 1,
+):
+    """
+    Batch samples by coverage, dosage bias, and chrX ploidy
```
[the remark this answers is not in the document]
MattWellie: See https://github.com/populationgenomics/production-pipelines/pull/374
Thoughts are that we won't have the sample count on a regular basis to hit the 100-sample minimum batch size while grouping by coverage then sub-grouping by WGD (minimum 400 samples required), so we're best off dumping WGD for now and revisiting later

cpg_workflows/jobs/sample_batching.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Runs this process
+    - loads the input data
+    - obtains batches
+    - writes out the data into one single cross-batch file
+    - also writes out per-batch sample lists
+
+    Args:
+        metadata_file (str): path to the metadata file
+        output_json (str): location to write the batched samples outå
```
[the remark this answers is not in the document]
MattWellie: I actually don't know how that character åppeared.
Will be deleted if we merge #374

cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
```diff
@@ -0,0 +1,249 @@
...
+        if get_config()['workflow'].get('sequencing_type') != 'genome':
+            sample_types = {'exome': [sam.id for sam in cohort.get_samples()]}
+        # within exomes, divide into PCR- and all other samples
+        else:
+            sample_types = {'positive': [], 'negative': []}
+            for sample in cohort.get_samples():
+                if sample.meta.get('pcr_status', 'unknown') == 'negative':
+                    sample_types['negative'].append(sample.id)
+                else:
+                    sample_types['positive'].append(sample.id)
```
MattWellie, line 221: @EddieLF can you check this logic in terms of keywords, location in meta/database schema, see if it gels with what you are thinking?

cpg_workflows/jobs/sample_batching.py
```diff
@@ -0,0 +1,167 @@
...
+    # read in the metadata contents
+    logging.basicConfig(level=logging.INFO)
+
+    # load in the metadata file
+    md = pd.read_csv(metadata_file, sep='\t', low_memory=False)
+    md.columns = [x.replace('#', '') for x in md.columns]
+
+    # filter to the PCR-state samples we're interested in
+    # surely there's a neater way to do this...
+    md = md[np.array([sam in sample_ids for sam in md.ID.tolist()])]
```
EddieLF, line 150: Perhaps `md = md[md['ID'].isin(sample_ids)]` or `md = md.query("ID in @sample_ids")` ?

MattWellie: Nice!

MattWellie:
Final comment - This is still waiting for various PCR status labels to be populated into/recalled from Metamist. At this stage we are in a transitional period for sample -> sequencing groups, so it makes sense to run this on a test-basis for now, and update fully with PCR state etc. when the rest of the transformational changes are going on

PR #365, opened 2023-05-29
Add gCNV workflow

jmarshall:
Currently very draft and incomplete, but it is more than time to start running it via `analysis-runner` and watch it fall over properly…


configs/defaults/gcnv.toml
```diff
@@ -0,0 +1,7 @@
+[workflow]
+intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.v1.interval_list'
```
MattWellie, line 2: This file is already part of the core reference set, so you can pull from there

```toml

[references.broad]
dragmap_prefix = "gs://cpg-common-main/references/hg38/v0/dragen_reference"
...
exome_calling_interval_lists = "gs://cpg-common-main/references/hg38/v0/exome_calling_regions.v1.interval_list"
```

configs/defaults/gcnv.toml
```diff
@@ -0,0 +1,7 @@
...
+# TODO Where can this file come from, and does it need to vary
+ploidy_priors = 'gs://cpg-common-main/ploidy-priors-TODO.tsv'
```
MattWellie: Similarly there's a ploidy priors under

```toml
[references.gatk_sv]
...
contig_ploidy_priors = "gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv"
```

I'm not sure if this differs from the one required for exomes

cpg_workflows/stages/gcnv.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@stage(required_stages=Align)
+class PrepareIntervals(DatasetStage):
+    def _intervals_file(self, dataset):
+        h = dataset.alignment_inputs_hash()
+        return dataset.prefix() / 'gcnv' / h / f'{dataset.name}.interval_list'
+
+    def expected_outputs(self, dataset):
+        return {'intervals': self._intervals_file(dataset)}
```
MattWellie, line 19: A couple of thoughts here... Happy to discuss on all accounts.

* The Dataset Stage will always run once per Dataset, so splitting off the intervals file generation into a private method with a dataset as an argument feels a little redundant. The logic is sound but I don't think it provides an explicit benefit.

The prefix setting is interesting for 2 reasons:

1. The alignment hash for the workflow can be overridden (see [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/workflow.py#L857-L884)), but by explicitly calculating it you'd potentially be overriding the override.
2. Instead of building on `dataset.prefix()` have a look at `self.prefix`:
    - returns `workflow.prefix` / name (Stage name, so `PrepareIntervals`) 
    - Workflow prefix is:
        - bucket the workflow is running under (e.g. if you're running across multiple projects as a higher level dataset, it will be that higher dataset bucket. If you're running the workflow for one dataset, this will == that project bucket)
        - workflow name (here `gcnv`)
        - alignments hash for the workflow (can be set explicitly)
    - example: `gs://cpg-broad-rgp-test/gatk_sv/b1b0be5754cfebd117b2a0e50b2d4087fe1af6_100/ClusterBatch/<files>`

cpg_workflows/stages/gcnv.py
```diff
@@ -0,0 +1,76 @@
...
+class CollectReadCounts(SampleStage):
+    """
+    Per-sample stage that produces .counts.hdf5 files.
+    """
+
+    def _counts_file(self, sample):
+        return sample.dataset.prefix() / 'gcnv' / f'{sample.id}.counts.hdf5'
+
+    def expected_outputs(self, sample):
+        return {'gcnv_counts': self._counts_file(sample)}
```
MattWellie, line 41: similar to comment above

MattWellie (commented):
thoughts

configs/defaults/gcnv.toml
MattWellie: I'll track down the better way of getting these, but you can use the `config` mode of analysis-runner to generate the combined config your current run will create, e.g. 
```
analysis-runner config --dataset broad-rgp -o here.toml --access-level test --config configs/genome.toml --config configs/gatk-sv-broad-rgp-test.toml
```

This will generate the exact run config you'd create by submitting a workflow run instruction + those two manually provided configs

configs/defaults/gcnv.toml
jmarshall: Yes, I tracked that one down by looking in the bucket. It's a simple enough file that I think it'll be fine for exomes.

configs/defaults/gcnv.toml
jmarshall: Um, yes, that's the `gs:` URL I copied in there?

Unless you're suggesting to just use `gs://cpg-common-main/references/…/preprocessed_intervals.interval_list` rather than generating it in the `PrepareIntervals` stage?

cpg_workflows/jobs/gatk.py
```diff
@@ -0,0 +1,73 @@
...
+def collect_read_counts(b, sample, intervals, job_attrs, output_path):
+    j = b.new_job(f'Collect gCNV counts for {sample.id}', job_attrs)
+    j.image(image_path('gatk'))
+
+    reference = fasta_res_group(b)
+    cmd = f"""
+    gatk CollectReadCounts \\
+      --reference {reference.base} --intervals {intervals} \\
+      --interval-merging-rule OVERLAPPING_ONLY \\
+      --input {sample.bam} --read-index {sample.bai} \\
```
jmarshall, line 47: Clearly this line needs updating to how you actually get the CRAM and CRAI path out of a sample.

However I just came across this commentary in `_haplotype_caller_one` in _jobs/genotype.py_:

```python
# …can't pass GCS URL to CRAM to gatk directly because we will hit GCP egress bandwidth limit…
…
# Retrying copying to avoid google bandwidth limits
retry_gs_cp {str(cram_path.path)} $CRAM
retry_gs_cp {str(cram_path.index_path)} $CRAI
```

Uh oh! Perhaps this step will also need to make a temporary local copy of the CRAM file?

MattWellie: That sounds reasonable!

configs/defaults/gcnv.toml
jmarshall: Oh, you mean access it via the existing config key using something like `reference_path('broad/exome_calling_interval_lists')`.

Probably. One would have to think about what would change if we ever switched target intervals…

MattWellie: Yeah sorry, I wasn't clear about what I meant here. This exact file is already part of the standard references so it can be pulled from there. It's a really minor nit so I wouldn't worry about it

jmarshall:
Some statistics on having CollectReadCounts produce its output as HDF5 versus TSV, based on the 11 samples currently running in acute-care-test:

As HDF5, each sample's .counts.hdf5 file is 8.4 MB, needs to be explicitly copied to local storage for DetermineGermlineContigPloidy to run, and is read from local disk by DetermineGermlineContigPloidy in <1 second.

As .tsv.gz (bgzipped and IndexFeatureFile'd), each sample's file is ~2.5 MB (7.4 MB uncompressed) with ~1.1 MB of associated .tbi index, can be read directly from GCS by this GATK subcommand, and is read by DetermineGermlineContigPloidy in ~15 seconds.

I have not measured the time taken to copy the files locally, and it would be interesting to see how long DetermineGermlineContigPloidy would take to read local .tsv.gz files.

PR #367, opened 2023-06-01
link stage outputs correctly

MattWellie:
The original implementation tries to find the outputs from the AnnotateDataset cohort by passing the `Cohort`, instead of passing the `Dataset`. This will probably cause the whole thing to explode.

Related: from recent runs the `validation` & `udn-aus` datasets have been removed from the default seqr-loader config
https://batch.hail.populationgenomics.org.au/batches/424408/jobs/1
https://github.com/populationgenomics/production-pipelines/blob/main/configs/defaults/seqr_loader.toml#L25

Was this removed because the stage was failing?

- From discussion on slack, this stage was deliberately blocked off, probably because of the aforementioned failure
- After this is fixed, we can use the output_prefix override to continue from the latest complete index and just generate VCFs

### DIFF
--- a/configs/defaults/seqr_loader.toml
+++ b/configs/defaults/seqr_loader.toml
@@ -22,7 +22,7 @@ use_as_vqsr = true
 # create indices for all input datasets.
 #create_es_index_for_datasets = []
 
-write_vcf = ["udn-aus", "validation"]
+write_vcf = ["udn-aus"]
 
 [vqsr]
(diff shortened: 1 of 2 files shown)


cassimons:
Thanks Matt. Yes it died and I assumed this was the cause - I just had not had a chance to go back to it. Thanks for digging it out.

cassimons (approved):

PR #369, opened 2023-06-02
Local Testing

daniaki:
Fixed a bunch of bugs in the test suite caused by persisting global state variables (workflow, batches, configs etc) between tests. Refactored the test code to make sure config, pre-existing stage results and other persistent state are not re-used between tests. Tests should be running in isolated environments now.

Also added dev docs on how to write a unit test using some helper set-up functions that I added.


test/test_seqr_loader_dry.py
```diff
@@ -161,11 +155,29 @@ def _mock_cohort():
...
+def selective_mock_open(*args, **kwargs):
+    if str(args[0]).endswith('.toml'):
+        # Don't mock calls to load a config file
+        if isinstance(args[0], Path):
+            if len(args) > 1:
+                return open(str(args[0]), *args[1:], **kwargs)
+            return open(str(args[0]), **kwargs)
+        return open(*args, **kwargs)
+    else:
+        return mock_open(read_data="<stub>")(*args, **kwargs)
```
MattWellie: Genius. I was making tracks with the `tmp_path`/set_config_paths but this mocked open killed me. Good solution.

.github/workflows/test.yaml
```diff
@@ -44,4 +29,4 @@ jobs:
         timeout-minutes: 20
         run: |
           PYTHONPATH=$PWD/gnomad_methods:$PWD/seqr-loading-pipelines \
-          python -m pytest test/test_${{ matrix.suffix }}.py
+          pytest test
```
MattWellie, line 32: With `pytest-xdist` in the dev requirements we can parallelise out all these tests per core. It looks like the super long running large_cohort workflow test is still a bottleneck at the moment, but if/when that's trimmed down the execution times will drop

```suggestion
          pytest -n auto test
```

MattWellie (approved):

.github/workflows/test.yaml
MattWellie: running it locally the parallel version actually takes a few seconds longer 😓 the bottlenecking around that one test runtime + the few seconds overhead setting up the parallel runs + low overall number of tests. Ignore this!

daniaki: sadface

test/test_seqr_loader_dry.py
daniaki: Yeah, this one had me scratching my head for a while! Got the idea from Stack, bless 'em.

vivbak (commented):
This looks amazing @daniaki, thanks! I can't seem to get all of them working locally though. Keen to figure out whats up before we merge (though it's likely something small on my end)

vivbak (approved):
My issues are resolved. Conda is awful xo

test/conftest.py
```diff
@@ -10,8 +12,15 @@
 
 @pytest.fixture(autouse=True, scope='function')
 def pre_and_post_test():
+    # Set a dummy google cloud project to avoid errors when running tests for tests
+    # that use the google cloud.
+    os.environ["GOOGLE_CLOUD_PROJECT"] = "dummy-project-for-tests"
```
MattWellie, line 17: you could set/unset this using a monkeypatch

daniaki: I replaced this with a `mock.patch.dict`. It's just a temporary solution while we figure out how to properly test production pipelines.

PR #372, opened 2023-06-06
Resume task functionality for HaplotypeCaller

MattWellie:
- Give each fragment a location, the same way we do for the `scatters == 1` condition (i.e. write output to tmp upon completion)
- `_haplotype_caller_one` returns either the resource group (if a new one was generated) or a String filepath (if the output was reusable)
- The merge method receives the list of [String | ResourceGroups] 
- When building the input command, if the value is a String, read into the batch before adding the location to the merge cmd

reasoning:
- Without a write location (out_gvcf_path) the method doesn't persist the output to GCP, so we can never do a reuse check
- `_haplotype_caller_one` returning a specific output, whether the job runs or not, means that every interval (in order) has its output added to a list. This was found previously by doing `[j.output_gvcf for j in hc_jobs]`, but the different Hail structure between reading input files and declaring a resource group means that we can never satisfy this syntax requirement with new files read in from GCP.

```
when reading input files in, "resourceGroup.something" is always a file
when _declaring_ a resource group, "resourceGroup.something" is always a collection of files
```

- Passing results back as a String instead of a Path
  - A `str` is a much more specific test than Path (CloudPath | GCPath | Path)
  - `read_input_group` needs a String to work anyway

### DIFF
--- a/cpg_workflows/jobs/genotype.py
+++ b/cpg_workflows/jobs/genotype.py
@@ -252,10 +252,7 @@ def merge_gvcfs_job(
         # if the output was recoverable, read into the batch
         if isinstance(gvcf_group, str):
             gvcf_group = b.read_input_group(
-                **{
-                    "g.vcf.gz": gvcf_group,
-                    "g.vcf.gz.tbi": gvcf_group
-                }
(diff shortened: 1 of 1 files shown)


cassimons (approved):
Thanks Matt. Tested by canceling/resuming some runs and it and it does what we need it to 👍

PR #374, opened 2023-06-06
Simplify batching

MattWellie:
A re-think of the batch process. Simplified, better comments, a test case. The involvement of WGD in batching is removed as we're using smaller cohorts of samples, so we don't necessarily have the numbers required to batch on multiple different parameters. 

This doesn't address the PCR +/- state as the data isn't currently available here. We could add this in at sample batching if we can draw from another source, or we could use PCR status in the custom cohort builder (when that functionality comes around) to ensure that this stage will only run with either PCR- or PCR+/unknown samples, meaning that the batch generation doesn't need to consider it.

Process:
1. split samples by sex (inferred from chrX ploidy)
2. order samples by median coverage
3. calculate appropriate number of batches to get the # samples in each batch within parameters
4. split each sex-list into equal parts, one for each batch
5. create batches by joining lowest cov. male &female samples, up to highest coverage male & female
6. annotate each batch with mf ratio, coverage medians
    - thinking here is that if we need to drop additional samples into a batch later, this metadata will help determine the best choice instead of requiring a complete re-run


cassimons:
Can you assume there is a metamist field: `pcr_status` [positive, negative, unknown] and add that to the batching logic. At the moment I think we should group the unknowns with the +ve's. I will talk to Eddie about populating the field in metamist.

MattWellie:
> assume there is a metamist field: `pcr_status`

Will this be in the Meta, or as a top-level attribute in the database? Should be easy enough to adjust, might have to immediately change when we shuffle to SequencingGroups... but so will everything else. 
Will throw this on Slack for discussion

PR #375, opened 2023-06-07
Speed up pipeline startup

MattWellie:
This PR is an attempt to speed up pipeline initiation in two ways:

1. Instead of running existence checks on each individual file in the per-sample stage outputs (gVCF, alignment, QC), update this to find and listdir in each of the output directories, then check for existence of the output file(s) in the already collected files. This should scale far better, so long as we stick with the current paradigm of adding single-sample outputs into a per-cohort folder (e.g. `gs://cpg-mcri-lrp-main/exome/cram/CPG289413.cram`)
2. When genotyping jobs (calling per interval, merging, post-processing) are required, the current `main` code creates a Batch job for each task, with all the corresponding dependency setting. This occurs even if the job doesn't do anything (i.e. `[reuse]` output). This change alters this behaviour so that the dummy jobs are not created, and the Stage(s) permit tasks to be `None`. 

At the sample Stage, before running existence checks on all SampleStage outputs individually:
- find all per-sample output files using the `expected_outputs` call
- find all unique directories containing those files
- list all contents in all those directories (here using `iterdir`, previously I've used `glob` to do the same thing. I've done a little check and it produces the same output)
- pass this as a [optional] set of Paths to `_get_action`, forwarded to `_is_reusable`
- if a set of existing Paths was given as an argument, instead of doing existence checks, check for existence in the list of files
- the rest of the does-the-output-exist logic is unaffected

e.g. 

```
@StageMakeABC

samples CPG1, CPG2

CPG1 expected files: [gs://cpg-test/folder1/foo.txt, gs://cpg-test/folder1/folder2/bar.txt]
CPG2 expected files: [gs://cpg-test/folder1/homer.txt, gs://cpg-test/folder1/folder2/bart.txt]

unique parents: {gs://cpg-test/folder1, gs://cpg-test/folder1/folder2}
listdir on all parents: {
    gs://cpg-test/folder1/foo.txt, 
    gs://cpg-test/folder1/homer.txt, 
    gs://cpg-test/folder1/folder2/bar.txt
}

gs://cpg-test/folder1/folder2/bart.txt will be identified as 'not in current file list'
without having to run exists() on all files

That's not a very succinct example is it...
```

Some other tweaks I spotted that I don't feel strongly about:

- name/description/analysis dataset is handled so weirdly... These are the relevant lines in main
```
analysis_dataset = get_config()['workflow']['dataset']
name = get_config()['workflow'].get('name')
description = get_config()['workflow'].get('description')
...
name = name or description or analysis_dataset
...
description = description or name
```

This is just a weird way of processing these 3 values, each capable of overwriting any other value. I've replaced this with
```
analysis_dataset = get_config()['workflow']['dataset']
name = get_config()['workflow'].get('name', analysis_dataset)
description = get_config()['workflow'].get('description', name)
```
Which might not achieve the exact same result under all circumstances, but isn't ridiculous to look at.

### DIFF
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -1246,7 +1246,7 @@ class SampleStage(Stage[Sample], ABC):
             # collect all expected outputs across all samples
             # find all directories which will be checked
             # list outputs in advance
-            all_outputs = set()
+            all_outputs: set[Path] = set()
             for sample in dataset.get_samples():
                 all_outputs = path_walk(self.expected_outputs(sample), all_outputs)
             all_parents = list_all_parent_dirs(all_outputs)
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -28,7 +28,7 @@ from cpg_utils import Path
 from .batch import get_batch
 from .status import MetamistStatusReporter
 from .targets import Target, Dataset, Sample, Cohort
-from .utils import exists, timestamp, slugify, ExpectedResultT
+from .utils import exists, exists_on_pre_collected, timestamp, slugify, ExpectedResultT
 from .inputs import get_cohort
 
 
@@ -728,9 +728,8 @@ class Stage(Generic[TargetT], ABC):
 
             # check against the pre-scanned files if possible
             if existing_outputs:
-                first_missing_path = next(
-                    (p for p in paths if p not in existing_outputs), None
-                )
+
+                first_missing_path = exists_on_pre_collected(paths, existing_outputs)
             # otherwise individual .exists() tests
             else:
                 first_missing_path = next((p for p in paths if not exists(p)), None)
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -41,6 +41,33 @@ StageDecorator = Callable[..., 'Stage']
 TargetT = TypeVar('TargetT', bound=Target)
 
 
+def list_all_parent_dirs(all_outputs: set[Path]) -> set[Path]:
+    """
+    take all possible outputs, find unique parent directories
+
+    Args:
+        all_outputs (set[Path]): all possible outputs across multiple stages
+
+    Returns:
+        a set of all unique parent directories
+    """
+    return {output.parent for output in all_outputs}
+
+
+def list_of_all_dir_contents(locations: set[Path]) -> set[Path]:
+    """
+    take a collection of parent directories and identify their contents
+
+    Args:
+        locations (set[Path]): all places to look
+
+    Returns:
+        set of all files across all locations
+    """
+
+    return {file for location in locations for file in location.iterdir()}
+
+
 def path_walk(expected, collected: set) -> set:
     """
     recursive walk of expected_out
@@ -64,6 +91,9 @@ def path_walk(expected, collected: set) -> set:
     >>> path_walk({'a': Path('b'),'c': {'d': 'e'}, {'f': Path('g')}}, set())
     {Path('b'), Path('g')}
     """
+    if collected is None:
+        collected = set()
+
     if expected is None:
         return collected
     if isinstance(expected, dict):
@@ -587,7 +617,9 @@ class Stage(Generic[TargetT], ABC):
 
         return outputs
 
-    def _get_action(self, target: TargetT) -> Action:
+    def _get_action(
+        self, target: TargetT, existing_files: set[Path] | None = None
+    ) -> Action:
         """
         Based on stage parameters and expected outputs existence, determines what
         to do with the target: queue, skip or reuse, etc..
@@ -607,7 +639,9 @@ class Stage(Generic[TargetT], ABC):
                 return Action.SKIP
 
         expected_out = self.expected_outputs(target)
-        reusable, first_missing_path = self._is_reusable(expected_out)
+        reusable, first_missing_path = self._is_reusable(
+            expected_out, existing_outputs=existing_files
+        )
 
         if self.skipped:
             if reusable and not first_missing_path:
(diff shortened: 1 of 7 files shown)


cassimons (commented):
This looks awesome to me Matt (thank you!). The logic all seems sound to me (and desperately needed) but I will leave it to @vivbak to sign off on.

cassimons (commented):
This looks awesome to me Matt (thank you!). The logic all seems sound to me (and desperately needed) but I will leave it to @vivbak to sign off on.

test/test_seqr_loader_dry.py
```diff
@@ -188,7 +188,9 @@ def mock_create_new_analysis(*args, **kwargs) -> int:
 
     mocker.patch('pathlib.Path.open', selective_mock_open)
     # functions like get_intervals checks file existence
-    mocker.patch('cloudpathlib.cloudpath.CloudPath.exists', mock_exists)
+    mocker.patch('cpg_workflows.workflow.list_all_parent_dirs', lambda *args: {})
+    mocker.patch('cpg_workflows.workflow.list_of_all_dir_contents', lambda *args: {})
+    mocker.patch('cpg_workflows.workflow.exists_on_pre_collected', lambda *args: False)
```
jmarshall, line 193: Lambda's return should be `None`.

MattWellie: good spot

cpg_workflows/workflow.py
```diff
@@ -64,6 +91,9 @@ def path_walk(expected, collected: set) -> set:
     >>> path_walk({'a': Path('b'),'c': {'d': 'e'}, {'f': Path('g')}}, set())
     {Path('b'), Path('g')}
     """
+    if collected is None:
```
jmarshall, line 94: If this can be none, should you adjust the type annotation on the `def path_walk` line above?

test/test_seqr_loader_dry.py
MattWellie: Somehow despite being a copy-paste from the other test case, I only got it wrong here 🙃

cpg_workflows/workflow.py
```diff
@@ -668,7 +702,17 @@ def _get_action(self, target: TargetT) -> Action:
         logging.info(f'{self.name}: {target} [QUEUE]')
         return Action.QUEUE
 
-    def _is_reusable(self, expected_out: ExpectedResultT) -> tuple[bool, Path | None]:
+    def _is_reusable(
+        self, expected_out: ExpectedResultT, existing_outputs: set[Path] | None = None
+    ) -> tuple[bool, Path | None]:
+        """
+        Checks if the outputs of prior stages already exists, and can be reused
```
jmarshall, line 709:
```suggestion
        Checks if the outputs of prior stages already exist, and can be reused
```

Also what is the policy on filling in `Args` and `Returns` in these sorts of comments…?

cpg_workflows/workflow.py
```diff
@@ -679,9 +723,17 @@ def _is_reusable(self, expected_out: ExpectedResultT) -> tuple[bool, Path | None
...
         if get_config()['workflow'].get('check_expected_outputs'):
             paths = path_walk(expected_out, set())
-            first_missing_path = next((p for p in paths if not exists(p)), None)
             if not paths:
                 return False, None
+
+            # check against the pre-scanned files if possible
+            if existing_outputs:
+
+                first_missing_path = exists_on_pre_collected(paths, existing_outputs)
```
jmarshall, line 732: I think you need to workshop this function name a bit more. The `next(… if not exists(p), …)` previous code gives us the clue: its functionality is really closer to `not_exists_on_pre_collected()`.

cpg_workflows/workflow.py
MattWellie: Ah I definitely left that one dry. I always try and populate with type/description where possible. It's a shame that neither Pycharm or Copilot seem capable of taking the type hints from the method definition and including them in the docstring.

cpg_workflows/workflow.py
MattWellie: Fair point, I've replaced it with `missing_from_pre_collected` as both None and "a file" semantically fit with that name

MattWellie:
I fired the pipeline up (on a single stage workflow where the output was already completed): https://batch.hail.populationgenomics.org.au/batches/424560/jobs/1

Pros: The code looks like it works alright, is producing the same result as the version on main
Negs: I doubt this will achieve the time saving I'm looking for. Looking at that long running batch I referenced in the top-level description, the timestamps in the logging suggest that the 20-minute chunks of time passing occur when submitting the jobs, rather than assessing them (in this instance it looks like ~17500 Genotyping jobs took 20 minutes to submit)

```
2023-06-07 01:09:18 INFO (root 1104): ************************************************************
2023-06-07 01:09:18 INFO (root 1105): Stage #3: JointGenotyping <- [Genotype]
2023-06-07 01:09:20 INFO (root 668): JointGenotyping: Cohort("seqr", 24 datasets) [QUEUE]
2023-06-07 01:27:02 INFO (root 69): Submitting joint-calling jobs
2023-06-07 01:28:30 INFO (root 170): Queueing gather VCFs job
2023-06-07 01:28:30 INFO (root 182): Queueing gather site-only VCFs job
2023-06-07 01:41:36 INFO (root 1113): 
2023-06-07 01:41:36 INFO (root 1104): ************************************************************
2023-06-07 01:41:36 INFO (root 1105): Stage #4: Vqsr <- [JointGenotyping]
2023-06-07 01:41:37 INFO (root 668): Vqsr: Cohort("seqr", 24 datasets) [QUEUE]
2023-06-07 01:42:28 INFO (root 1113): 
```

MattWellie:
I've updated this PR with change from #377  - that change is specifically aimed at removing 'placeholder' jobs from the Hail Batch workflow dependency graph. In recent workflows like the one linked in this PR's description, 90% of the jobs would not be scheduled at all, which should make pipeline startup much quicker. 

When a pipeline runs for the first time on new samples, and 000s of jobs are actually required, this won't improve the time taken to get started.

vivbak (approved):

PR #376, opened 2023-06-07
i/o suggestions

MattWellie:
Jjust making a suggestion PR. 

It looks like the copy-cloud-to-local functionality you're looking for here should be handled by the `batch.read_input()` methods, rather than by recreating the copy into batch TMP manually.

The exact syntax to use for declaring/extracting the outputs as a resource group I'm not 100% on.

### DIFF
--- a/cpg_workflows/jobs/gatk.py
+++ b/cpg_workflows/jobs/gatk.py
@@ -74,7 +74,20 @@ def collect_read_counts(b, sample, intervals, job_attrs, output_path):
     return [j]
 
 
-def determine_ploidy(b, cohort_name, ploidy_priors, inputs, job_attrs, output_dir):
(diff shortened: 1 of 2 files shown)


jmarshall:
Yes, by a remarkable coincidence I've just been looking at using `read_input` for the `ploidy_priors` file too.

For the HDF5 files, that copying is just a temporary hack: I think the better approach will be to revert to using text `.tsv.gz` files instead HD5, which `DetermineBlahPloidy` says it can read directly from GCS. I suspect they will be smaller too.

Thanks for the suggestions about handling the outputs. I reckon the gcnvCaller scatter and Postprocess gather jobs can go in this single batch / stage so these temporary files will want to be done entirely in hail batch.

jmarshall:
Cheers — suggestions now largely incorporated into the base PR.

PR #366, opened 2023-05-31
Sequencing groups update

vivbak:
Changes described in full here https://docs.google.com/document/d/1FxpsF5LfZZ5ZXWZL-JtELxPoToXMzRBrPjxyAYOba_k/edit?usp=sharing

- [x] Sample -> SequenceGroup
- [x] Sequence -> Assay 
- [x] Variable Clean up 

RestAPI Calls -> GraphQL 

- [x] Populate SequenceGroups
- [x] Populate Assays 
- [x] Populate Analysis 
- [x] Populate Participants 
- [x] General refactor inline with above. 

Testing 
- [x] cohort population 

A couple of things have deliberately been excluded from this PR:
- Filenames that are created have not been renamed, for example `sample_qc.ht`. 
- Parameter names that feed into external tools. For example, `--sample-name-map` is a parameter for `gatk GenomicsDBImport`. In this case, the `sample_map` variable that's passed here has not been renamed. 

There are several reasons for the above omissions, that can be discussed further in a separate forum e.g. Are these filename changes going to impact any downstream analyses? When we use 'sg' centred language, and other tools use 'sample' centred language, when does it make sense to draw the line? etc.
But most importantly, this PR is already  chonky enough, and we can functionally draw the line here. 

Also, the tests are not as clean as they could be, to be refactored in a subsequent PR. 

**Test Run**
https://batch.hail.populationgenomics.org.au/batches/424753/jobs/1

**Review Roster**
Hello valued reviewers, thank you for taking the time to look at this PR. Please don't be alarmed by the size, as the vast majority of the changes are trivial. 

Please be sure to create issues for changes that fall outside of the scope of the PR rather than requesting changes here. We're under a bit of a crunch to get this out. 

You are of course welcome to review the PR in it's entirety, but I've allocated everyone specific files to look at. 
| Reviewer  | Content |
| ------------- | ------------- |
| @daniaki   | /test, mt_to_es.py, benchmark_combiner.py, benchmark_metamist_read_pedigree.py   |
| @jmarshall   | batch.py, filetypes.py, inputs.py, resources.py, utils.py, workflow.py, status.py, /large_cohort, /stages (excluding gatk_sv)  | 
| @MattWellie  | /jobs, main.py, /stages/gatk* |
| @illusional | metamist.py, targets.py, populate_metamist.py, setup.py |
| @silkm  | README.MD, /configs (i.e. all the tomls)|

Thank you all in advance for your help!

PR #381, opened 2023-06-15
Increase mem for merge_gvcfs_job and make storage_gb configurable for _haplotype_caller_one

cassimons:
Two small changes for genotype jobs:

* Increase mem for merge_gvcfs_jobs to prevent out-of-memory errors on output. As discussed [here](https://centrepopgen.slack.com/archives/C051EPKNJH1/p1686733537921009).
* Attached storage is a large cost driver for _haplotype_caller_one, the single most expensive step in the WGS pipeline. This drops the default to 40gb which should work for 99% of genomes. By making it configurable, it will be easy to set higher for any needed samples.


MattWellie (approved):

cpg_workflows/jobs/genotype.py
```diff
@@ -189,6 +189,15 @@ def _haplotype_caller_one(
     # of a Hail worker (2 cores), plus with the `highmem` instance that would
     # give enough memory: 13G. That's not going to give enough disk storage, so we
     # are explicitly requesting more storage.
+    #
+    # Based on an audit of RD crams on 19/05/23, 99% of crams are <34Gb. Will set the
+    # default to 40Gb for genomes then use a run specific confg to run the rare
+    # sample that will fail from this limit.
```
vivbak, line 195:
```suggestion
    # sequencing group that will fail from this limit.
```
In anticipation for this change coming v. soon https://github.com/populationgenomics/production-pipelines/pull/366

vivbak (approved):

jmarshall:
I just looked at this one again, and surely it's not going to be effective because `storage_gb` is still being set back to `None`/`100` immediately after the lines added?

cassimons:
Hmmm... that is not so great! The memory fix worked at least.  I will put in an new PR.

cassimons:
@vivbak from #383 and before I fix the dumb mistake pointed out by John above:
> can we add this new param to the defaults.toml file with a comment on what it's for.

`haplotypecaller_storage_gb` is only being applied for genomes not exomes. If I am going to move it to defaults.toml should I rename it to something like `haplotypecaller_storage_gb_genome`? Do I need to add an equivalent for exomes or is that ok to leave in the logic of the job?

vivbak:
@cassimons Perhaps better to put it somewhere nested like [workflows.genome] in the config, to make the distinction clear. Then you can access it with := and it might simplify a little. That way it's clear its genome only without the var name becoming too verbose.

MattWellie:
@cassimons I've bundled a candidate change for this into #383 
- if there's a manual override setting, use that
- if not and it's a genome, 40GB
- else, no extra storage

PR #382, opened 2023-06-16
371 only stages still checks output of other stages

### LINKED ISSUE #371: Only_Stages still checks output of other stages
I was being a little naive maybe, and expected that listing a stage in config under only_stages would run and check that stage exclusively, e.g. https://batch.hail.populationgenomics.org.au/batches/424485/jobs/1

Here the config contained:

```
only_stages = ['DatasetVCF']
```

The job failed because the output of `MtToEs` was not created. MtToEs is a separate leaf node to DatasetVCF, so it shouldn't really have been considered. I managed to get the behaviour I expected by combining:

```
first_stages = ['DatasetVCF']
last_stages = ['DatasetVCF']
only_stages = ['DatasetVCF']
```

Can we update the workflow graph assessment to treat `only_stages` as a combination of `first_stages` and `last_stages`, i.e. don't consider any inputs/outputs other than ones immediately relevant to the ones listed in `only_stages`. I've not looked at this in great detail, and there is a workaround. If there's a combination of `check_inputs`, `check_intermediates`, and `check_expected_outputs` which creates an equivalent behaviour that would be grand too.

*related side note, the MtToEs `.done` file looks like the only one in our pipeline which contains a full timestamp in the name. It can be overridden, but is there a reason we are treating it differently to all other naming conventions? There almost certainly is, this is just an observation 🙃

daniaki:
PR to fix #371

Implements a fix so that only stages that have outputs that are required by any stage in `only_stages` are checked for existence.  Let me know if this is the expected behaviour.

I also added some code that raises an error if `first_stages/last_stages` and `only_stages` are both defined in the config, since they seem incompatible. Are there any workflow cases where we would want to specify all of these simultaneously?


MattWellie (approved):
LGTM. This behaviour comes after adding implicit stages (which we may not want), but will then skip them so this looks like a least intrusive solution 👌

daniaki:
Waiting for #366 before merging into main

PR #383, opened 2023-06-19
Split picard markdup in 2

MattWellie:
Breaks Picard Markdup and Samtools view into distinct steps, within the same batch job.

Despite setting `set -x` and `pipefail`, the error messages being created at this stage are not useful - Picard's random termination is read as a samtools reading error, or a missing EOF. Decoupling these two steps is marginally 'less efficient', but we're getting desperate. A useful error message here may be helpful in diagnosing.

Picard also complains about having insufficient memory, so it resorts to an unbuffered mode prior to many of the crashes. A higher memory limit could mitigate this, but we don't want to just keep doubling it. I've added a config option `workflow. exceptional_markdup_mem_gb ` to boost memory provision, hopefully able to clean up these instances where there are only a few samples failing but they're failing consistently. The default remains the same as it is now.

Please review critically, I've no idea if this is a good idea or not. 
ref: https://batch.hail.populationgenomics.org.au/batches/424719?q=failed

### DIFF
--- a/cpg_workflows/jobs/picard.py
+++ b/cpg_workflows/jobs/picard.py
@@ -134,7 +134,7 @@ def markdup(
     resource.attach_disk_storage_gb = 250
 
     # check for a memory override for impossible samples
-    if (memory_override := get_config()['workflow'].get('exceptional_markdup_mem_gb')) is not None:
+    if (memory_override := get_config()['resource_overrides'].get('picard')) is not None:
         assert isinstance(memory_override, int)
         # Hail will select the right number of CPUs based on RAM request
         j.memory(f'{memory_override}G')
--- a/cpg_workflows/jobs/picard.py
+++ b/cpg_workflows/jobs/picard.py
@@ -134,6 +134,8 @@ def markdup(
     resource.attach_disk_storage_gb = 250
 
     # check for a memory override for impossible sequencing groups
+    # if RAM is overridden, leave CPU setting up to Hail based on
+    # RAM/CPU core ratios
     if (memory_override := get_config()['resource_overrides'].get('picard')) is not None:
         assert isinstance(memory_override, int)
         # Hail will select the right number of CPUs based on RAM request
--- a/cpg_workflows/jobs/picard.py
+++ b/cpg_workflows/jobs/picard.py
@@ -134,7 +134,7 @@ def markdup(
     resource.attach_disk_storage_gb = 250
 
     # check for a memory override for impossible samples
-    if (memory_override := get_config()['workflow'].get('exceptional_picard')) is not None:
+    if (memory_override := get_config()['workflow'].get('exceptional_markdup_mem_gb')) is not None:
         assert isinstance(memory_override, int)
         # Hail will select the right number of CPUs based on RAM request
         j.memory(f'{memory_override}G')
--- a/cpg_workflows/jobs/picard.py
+++ b/cpg_workflows/jobs/picard.py
@@ -132,17 +132,14 @@ def markdup(
     resource = HIGHMEM.request_resources(ncpu=4)
     # enough for input BAM and output CRAM
     resource.attach_disk_storage_gb = 250
+    resource.set_to_job(j)
 
     # check for a memory override for impossible sequencing groups
-    # if RAM is overridden, leave CPU setting up to Hail based on
-    # RAM/CPU core ratios
+    # if RAM is overridden, update the memory resource setting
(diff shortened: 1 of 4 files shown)


cpg_workflows/jobs/picard.py
```diff
@@ -132,7 +132,16 @@ def markdup(
     resource = HIGHMEM.request_resources(ncpu=4)
     # enough for input BAM and output CRAM
     resource.attach_disk_storage_gb = 250
-    resource.set_to_job(j)
+
+    # check for a memory override for impossible samples
+    if (memory_override := get_config()['workflow'].get('exceptional_picard')) is not None:
```
cassimons, line 137: Should this config name be a bit more explicit?
```suggestion
    if (memory_override := get_config()['workflow'].get('exceptional_markdup_mem_gb')) is not None:
```

cassimons (commented):
I think the general idea here is pragmatic. @vivbak do we have/should we have a std way of naming config options that we only expect to use for outlier samples? This is similar to my [PR](https://github.com/populationgenomics/production-pipelines/pull/381) from last week.

cpg_workflows/jobs/picard.py
```diff
@@ -132,7 +132,16 @@ def markdup(
     resource = HIGHMEM.request_resources(ncpu=4)
     # enough for input BAM and output CRAM
     resource.attach_disk_storage_gb = 250
-    resource.set_to_job(j)
+
+    # check for a memory override for impossible samples
```
vivbak, line 136:
```suggestion
    # check for a memory override for impossible sequencing groups
```
lol

vivbak:
Nothing in practice at present, but a new category might be better instead of lumping everything directly under 'workflow'. 
I should have mentioned this in the other PR too, but can we add this new param to the defaults.toml file with a comment on what it's for.

MattWellie:
> can we add this new param to the defaults.toml file

I've added a section to the `seqr_loader.toml` default for resource_overrides. The changes from this and #381 are both moved to there

cpg_workflows/jobs/picard.py
```diff
@@ -132,7 +132,16 @@ def markdup(
...
-    resource.set_to_job(j)
+
+    # check for a memory override for impossible sequencing groups
+    if (memory_override := get_config()['resource_overrides'].get('picard')) is not None:
+        assert isinstance(memory_override, int)
+        # Hail will select the right number of CPUs based on RAM request
+        j.memory(f'{memory_override}G')
+        j.storage(f'{resource.get_storage_gb()}G')
+    else:
+        resource.set_to_job(j)
```
jmarshall, line 143: I'd suggest doing `set_to_job` unconditionally, thus:

```python
resource.set_to_job(j)
if (memory_override := …)
    assert isinstance(memory_override, int)
    # Hail will select the right number of CPUs based on RAM request
    j.memory(f'{memory_override}G')
```

This ensures that in the (unlikely) event more unrelated settings are added to `resource` in future, they don't get ignored when the override is set.

(Unless it's important that `j.cpu(…)` is not called so that `j._cpu` stays None in order for that “Hail will select” to happen, in which case it might be good to mention that in the comment.)

cpg_workflows/jobs/picard.py
```diff
@@ -144,12 +153,16 @@ def markdup(
     assert isinstance(j.output_cram, hb.ResourceGroup)
     cmd = f"""
     picard MarkDuplicates -Xms{resource.get_java_mem_mb()}M \\
-    I={sorted_bam} O=/dev/stdout M={j.markdup_metrics} \\
+    I={sorted_bam} O={j.output_bam} M={j.markdup_metrics} \\
```
jmarshall, line 156: Possibly call this `j.temp_bam` or so, so that which is in and which is out in the `samtools view` arguments below is clearer. But no biggie.

jmarshall (approved):
Couple of minor comments that you might or might not act on, but otherwise LGTM.

Hopefully running the two commands sequentially sheds some light on the matter, but I'm not holding my breath… 🤞

cpg_workflows/jobs/picard.py
MattWellie: I've added a clarifying docstring here - AFAIK by setting either CPU or memory Hail will find a machine configuration to satisfy that based on the fixed CPU-RAM ratio, so no need to set both

MattWellie: I had a second look and yeah, makes sense to set everything, then override if needed. I'm not sure which other options are going to appear in `resource`, but just in case

PR #385, opened 2023-06-21
Extract validation steps into separate workflow

MattWellie:
The Validation Sample/SequencingGroup stages are taking almost an hour to process, and that's with the intention of only running on ~5 individuals.
Not good enough, snap this whole series of stages off until we've addressed our pipeline startup timing issues.


cassimons (approved):
Thanks Matt! Looks good to me.

Issue #390, opened 2023-06-27
Analysis entries block pipeline runs 


vivbak:
In the case where an analyses doesn't have sequencing groups, we should flag a warning, and send an appropriate notification, but it shouldn't be blocking, as it is here 
https://batch.hail.populationgenomics.org.au/batches/424948/jobs/1 

We should

- [ ] Switch the assertion to a warning 
- [ ] Log the error 
- [ ] Send a notification

vivbak:
To be lumped into an analysis overhaul issue.

PR #392, opened 2023-06-27
Fix premature assertion

vivbak:
Addresses part of https://github.com/populationgenomics/production-pipelines/issues/390


jmarshall (approved):
It will be interesting to see how many of these are triggered in a run, and hence whether we may want to put e.g. `{analysis.id}` into the message instead of printing out the whole `{analysis}`.

But in general, LGTM, continue processing the rest of the analyses, just like the `continue` a few lines earlier.

PR #396, opened 2023-06-30
Explicitly set index path so bazam can find it.

cassimons:
During cram re-alignemnt, if an index is missing the alignment job regenerates the index. However, bazam only looks for `foo.crai` not `foo.cram.crai`. This change explitly sets the index file name to `foo.crai`

Slack Thread: https://centrepopgen.slack.com/archives/C051EPKNJH1/p1688010579371479


jmarshall (approved):
LGTM

PR #398, opened 2023-07-03
Update to valid file extensions

MattWellie:
GATK-SV workflow correction

Somewhere in the transition from Dataset Stage to Cohort Stage I mangled the file names.

For a set of file extensions enumerated in the attached screenshot a strict match to the expected naming convention is required. I dropped this `***.sr.txt`, erroneously moving to `**/sr.txt` which can not be read. AFAIK this change captures all `[sd|sr|rd|baf|pe].txt` files which are used as inputs downstream. The file & index in EvidenceQC is not used downstream, shouldn't be a problem, but is being renamed out of an abundance of caution.

Thanks for @jmarshall and @illusional for the troubleshooting. 

Screenshot 2023-07-03 at 3 01 10 pm


MattWellie:
NB this change will update the expected output file names, so a couple of the Stages will be 're-run' - the outputs should still exist on the Cromwell side and be copied over to new names, so not expecting this to incur much re-compute

cassimons (approved):

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
```diff
@@ -87,11 +87,11 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
             for batch_name in batch_names
         ]
         bincov_files = [
-            batch_prefix / batch_name / 'GatherBatchEvidence' / 'RD.txt.gz'
+            batch_prefix / batch_name / 'GatherBatchEvidence' / f'{self.name}.RD.txt.gz'
```
jmarshall, line 90: Does this need to match the RD filename from _gatk_sv_multisample_1.py_?

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
```diff
@@ -46,14 +46,14 @@ def expected_outputs(self, cohort: Cohort) -> dict[str, Path]:
...
             'cnmops_large_dup_index': 'DUP.large.bed.gz.tbi',
-            'merged_SR': 'sr.txt.gz',
-            'merged_SR_index': 'sr.txt.gz.tbi',
-            'merged_PE': 'pe.txt.gz',
-            'merged_PE_index': 'pe.txt.gz.tbi',
-            'merged_BAF': 'baf.txt.gz',
-            'merged_BAF_index': 'baf.txt.gz.tbi',
-            'merged_bincov': 'RD.txt.gz',
-            'merged_bincov_index': 'RD.txt.gz.tbi',
+            'merged_SR': f'{self.name}.sr.txt.gz',
```
jmarshall, line 49: I'm a bit late suggesting this 😄 but I'd suggest a comment at the top of this bunch mentioning that they need the `.sr.txt.gz` etc extensions. Hopefully help avoid us unfixing this again in future!

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
MattWellie: Well... that's an entirely fair point. Methinks a follow-up to change this is needed 🤦‍♂️

PR #399, opened 2023-07-03
Align Job Tests 

vivbak:
Adding some helper methods and factories to make testing jobs easier


violetbrina (approved):
Looks good to me. Those factories will be super useful.

PR #400, opened 2023-07-03
Followup Path Correction

MattWellie:
Spotted by @jmarshall [here](https://github.com/populationgenomics/production-pipelines/pull/398#discussion_r1250303405)

A later step manually rebuilds a path to one of these files - that file also needs to have the correctly altered GCP path.


jmarshall (approved):
Thanks, that looks good and addresses my comments on the previous PR.

If the `GatherBatchEvidence` name wasn't determined by the WDL workflow name, I would worry that we might rename the stage in _gatk_sv_multisample_1.py_ and have a problem. However it is, so in the full context this is right and good.

I assume _gatk_sv_single_sample.py_ is a separate workflow so it doesn't matter that the RD files there are named separately — in fact, it's possibly even a good thing.

I have a feeling that these files are supposed to be lower case `*.rd.txt.gz` (based on grepping GATK source code for `rd.txt`), but if they don't get set as `localization_optional` it doesn't matter.

MattWellie:
> If the `GatherBatchEvidence` name wasn't determined by the WDL workflow name, I would worry that we might rename the stage in _gatk_sv_multisample_1.py_ and have a problem. However it is, so in the full context this is right and good.

Yeah, they're explicitly tied to the WDL name so it won't be possible to tweak it. It's still a rubbish file name, and something more descriptive would be better.... The Hash and Stage Name are both reasonably logical, but the filepath contains both, so either causes some silly redundancy 

> I assume _gatk_sv_single_sample.py_ is a separate workflow so it doesn't matter that the RD files there are named separately — in fact, it's possibly even a good thing.

Yup, that's a separate workflow, and AFAIK that output file is never read by anything downstream so shouldn't cause headaches

> I have a feeling that these files are supposed to be lower case `*.rd.txt.gz` (based on grepping GATK source code for `rd.txt`), but if they don't get set as `localization_optional` it doesn't matter.

Correct!* I had a quick check through the code you linked, and when they're tested for extensions there is a `.toLowerCase()` called on the path prior to matching so it should be safe

jmarshall:
Oh, that's good news — my eyes had skipped over the `toLowerCase()`. I see now it's present in most of these GATK `canDecode` routines.

PR #401, opened 2023-07-05
Fix index path for bazam

cassimons:
Bazam requires an index at foo.crai (not foo.cram.crai) so we must set the
index path explicitly.  The Input ResourceGroup does not provide any easy way to access the basename of the  localized cram file so we need to do sting manipulation 
in bash . This should work for both .cram and .bam files.

Thread: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1688365368231359

### DIFF
--- a/cpg_workflows/jobs/align.py
+++ b/cpg_workflows/jobs/align.py
@@ -383,12 +383,6 @@ def _align_one(
 
         if not alignment_input.index_path:
 
-            # Explicitly set the index path so bazam can find it.
-            if isinstance(alignment_input, CramPath):
-                index_path = group[alignment_input.ext].removesuffix('.cram') + '.crai'
-            else:
-                index_path = group[alignment_input.ext].removesuffix('.bam') + '.bai'
-
             sort_index_input_cmd = dedent(
                 f"""
             mkdir -p $BATCH_TMPDIR/sorted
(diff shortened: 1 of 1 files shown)


cpg_workflows/jobs/align.py
```diff
@@ -402,7 +396,12 @@ def _align_one(
             mv $BATCH_TMPDIR/sorted.{alignment_input.ext} {group[alignment_input.ext]}
             rm -rf $BATCH_TMPDIR/sort_tmp
 
-            samtools index -@{nthreads - 1} {group[alignment_input.ext]} {index_path}
+            # bazam requires an index at foo.crai (not foo.cram.crai) so we must set the
+            # path explicitly. We can not access the localized cram files basename via the
+            # Input ResourceGroup so using bash magic to strip the trailing "m" and add an
```
jmarshall, line 401:
```suggestion
            # path explicitly. We can not access the localized cram file's basename via the
            # Input ResourceGroup so using shell magic to strip the trailing "m" and add an
```

jmarshall (approved):
LGTM, and unlike the previous PR to address this, this time your reviewer has also tested it and checked that it works rather than just eyeballing it :smile:

Two very pedantic minor comments on the comment (the `${foo#suffix}` syntax is not bash-specific, so in the unlikely event we ever want to use a different shell…) which you might or might not bother with.

PR #402, opened 2023-07-05
Correction to Batch names

MattWellie:
See https://github.com/broadinstitute/gatk-sv/issues/564

We have an issue whereby the merging of data from two separate runs is causing downstream duplication. With Emma's input I've tracked this back to the ClusterBatch step - 

- Each separate group of SGs being processed independently needs a distinct ID
- This ID is incorporated into the Variants identified as [part of a key](https://github.com/broadinstitute/gatk-sv/blob/main/wdl/PESRClustering.wdl#L77)
- When I set up these jobs I was erroneously passing down the `cohort.name`, i.e. `seqr`, as the batch name for all batches
- `seqr` is then included into both sets of IDs, which leads to duplicated IDs when multiple VCFs are merged together (unique variants are determined by contig, start, end, DUP/DEL/etc., not by ID)

This change corrects this behaviour - for all stages processing sample batches in parallel, the hash of all relevant SGs is taken as the 'batch id' to be incorporated into the output files.

### DIFF
--- a/cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
+++ b/cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
@@ -23,6 +23,7 @@ from cpg_workflows.stages.gatk_sv.gatk_sv_common import (
     make_combined_ped,
     SV_CALLERS,
 )
+from cpg_workflows.workflow import get_workflow
 
 
 @stage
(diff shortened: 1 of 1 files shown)


MattWellie:
I'll need to scrub some previously generated results when re-running this

cassimons (approved):
Good find!

PR #405, opened 2023-07-10
GATK-SV sample finder script

MattWellie:
Leaving this here as a draft - a self-documenting script to trawl through metamist and build a config file suitable for starting a new GATK-SV run.


MattWellie:
Closing as superseded

PR #407, opened 2023-07-12
Clean up local storage in picard markdups jobs.

cassimons:
Followiing PR383, markdups jobs currnently take a sorted cram, run markdups to write an interum bam then convert to a final indexed cram. THis means we need enough local storage for three copies of the cram. This PR deletes the input cram after the markdups steps to reduce this to a max of two copies at any time.


MattWellie (approved):
LGTM

PR #410, opened 2023-07-12
Add storage override for markdups job

cassimons:
Default local storage is insufficnet for very large crams mkduplicates job. This PR adds a configerable overide to be used in when needed for very large crams.

eg https://batch.hail.populationgenomics.org.au/batches/425798/jobs/2


vivbak (approved):
Could we please also leave a comment about this new param in the defaults.toml file? Worth noting re-organising our configs is on the cards for us, but for now helpful to keep the record of all params there.

PR #413, opened 2023-07-13
Rare disease RNAseq workflow - trim stage

mgeaghan:
Created a stage for trimming RNAseq FASTQs using fastp.
Also contains an older class for constructing a cutadapt command - if we want to use this in the future instead.

One job is created per FASTQ pair, e.g. one per lane.


MattWellie:
AFAIK to get this working as a workflow stage (and to get all the SequencingGroup objects populated from the cohort metadata) you'll need 2 more things:
- You'll need to import your Stage into `main`, and add it to a named workflow (maybe adding a new workflow?) [here](https://github.com/populationgenomics/production-pipelines/blob/rare-disease-rnaseq/main.py#L35)
- Add a new default TOML config file [here](https://github.com/populationgenomics/production-pipelines/tree/main/configs/defaults), where the name of the file matches the workflow name

You'd then be able to run this stage as part of a workflow like in [the examples](https://github.com/populationgenomics/production-pipelines/tree/rare-disease-rnaseq#example-usage), even if it's just a single stage workflow

MattWellie (approved):
Looks good, there might be an easier way to retrieve fq paths if they're currently stored in metamist

configs/defaults/rare_disease_rnaseq.toml
```diff
@@ -0,0 +1,5 @@
+[workflow]
+sequencing_type = "rna"
+
+[trim]
+adapter_type = "ILLUMINA_TRUSEQ"
```
cassimons, line 5: All default configs also need to be listed in cpg_workflows/defaults.toml with an explanation of what is being configured

cpg_workflows/jobs/trim.py
```diff
@@ -0,0 +1,239 @@
...
+        two_colour: bool = True,
+        polyA: bool = False,
+        quality_trim: int | None = None,  # TODO: determine default
+    ):
+        try:
+            adapters: AdapterPair = AdapterPairs[adapter_type].value
+        except AttributeError:
+            raise ValueError(f'Invalid adapter type: {adapter_type}')
+        self.command = [
+            'cutadapt',
```
cassimons, line 90: It is a matter of style, but using f-strings to build cmd blicks is generally preferred.

cpg_workflows/stages/trim.py
```diff
@@ -0,0 +1,142 @@
...
+class Trim(SequencingGroupStage):
+    """
+    Trim raw FASTQ reads using cutadapt
+    """
+
+    def expected_outputs(self, sequencing_group: SequencingGroup) -> dict[str, Path]:
+        """
+        Expect a series of trimmed FASTQ files
+        """
+        return get_output_dict(sequencing_group)
```
cassimons, line 108: It would improve readability if the logic from get_output_dict as put here directly (when reading pipelines this is a place you always come to look at).

PR #416, opened 2023-07-18
Silence some mypy diagnostics

jmarshall:
This gets rid of some existing trivia when you run

```
mypy .
```

looking for new diagnostics coming from your own changes.

The remaining messages are of two kinds:

```
note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs
error: Cannot instantiate abstract class "CloudPath" with abstract attributes "drive", "is_dir", "is_file", "mkdir" and "touch"  [abstract]
```

It might be worth running with `--check-untyped-defs` to see if there is further low-hanging trivia that can be cleared up.

The `CloudPath` problem is one that hopefully ought to be / could be fixed by cloudpathlib changes (see drivendataorg/cloudpathlib#328).

### DIFF
--- a/cpg_workflows/dataproc_scripts/mt_to_es.py
+++ b/cpg_workflows/dataproc_scripts/mt_to_es.py
@@ -119,7 +119,7 @@ def main(
     mt_path: str,
     es_index: str,
     done_flag_path: str,
-    password: str = None,
+    password: str | None = None,
 ):
     """
     Entry point.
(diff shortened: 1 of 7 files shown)


jmarshall:
Hmmm… Test failures appear to be because we've killed metamist… 🙃

vivbak (approved):
sorry for delay, got distracted by metamist imploding

PR #423, opened 2023-07-28
Fixes to get things running with Analysis runner.

cassimons:
This adds a few fixes to get the stage running successfully (eg [job](https://batch.hail.populationgenomics.org.au/batches/426223/jobs/1)). Costs about 8c per fastq to run ✅

- Fixes weird error when generating expected outputs that was causing the input fastq to be localised to the driver job (🤷‍♂️).
- Changes outputs to from ResourceFiles to ResourceGroups with a defined .gz suffix. Without this fastp was outputting unzipped fastqs.
-  Uses prod Fastp image that has now been deployed
- Changes output to tmp_prefix (deleted after 8 days) 
- Applied black formatting

### DIFF
--- a/cpg_workflows/jobs/trim.py
+++ b/cpg_workflows/jobs/trim.py
@@ -222,10 +222,7 @@ def trim(
     trim_j_name = base_job_name
     trim_j_attrs = (job_attrs or {}) | dict(label=base_job_name, tool=trim_tool)
     trim_j = b.new_job(trim_j_name, trim_j_attrs)
-    # trim_j.image(image_path('fastp'))  # PRODUCTION
-    trim_j.image(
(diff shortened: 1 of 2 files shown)


jmarshall:
> Changes outputs to from ResourceFiles to ResourceGroups with a defined .gz suffix. Without this fastp was outputting unzipped fastqs.

If they're naturally single files, a maybe simpler alternative is to use `trim_j.output_r1.add_extension('.fastq.gz')` (and similarly for the others) to give the filenames the extension `fastp` wants. (Lots of examples of using this in PR #365.)

cassimons:
> trim_j.output_r1.add_extension('.fastq.gz') 

How did I never know that existed!?

jmarshall:
The advantage of having new employees who've never used hail before is that they read the manual cover to cover :smile:

PR #424, opened 2023-07-30
Exercise tests with hail 0.2.120

jmarshall:


jmarshall:
Tests fail with:

```
FAILED test/test_large_cohort.py::TestAllLargeCohortMethods::test_with_sample_data - AttributeError: 'StringExpression' object has no attribute 'is_non_ref'
```

_test_large_cohort.py_ runs `combiner.run()` (essentially `hl.vds.new_combiner`/`run`) to create a vds from some GVCF files, then runs `dense_subset.run()` (which starts with `hl.vds.split_multi`) on it. The AttributeError occurs within `split_multi()`.

This is because (with 0.2.120) the data structure returned by `hl.vds.new_combiner`/`run` lists LPGT as `str` instead of `call`, which later causes the exception when _sparse_split_multi.py_'s `transform_entries()` looks at the LPGT entry and tries to use a `CallExpression` method on it.

```
----------------------------------------
Global fields:
    None
----------------------------------------
Column fields:
    's': str
----------------------------------------
Row fields:
    'locus': locus<GRCh38>
    'alleles': array<str>
    'rsid': str
----------------------------------------
Entry fields:
    'LA': array<int32>
    'LGT': call
    'LAD': array<int32>
    'LPGT': str           # BROKEN, should be "call"
    'LPL': array<int32>
    'RGQ': int32
    'gvcf_info': struct {
        […etc…]
```

With hail 0.2.119 this entry is `'LPGT': call` as expected, and the test runs successfully. Thus this failure is new in 0.2.120 and I suspect caused by PR hail-is/hail#13206.

PR #425, opened 2023-08-01
Avoid hail 0.2.120, which breaks test/test_large_cohort.py

jmarshall:
Downgrade to 0.2.119 to avoid apparent Hail bug causing #424.

Using `!=` means this will choose an upcoming 0.2.121 when that is released, at which point we can remove this (if the problem is fixed) or widen it (if the problem persists). The available version comparison operators are [documented here](https://peps.python.org/pep-0440/#version-specifiers).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         'cpg-utils',
         'cyvcf2==0.30.18',
         'analysis-runner>=2.41.2',
-        'hail',
+        'hail!=0.2.120',  # Temporarily work around hail-is/hail#13337
         'networkx',
         'metamist>=6.0.4',
         'pandas',


jmarshall:
This has selected hail-0.2.119 as desired:

```
Collecting hail!=0.2.120 (from cpg-workflows==1.14.3)
  Obtaining dependency information for hail!=0.2.120 from https://files.pythonhosted.org/packages/05/b9/2ccf19f4895b62da5785c945a274813cad9818b42f9e8b8301977feb38ba/hail-0.2.119-py3-none-any.whl.metadata
  Downloading hail-0.2.119-py3-none-any.whl.metadata (6.0 kB)
```

and the unit tests now succeed.

vivbak (approved):
Truly love that you can set !=, I have learnt something new today :) 🙌

jmarshall:
You would be surprised how long it took to track down where the available operators are documented :-)

PR #426, opened 2023-08-01
Michael h changes

michael-harper:
@daniaki


test/test_large_cohort.py
```diff
@@ -20,12 +23,20 @@
...
-from cpg_workflows.targets import Cohort
+from cpg_workflows.targets import Cohort, Dataset
 
 from . import set_config
 from .factories.config import HailConfig, PipelineConfig, StorageConfig, WorkflowConfig
+from .factories.sequencing_group import create_sequencing_group
 from .factories.types import SequencingType
 
+# test imports
+import pytest
```
daniaki, line 34: This is already imported above

test/test_large_cohort.py
```diff
@@ -20,12 +23,20 @@
...
+from hail.utils.java import FatalError
```
daniaki, line 35: This is also already imported above: `from hail.utils import FatalError`

cpg_workflows/large_cohort/ancestry_pca.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 from cpg_utils import Path, to_path
 from cpg_utils.config import get_config
-from gnomad.sample_qc.ancestry import run_pca_with_relateds, assign_population_pcs
+from gnomad_methods.gnomad.sample_qc.ancestry import run_pca_with_relateds, assign_population_pcs
```
daniaki, line 8: We should revert this change, since we found out the real issue was that we didn't run pip install on `gnomad_methds`

test/jobs/test_samtools.py
```diff
@@ -0,0 +1,111 @@
...
+                    'ref_fasta': 'hg38_reference.fa',
+                    'dragmap_prefix': 'gs://a-cpg-bucket/dragen_reference/',
+                }
+            },
+        },
+    )
+
+
+class TestSamtoolsRun:
+    def test_create_cram_input_method_creates_inputs(self, tmp_path: Path):
```
daniaki, line 40: Although a good idea  to test the `create_cram_input` factory, these tests should only be testing the samtools job module. Factory tests will live in their own test files and will be done at a later date.

test/jobs/test_samtools.py
```diff
@@ -0,0 +1,111 @@
...
+        set_config(config, tmp_path / 'config.toml')
+
+        cram_pth = create_cram_input(
+            location=tmp_path, prefix='test', index=True, reference_assembly='GRCh38.fa'
+        )
+        # ---- Check CRAM and CRAI files exist
+        assert os.path.exists(tmp_path / 'test.cram')
+        assert os.path.exists(tmp_path / 'test.cram.crai')
+
+    def test_run_samtools(self, tmp_path: Path):
```
daniaki, line 52: This is a good first attempt which is asserting a lot of the things we'd want to test. However, ideally I'd like to see these tests split into smaller tests with much more descriptive names testing one aspect at a time. It should be clearly obvious what a test is doing just by reading the function name as opposed to a general names like `test_run_samtools` and `test_sample_command` in which the intent is not clear. Some examples of how I would structure tests:

test_creates_one_job
test_can_reuse_existing_output_path
test_can_overwrite_existing_output_path
test_sets_job_attrs_or_sets_default_attrs_if_not_supplied
test_uses_samtools_image_specified_in_config
test_uses_reference_in_workflow_config_section_if_set
test_uses_broad_reference_as_default_if_reference_not_set_in_workflow_config_section
test_uses_fail_safe_copy_on_cram_path_and_index_in_bash_command
test_samtools_writes_to_resource_file_named_output_stats
test_batch_writes_samtools_stats_file_to_output_path

You can test for things in the job's bash command using a regex assertion, and the `get_command_str` helper.

test/jobs/test_samtools.py
```diff
@@ -0,0 +1,111 @@
...
+            out_samtools_stats_path=tmp_path,
+            job_attrs=None,
+            overwrite=True,
+        )
+        samtools_jobs = batch.select_jobs('samtools stats')
+        # ---- Assertions
+        cmd = get_command_str(samtools_jobs[0])
+        print()
+
+    def test_check_batch_command_executed(self, tmp_path: Path):
```
daniaki, line 107: Note that these jobs are not actually executed so you can't specifically test what happens if they run.

test/test_large_cohort.py
```diff
@@ -20,12 +23,20 @@
...
+import hail as hl
+from cpg_workflows.inputs import get_cohort
+from cpg_utils.config import get_config
+from unittest.mock import _ANY
```
daniaki, line 39: Don't forget to delete unused imports

test/test_large_cohort.py
```diff
@@ -122,28 +133,271 @@ def create_config(
...
+            'this_shouldnt_work',  # Junk string
+            'CHR1:POS1-CHR1:100M',
+            'CHR2:POS2-CHR1:POS1',  # The start locus ("CHR2:POS2") must precede the end locus ("CHR1:POS1").
+            'CHR1:POS2-CHR1:POS1',  # The start position ("POS2") must be less than the end position ("POS1") within the same contig ("CHR1")
+            'CHR1:POS1-CHR1:START',  # The position "START" stands for 1, so it should not be used as an end position in the interval
+        ]
+
+        context_started = False
+
+        with open(tmp_path / 'intervals.txt'):  # test if can read intervals from txt
```
daniaki, line 267: This should be in a separate test named `test_can_load_intervals_from_a_text_file`

test/test_large_cohort.py
```diff
@@ -122,28 +133,271 @@ def create_config(
...
+        combiner.run(out_vds_path=vds_path, tmp_prefix=res_pref / 'tmp')
+        print()
+
+    def test_fails_if_given_two_sequencing_groups_with_the_same_gvcf_path(
+        self, mocker: MockFixture, tmp_path: Path
+    ):
+        pass
+
+    @pytest.mark.parametrize('seq_type', ['exome', 'genome'])
+    def test_calls_hail_combiner_with_correct_parameters(
```
daniaki, line 331: I would rename this to `test_can_use_default_intervals_if_intervals_are_not_set_in_config`, and delete the tests `test_uses_default_exome_intervals_if_intervals_are_not_specified` and `test_uses_default_genome_intervals_if_intervals_are_not_specified` since they are testing the same thing

daniaki (changes requested):
Good work so far! I've left some comments on some minor points to address along with recommendations on how I would go about splitting up your current implementations into more fine-grained tests to follow best practices.

I would also like you to install the `isort` VSCode python extension and sort your import statements using this tool.

daniaki:
John recently pushed an update fixing the breaking hail change, so update your branch with the 'update branch' button when you have time.

cpg_workflows/large_cohort/ancestry_pca.py
michael-harper: Hmmm after you posted those instructions I did a fresh fork so I'm not sure why they're still there. Will investigate

PR #463, opened 2023-09-13
gCNV merge + annotate

MattWellie:
Additional post-calling gCNV stages:

- FixDodgyVcfs runs a per-sample alteration on the Intervals VCFs, swapping the Genotype (GT) field from Integer (bad, wrong), to String (great, cool) - [link to GATK issue](https://github.com/broadinstitute/gatk/pull/8621). I'm not attached to the name, but the VCFs are dodgy, and hopefully this stage won't be required in future
- Throws a quick BCFtools merge on the already co-called single-sample VCFs, and follows it up with a python script to insert SVTYPE (always CNV) and SVLEN (derived from POS and END) so that the joint-called VCF can be annotated by the same process as the GATK-SV WDL
- Abstracts out the bulk of SV VCF annotation from the GATK-SV multisample 2 file: both gCNV and GATK-SV stages call the same method; we can't have a single named stage called in 2 contexts from 2 separate workflows, so this reduces code lines and complexity whilst allowing both workflows to access the same task
- Includes annotation of the resulting VCF with StrVCTVRE

Bonus content: 
- Linted until the workflow went green

Stopping point: 
This workflow isn't finished up to the Seqr ES Index generation. That will follow a similar flow to the SV version, and some code may be share-able. The Luigi pipeline we're emulating is [here](https://github.com/populationgenomics/seqr-loading-pipelines/blob/master/luigi_pipeline/seqr_gcnv_loading.py) (and other linked files), but as-of now there are a number of annotations (SC/SN/SF?) which are expected in that pipeline but absent in our files.

Run using this content: https://batch.hail.populationgenomics.org.au/batches/430825
- FixDodgyVCFs was run previously, I can fish that out of the batch logs if required
- Merge and modify single-sample VCFs
- Annotate with GATK-SV's AnnotateVCF WDL
- Annotated with StrVCTVRE


cpg_workflows/jobs/gcnv.py
```diff
@@ -301,3 +301,63 @@ def postprocess_calls(
...
+        job_attrs | {'tool': 'bcftools'}
+    )
+    j.image(image_path('bcftools'))
+
+    j.declare_resource_group(
+        output={'vcf.bgz': '{root}.vcf.bgz', 'vcf.bgz.tbi': '{root}.vcf.bgz.tbi'}
+    )
+
+    # make a fuse connection to the data source
+    # to require only one fused connection we need files in one bucket
```
cassimons, line 342: Will this be true? Do we write SG calls into the seqr bucket currently?

jmarshall:
Re the most recent commit, I actually like the technique of using `add_extension` rather than artificially bundling files into a group via  `declare_resource_group`…

MattWellie:
> Re the most recent commit, I actually like the technique of using `add_extension` rather than artificially bundling files into a group via `declare_resource_group`…

Yees, but you're missing the indexes - the index files are created as an implicit path (`other file + tbi`), so you can't write them to a nominated output path, which means you can't read them into permanent storage

jmarshall:
FYI I am unable to review this PR. It makes unrelated changes to functions that it doesn't actually affect, for example formatting changes (which are presumably fallout from black), so there is too much to meaningfully review.

MattWellie:
cancelled in favour of fragmented PRs

PR #481, opened 2023-10-05
VEP 110

MattWellie:
Ok, there's a lot here

- https://github.com/populationgenomics/analysis-runner/tree/main/scripts/vep_backend was the original location of the 'Run VEP on my VCF' scripts folder. I've transplanted that over to here which feels much more appropriate, given that this is where all the corresponding code lives
- Alters the github workflow to only run the linting workflow when the commit being pushed is related to an open PR, or a push to main. Hoping that this will cut down on the number/runtime of github actions (1minute/run, the testing workflow is already restricted in the same way)
- Adjustments to the VEP script and workflow to allow for generating annotated VCF output (this workflow is partially implemented and broken, but we probably never found out because we don't try to generate VCF output). Mainly reading VCF fragments as local files to `bcftools concat` instead of wrongly using ResourceGroups
- implements a `use_110` variable to pass around to switch from the standard VEP 105 process/image/post-annotation processing to the VEP 110 version. Default remains 105.
- `references.vep_110_mount` is used as the location of the VEP 110 cache files, though this is currently only available in test. (see https://github.com/populationgenomics/references/pull/35)
- Adds an escape clause if we are only running VEP as one fragment (scatter_count = 1). I picture this low throughput being the only time we would use a VEP -> VCF output annotation in production-pipelines
- DRAMATICALLY cuts down the VM resourcing for VEP. 50GB RAM, 16 CPU -> 8GB RAM & 2 cores. I've tested this on a few jobs, and even with multiple plugins and a 2+ hour run time we don't break 4GB memory consumption. The LOFTEE plugin restricts us from running a forked annotation process, so the current spec wastes at least 15 cores... Wild. 
- Implements both the VEP 110 and VEP 105 schema separately in the VEP JSON -> HailTable script, with a corresponding selection script.
- The vep backend script also includes a `--to-mt` option, which completes the VEP annotation, and then goes further, building the same MT as the standard pipeline annotation process (including ClinVar and gnomAD annotations)

Chunky changes in `vep_one`:
- A few switched behaviours depending on if we use VEP 105 or 110
- Loftee plugin path - only required in 105 due to the borky micromamba installation
- AlphaMissense only available in 110 due to the sexy chad installation
- If 110 & VCF output we use UTRannotator as an additional active plugin
- `--no_stats` - marginal runtime gains, and we don't use this file (not even removed from tmp)
- We do not authenticate GCloud in the VEP image - VEP 105 has the google SDK installed for absolutely no reason, and 110 doesn't have the library. Removing this authentication (and related imports) means that the same command is executable in both docker image

Sidebar on Cache files:
- https://github.com/populationgenomics/references/blob/main/vep/copy-references.py
- This script was used to copy over the VEP cache & LOFTEE files into `cpg-common-main`
- That process feels stupid - it downloads all the cache files (probably takes 24-48hrs, based on the rate I achieved locally and in GCP of `233kb/s`), then runs tabix on each separate bundle (when I ran this it took another 16 hours)... (this is my understanding of the process using VEP's `INSTALL.pl` script and the `--convert` argument - instead of sourcing the indexed data, the process is a long download followed by a long local re-index)
- THE INDEXED FILES ARE AVAILABLE TO DOWNLOAD DIRECTLY FROM ENSEMBL. AAAARGH.
- The download speed of data from Ensembl is pitiful, so this requires a non-preemptible machine to be running for over a day. That sounds like a wildly expensive option. 
- That approach also means that the VEP image needs the gcloud SDK to be installed to copy the files into the buckets (raising the docker image size from ~800MB in v110 to ~2.3GB in v105)
- I did this more manually through a background globus transfer & rsync upload to GCP, documented in https://github.com/populationgenomics/references/pull/35
- The files come pre-indexed and ready to use, with a slow but steady background copy and a parallel `gcloud rsync` replacing the non-preemptible `16 CPU` task that could run for days

I appreciate that this data download/upload process I followed was less easily replicated and a bit of a pain, but the scripted process we have is very expensive short term thinking, that's probably cost us an absolute pile over the years.

Related PRs
- Completing the VEP 110 image https://github.com/populationgenomics/images/pull/105
- Building and copying in new resource bundle(s) https://github.com/populationgenomics/references/pull/35

Related Issue
- Treats (but doesn't fully solve) #486


MattWellie:
Additional Risk - We annotate some details (transcript consequence, gene ID, etc.) via VEP, but we annotate ClinVar and dbNSFP data (in silico tools) via a Broad resource bundle. It's possible that there will be minor differences here in the annotations provided between these two sources. Should be accompany this update with an update of the corresponding broad resource bundle? If there is one available...

MattWellie:
There is one feature I would like to complete - SpliceAI running on VCF data. This is ready to test soon, but the remainder of this PR can be treated as ready to roll

MattWellie:
> There is one feature I would like to complete - SpliceAI running on VCF data

This is ready, and completes, but at a wild runtime of ~5x the time of non-spliceAI running. This is something I'd like to have in the ol' back pocket if we get traction on a non-Hail version of AIP, but it's deactivated by default

scripts/vep_backend/vep_110_conf.toml
```diff
@@ -0,0 +1,12 @@
+[workflow]
+sequencing_type = 'genome'
+use_vep_110 = true
+
+[vep]
+scatter_count = 1
+
+[references]
+vep_110_mount = 'gs://cpg-common-test/references/vep/110/mount'
```
cassimons, line 9: Should this be `VEP_110_mount` or simply `VEP_mount`? I know this one is a bit of a breaking change from the old deployment, but going forward I would be keen to end up in a situation where we have a single config which is `VEP_version` or similar where we can just choose from a list of supported versions (and cognate mounts). How do we move to that?

Regardless, can you please add any new config options into the defaults file.

scripts/vep_backend/vep_110_conf.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[images]
+cpg_workflows = 'australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:1.16.1'
```
cassimons, line 12: Is this just a testing artefact?

MattWellie:
Decoding issue with VEP 110, and why that probably doesn't matter
https://asia.ensembl.org/info/docs/tools/vep/vep_formats.html#json

The population frequency section of the VEP JSON result is a dictionary **keyed on alt allele**, e.g.

```
"frequencies": {
      "G": {
          "amr": 0.5072,
          "gnomade_sas": 0.369,
          "gnomade": 0.4541,
          "gnomade_oth": 0.4611,
          "gnomade_asj": 0.3909,
          "gnomade_nfe": 0.4944,
          "gnomade_afr": 0.103,
          "afr": 0.053,
          "gnomade_amr": 0.5641,
          "gnomade_fin": 0.474,
          "sas": 0.3906,
          "gnomade_eas": 0.4598,
          "eur": 0.4901,
          "eas": 0.4623
   }
}
```

This means that the key is not limited to `A/C/G/T`, but can be `-` for a deletion, or any arbitrary alt allele for insertions or complex variants. This doesn't really matter...... In the pipeline (where we produce JSON, decode to a HailTable, then annotate a MatrixTable) the AF annotation is applied from a separate annotation source. We can just avoid describing this whole section in the schema, and when decoding it will be skipped but not cause a failure.

scripts/vep_backend/vep_110_conf.toml
MattWellie: The main version of this path is set up in the references repo, so it's in the default config for all runs. I figured specifying the version in the key enables us to remain backwards compatible and add optional new versions.

This was the test mount specified while I was iterating on it, that can probably be deleted now..

`vep_110_mount = "gs://cpg-common-main/references/vep/110/mount"`

cassimons (approved):

PR #484, opened 2023-10-13
AIP as pipeline stages

MattWellie:
Introduces DatasetStage wrappers for the workflow components of AIP. 

1. Pull all the participants and corresponding HPO terms from Metamist
2. Query PanelApp for all panel data related to this cohort
3. Run Hail Filtering/Labelling stage using this panel data (either choose latest MT (exome or genome) based on analysis entries, or take path from config)
4. Run MOI validation stage (store analysis entry in Metamist)
5. Run results -> HTML stage, generating a summary report, and a latest-only report. Log both in Metamist

I've run this in test on `validation`, which succeeded through the first two stages ([here](https://console.cloud.google.com/storage/browser/cpg-validation-test/reanalysis/2023-10-12)), then got stuck at the lack of MT in test (where the MT SG IDs would have to correspond to the test metamist entries). Easiest solution is probably to test this in main (still on validation to avoid potential mess), but happy to take suggestions.

Twinned with https://github.com/populationgenomics/automated-interpretation-pipeline/pull/320

Also sneaks in a bonus commit so that the 'push new image for this PR' and 'push manually deployed image' both go to dev instead of the main artifact repo (@illusional  👀 )


cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+                }
+            }
+        }
+    """
+    )
+    result = query(my_query, variables={'dataset': dataset})
+    seq_type_exome = get_config()['workflow'].get('sequencing_type') == 'exome'
+    mt_path = ''
+    for analysis in result['project']['analyses']:
+        if analysis['output'].endswith('.mt') and (
```
MattWellie, line 131: An analysis entry can have None for an `analysis['output']` - allow for this instead of failing to do .endswith on a None

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+            'labelled_vcf': dataset.prefix() / DATED_FOLDER / 'hail_labelled.vcf.bgz',
+            'pedigree': dataset.prefix() / DATED_FOLDER / 'pedigree.ped',
+        }
+
+    def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput:
+
+        # either do as you're told, or find the latest
+        # this could potentially follow on from the AnnotateDataset stage
+        # if this becomes integrated in the main pipeline
+        input_mt = get_config()['workflow'].get(
```
illusional, line 159: This could technically load a different config than what was run. Not sure if that's intentional, but you could move the get_config() to the start and define as a constant

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+        expected_out = self.expected_outputs(dataset)
+        pedigree = dataset.write_ped_file(out_path=expected_out['pedigree'])
+        # peddy can't read cloud paths
+        local_ped = get_batch().read_input(str(pedigree))
+        job.command(
+            f'python3 reanalysis/hail_filter_and_label.py '
+            f'--mt {input_mt} '
+            f'--panelapp {panelapp_json} '
+            f'--pedigree {local_ped} '
+            f'--vcf_out {str(expected_out["labelled_vcf"])}'
```
illusional, line 182: Would recommend quoting these to stop a shell injection attack, or an accidental space.

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    run the labelled VCF -> results JSON stage
+    """
+
+    def expected_outputs(self, dataset: Dataset) -> dict[str, Path]:
+        return {'summary_json': dataset.prefix() / DATED_FOLDER / 'summary_output.json'}
+
+    def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput:
+        job = get_batch().new_job(f'AIP summary for {dataset.name}')
+        STANDARD.set_resources(job, ncpu=1, mem_gb=4.0)
```
illusional, line 213: This might unintentionally give you two cores based on the batch machine type.

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+    return {
+        'type': 'aip_output_json',
+        'is_singleton': '',  # not doing this at the moment
+        'is_exome': get_config()['workflow'].get('sequencing_type') == 'exome'
+    }
+
+
+@stage(
+    required_stages=[ValidateMOI],
+    analysis_type='custom',
```
illusional, line 259: Ditto

cpg_workflows/stages/aip.py
MattWellie: it would be good to get a specific analysis type for MTs, at the moment they're all logged as `custom`. If there's an MT type the output would never be empty

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+    )
+    result = query(my_query, variables={'dataset': dataset})
+    seq_type_exome = get_config()['workflow'].get('sequencing_type') == 'exome'
+    mt_path = ''
+    for analysis in result['project']['analyses']:
+        if analysis['output'].endswith('.mt') and (
+            (seq_type_exome and '/exome/' in analysis['output'])
+            or (not seq_type_exome and '/exome/' not in analysis['output'])
+        ):
+            mt_path = analysis['output']
```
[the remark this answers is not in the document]
MattWellie: will pick up everything and sort on timestamp - in my experience it's come back sorted but it will be good to guarantee that

cpg_workflows/jobs/aip/hpo_panel_match.py
```diff
@@ -0,0 +1,270 @@
...
+            panels={137},  # baseline panel is always mendeliome
+        )
+    return hpo_dict, all_hpo
+
+
+def match_hpo_terms(
+    panel_map: dict[str, set[int]],
+    hpo_tree: networkx.MultiDiGraph,
+    hpo_str: str,
+    max_layer_delta: int = 3,
```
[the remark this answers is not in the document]
MattWellie: I've simplified this out in latest commit - the intention was to play around with the layer delta, but... 3 seemed good enough so this is never varied

cpg_workflows/jobs/aip/hpo_panel_match.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    hpo_graph = read_obo(hpo_file, ignore_obsolete=False)
+
+    hpo_to_panels = {}
+    for hpo in all_hpos:
+        panel_ids = match_hpo_terms(
+            panel_map=hpo_to_panel_map,
+            hpo_tree=hpo_graph,
+            hpo_str=hpo,
+            max_layer_delta=MAX_DEPTH,
```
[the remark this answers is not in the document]
MattWellie: simplified in latest commit

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+    stage,
+    Dataset,
+    DatasetStage,
+    StageOutput,
+    StageInput,
+)
+from cpg_workflows.resources import HIGHMEM, STANDARD
+from cpg_workflows.jobs.aip.hpo_panel_match import main as panel_match_main
+
+CHUNKY_DATE = get_config()['workflow'].get('fake_date', datetime.now().strftime('%Y-%m-%d'))
```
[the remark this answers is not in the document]
MattWellie: This use case is entirely to do with running AIP in a fake retrospective way (i.e. simulate analysis as-of the given date, using representative panels and clinvar data from that time point). I don't think this use case needs to be translated over to production pipelines, so I'll scrap this

cpg_workflows/stages/aip.py
```diff
@@ -0,0 +1,294 @@
...
+        only one output, the panel data
+        """
+        return {'hpo_panels': dataset.prefix() / DATED_FOLDER / 'hpo_panel_data.json'}
+
+    def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None:
+        py_job = get_batch().new_python_job('create_panel_data')
+        py_job.image(get_config()['workflow']['driver_image'])
+
+        # only needs a really tiny image
+        STANDARD.set_resources(py_job, ncpu=1, mem_gb=1.0)
```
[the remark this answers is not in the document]
MattWellie: I've re-applied these using a miniscule VM requirement (`.cpu(0.25).memory('lowemem')`)

cpg_workflows/stages/aip.py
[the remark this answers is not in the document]
MattWellie: Adding a less granular timestamp for general use would be a cleaner solution, might have other uses

MattWellie (commented):
content from review

configs/defaults/aip.toml
```diff
@@ -0,0 +1,169 @@
...
+[references]
+genome_build = 'GRCh38'
+
+# per-cohort section
+#clinvar_filter = when running a clinvar re-summary, remove submissions by these sites. Default = no removal
+#cohort_panels = any panel IDs to be applied to this cohort, even if not otherwise phenotype-matched
+#cohort_percentage = when generating an analysis pedigree, reduce the cohort to this percentage of the total. Default = 100
+#gene_prior = if a specific gene list is to be used to determine Cat 2 (new gene-disease associations), provide the filepath here
+
+[cohorts.acute-care]
```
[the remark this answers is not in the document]
MattWellie: Open to it - I'm guessing the plan here would be to create this part of the config over in cpg-infra-private/aip, and mention something about needing a config from elsewhere in a README

cpg_workflows/jobs/aip/hpo_panel_match.py
```diff
@@ -0,0 +1,270 @@
...
+    gql query, pull out family details and HPO terms
+    may be a little overloaded at the moment
+    Args:
+        dataset (str): dataset name
+
+    Returns:
+        dict of per-participant details, and set of all HPO terms
+    """
+
+    query_string = gql(
```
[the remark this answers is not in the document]
MattWellie: I get an error with validate unless I call with validate `use_local_schema=True` - is this to do with running in my local environment, or is this something to do with how the validation operates?

cpg_workflows/stages/aip.py
MattWellie: Currently Custom is consistent with previous AIP runs, I've like to get on making a new specific type for these reports

illusional: Should we do this as part of this change? It's easy to add new analysis types, and we can migrate old entries very quickly (one SQL update).

MattWellie: I'm open to it... could we create two entry types? One for the report JSON, one for the HTML?

illusional: Yeah for sure, it's just a value in a table (with a foreign key), there's less friction, so avoiding overloading existing types I think is a better approach here

cassimons (approved):
Very cool to have this folded into the a pipeline!

PR #488, opened 2023-10-23
Update SNV es-index analysis meta with seqr data type

EddieLF:
A small change to add the `seqr-data-type` field to the es-index `meta` for SNV indexes. Relates to https://github.com/populationgenomics/metamist/issues/539.


EddieLF:
Changed the es-index meta field to be named `seqr-dataset-type`, since "`dataset_type`" is the name of the field seqr uses. 

Also changing the values of the field to be the expected "`dataset_type`" values found in the seqr repo [here](https://github.com/populationgenomics/seqr/blob/staging/seqr/models.py#L642-L645).

cassimons (approved):

PR #492, opened 2023-10-26
AIP tweaks, Stage wrapper arg to allow-missing outputs

MattWellie:
Tweak to the AIP default config to include the new category (see https://github.com/populationgenomics/automated-interpretation-pipeline/pull/327)

Secondary item, up for discussion:

- AIP reports come in 2 flavours: **all variants**, and **new variants in this run only**
- It's possible for a run to discover 0 new variants. If this situation occurs, no 'latest' output is generated, causing failures [here](https://batch.hail.populationgenomics.org.au/batches/429252/jobs/11)
- AIP reports are aggregated into an index (http://popgen.rocks/aip_index) which is done by pulling from the analysis entries, so we need to keep registering all outputs which exist.

Situation:
We want to register outputs in metamist, but we are aware that not all outputs may be created. 

I've stuck a new attribute in the @stage wrapper - `tolerate_missing_output`. This cascades down to the metamist status updater, and if the output file doesn't exist, instead of updating its attributes & checking object size, the method called should silently quit. 

Q. Is this a rational thing to be doing?
Q2. Is this the best way to accomplish this change?
Q3. Should the `in-progress` analysis entry for an output that doesn't exist be deleted instead of left hanging
Q4. **BIG Q** What do we accomplish by registering an analysis output as pending or in progress - surely metamist should only care about extant analysis outputs, not just potential ones (e.g. analysis object created, but the run fails... we'll just have ghost entries in metamist)

### DIFF
--- a/configs/defaults/aip.toml
+++ b/configs/defaults/aip.toml
@@ -35,7 +35,7 @@ gnomad_max_hemi = 1
 
 [csq]
 # variables affecting how the VCF variants are parsed, and AnalysisVariant objects are populated
-csq_string = ['allele', 'consequence', 'symbol', 'gene', 'feature', 'mane_select', 'biotype', 'exon', 'hgvsc', 'hgvsp', 'cdna_position', 'cds_position', 'amino_acids', 'protein_position', 'variant_class', 'ensp', 'lof', 'sift', 'polyphen']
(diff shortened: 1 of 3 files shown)


illusional:
I'm not a good person to BA or approve this change, leaving that to @vivbak.

MattWellie:
I'd actually be keen to scrap the `queued` and `in-progress` analysis entites we're currently creating, and just have one job to write an analysis `completed` entry upon stage completion. If we plan to use metamist as a source of file state I think it hurts us to have failed outputs in the db that we'll have to craft queries to avoid. 

Maybe there was a reason for this approach I'm not seeing?

@cassimons @vivbak

MattWellie:
Closing as it doesn't really solve the whole problem

PR #493, opened 2023-10-26
Scrap early metamist registrations

MattWellie:
All in on removing the analysis registrations we currently use. 

Current:
- one job is created and runs immediately, registering an output Analysis entry as `queued`
- when the relevant stage starts an update to that analysis entry is updated to `in-progress`
- when the relevant stage completes, update the analysis entry to `completed`

I can't see a reason for this, it means that any pipeline failures create ghost entries in metamist, which would have to be manually removed.

Instead of using a status reporter to register output entries before they exist, this queues a single job to follow the completed jobs, registering each output once.


vivbak:
Hey Matt! As discussed, after careful examination this change looks ok to me but could you please run some sort of test first. 

Note to people down the line investigating this PR: We could not find a reason for these jobs, other than to mirror hail batch behaviour. So we're scrapping them for now. However, that doesn't mean there _isn't_ a use case for them. So if things start breaking downstream, reverting this PR may be a good place to start.

MattWellie:
> could you please run some sort of test first
Yup! It's on my list now that there's some post-GATK-SV breathing room. I've been trying to think of a good test and may have just come up with one. Stay tuned 📺

MattWellie:
Test run: https://batch.hail.populationgenomics.org.au/batches/429501

vivbak (approved):

vivbak:
Ah you might wanna bump version again

PR #495, opened 2023-10-29
Further reduce dragmap threads

cassimons:
Follows #494.

n_threads on a 16 core machine == 32. Reducing this to 31 did not alter CPU load and alignment jobs still failed after ~ an hour. This reduces the request to n_threads -3 == ~31~ 29 threads, hopefully enough to give some head room.


EddieLF:
It will be 29 threads yeah? Either way we should see within an hour or two if it does any better or worse than 31/32!

EddieLF (approved):

PR #496, opened 2023-10-30
Update get_batch import to cpg-utils

illusional:
Relies on: https://github.com/populationgenomics/cpg-utils/pull/145

Centralise the wrapper, because it's useful for other places, including ensuring the `ar-guid` is transferred across to sub-batches.


MattWellie (approved):
that's pretty comprehensive

vivbak:
I am happy with this change

MattWellie:
@illusional @vivbak This is ready to roll - I found that since importing the batch wrapper from cpg_utils there's a lot of cross-test carry over. This caused failures in various tests aimed at counting the number of jobs, or checking the exact content of jobs. I couldn't find an easy/clean way of resetting the global `batch` between tests.

I've managed to get around that by labelling all the affected jobs and running them separately - feels like a clunky solution, but it gets the job done.

MattWellie:
I've merged this with main to iron out some discrepancies, #538 was an absolute mess

PR #497, opened 2023-10-30
Add GCS specific implementation for exists

illusional:
From a conversation wayyy ago, we discussed potentially rewriting exists to speed it up. This inserts a specific case for gs paths, that gets the contents of a directory, and checks if the file is in that list.

I'm less confident how to test this change.


MattWellie (approved):
I like it, I'd manually implemented something like this for the pipeline output existence checks. I'm happy to try and conjure up some mocked unit testing

Issue #498, opened 2023-10-31
Large Cohort Tests Failing 


vivbak:
Due to an out of data gnomad_methods import. Will rectify soon.

PR #499, opened 2023-10-31
Set thread count to core count

cassimons:
Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1698721857838289

Production pipelines currently assumes that hyper-threading is enabled on workers so the number of compute threads should equal double core count. However, [on GCP](https://cloud.google.com/compute/docs/instances/set-threads-per-core) Compute Engine, each virtual CPU (vCPU) is implemented as a single hardware multithread, and two vCPUs share each physical CPU core by default.

This PR sets the thread count per CPU to 1 It also reverts a previous change to the dragmap command so that it will set the number of worker threads to one less than the core count.

### DIFF
--- a/cpg_workflows/resources.py
+++ b/cpg_workflows/resources.py
@@ -29,7 +29,7 @@ class MachineType:
     """
 
     min_cpu: int = 2
-    threads_on_cpu = 2  # hyper-threading
+    threads_on_cpu = 1  # GCP does not have hyper-threading
 
     def __init__(
         self,
--- a/cpg_workflows/resources.py
+++ b/cpg_workflows/resources.py
@@ -29,7 +29,9 @@ class MachineType:
     """
 
     min_cpu: int = 2
-    threads_on_cpu = 1  # GCP does not have hyper-threading
+    # GCP supports one active thread per vCPU.
+    # See https://cloud.google.com/compute/docs/instances/set-threads-per-core
+    threads_on_cpu = 1
 
     def __init__(
         self,
--- a/cpg_workflows/jobs/align.py
+++ b/cpg_workflows/jobs/align.py
@@ -475,7 +475,7 @@ def _align_one(
         {prepare_fastq_cmd}
         dragen-os -r {dragmap_index} {input_params} \\
             --RGID {sequencing_group_name} --RGSM {sequencing_group_name} \\
-            --num-threads {nthreads - 3}
+            --num-threads {nthreads - 1}
         """
 
     else:


cpg_workflows/resources.py
```diff
@@ -29,7 +29,7 @@ class MachineType:
     """
 
     min_cpu: int = 2
-    threads_on_cpu = 2  # hyper-threading
+    threads_on_cpu = 1  # GCP does not have hyper-threading
```
jmarshall, line 32: The documentation linked to is not incredibly clear, but I think it is saying that the vCPU count is a count of hyper-threads, not a count of cores (= 2 hyper-threads). It goes on to say that two vCPUs will share a physical core by default (thus GCP does have hyper-threading), but you can switch that off (to maybe get performance/security/licensing benefits) but then you'll still be charged for each whole core even though you're only using one hyper-thread apiece.

So the comment is a bit misleading and might be better as something like “GCP vCPUs correspond to hyper-threads, not physical cores”.

jmarshall (approved):
Imma let you merge your PR, but it feels like altering `threads_on_cpu` (code that's untouched since Vlad's initial add in October 2022) is a big change that needs an issue and careful consideration rather than being a byproduct of an alignment PR. `get_nthreads()` is used in nine _cpg_workflows/jobs/*.py_ source files, each of which probably ought to have its thread settings audited.

Having said that, I see that [batch 429255 job 1](https://batch.hail.populationgenomics.org.au/batches/429255/jobs/1) ends up with `"resources": { …, "req_cpu": "16", … }` but has `dragen-os […] --num-threads 31`. I have not checked every layer of code in play to see that they are all measuring threads rather than physical cores, but this is very suggestive that the proposed fix is right.

…Now it seems incredible that we have not always been failing regularly in this way!…

So I certainly support merging this and observing the effects on the `dragen-os` step. But I think we'll have some further work to do if this pans out!

jmarshall:
(Also meant to mention that the current test failure is due to #498 so can be ignored for this PR.)

cassimons:
> It feels like altering threads_on_cpu (code that's untouched since Vlad's initial add in October 2022) is a big change that needs an issue and careful consideration rather than being a byproduct of an alignment PR. get_nthreads() is used in nine cpg_workflows/jobs/*.py source files, each of which probably ought to have its thread settings audited.

Thanks John, I am happy if you think we should attach a ticket to this, but I see this as a clear bug fix rather than a change. `get_nthreads` has been returning an erroneous result and all current use cases of the output are based on incorrect assumptions at this point. 

I have reviewed all uses of `get_nthreads` and `max_threads` in production pipelines. In all cases, the output is passed directly to the worker threads (or equivalent) argument with the implicit assumption that this value represents the maximum number of concurrent active threads available on a worker. Given our direct use of `get_nthreads` and the tools it is passed to (aligners, samtools, bcftools etc)  It is difficult to imagine a scenario where this will harm performance.

cpg_workflows/resources.py
cassimons: Good point. I made it more explicit.

PR #506, opened 2023-11-04
Omit samples field for GATKSV callsets

cassimons:
Final gatk sv mt currently contains a `samples` field that lists all samples in the call set. This results in variants always being returned in seqr searches reguardless of genotype filter settings.  Removing the field should result in proper search results being returned.


MattWellie (approved):
Sounds good!

PR #507, opened 2023-11-04
Omit samples_cn field from gatksv final MT

cassimons:
As per https://github.com/populationgenomics/production-pipelines/pull/506

samples_cn fields appear to be gCNV callset specific and have a non-intuative effect due to the nature of gCNV callsets. For example the samples_cn_2 field is  used to select ALT_ALT variants, presumably because this genotype is only asigned in a gCNV callset when this CN is alt (ie on chrX).

PR #506 resolved most seqr genotype query issues with SV call sets. This PR should resove the remaining problem causing default recessive filters to return all ref/ref calls.

### DIFF
--- a/cpg_workflows/query_modules/seqr_loader_sv.py
+++ b/cpg_workflows/query_modules/seqr_loader_sv.py
@@ -461,13 +461,17 @@ def annotate_dataset_sv(mt_path: str, out_mt_path: str):
                 for i in range(0, 90, 10)
             }
         ),
-        samples_cn=hl.struct(
-            **{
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
Will have a better look tomorrow - sounds good, would there be value in populating this field for CNV calls within the broader GATK-SV callset?

cassimons:
No I don't think so. Gatk adds full ref/alt calls for each individual and these fields are used for genotype filtering. The cnv_n fields are only needed for pure gcnv (non gatksv) call sets.

PR #508, opened 2023-11-05
Rethink concordance filtering

MattWellie:
For discussion 
- removes all the annotations at the genotype level which are based on Concordance with a prior dataset (not relevant to our implementation at this time)
- fields are removed without replacement (how will Seqr/ES feel about this?)
- also overwrites the RSID (which was copied into the Row-field `variantId` with missing String values, to prevent population of non-existent dbSNP hyperlinks


cassimons (approved):
These changes all look good to me thanks.

`rsid=hl.missing('tstr')`: Sometimes hail seems to go out of its way to be non intuitive to read.

PR #509, opened 2023-11-06
Add config for machine type used for alignment.

cassimons:
Context: Further testing has identified upping the memory available during the alignment step mitigates the resource exhaustion bug discussed here https://centrepopgen.slack.com/archives/C030X7WGFCL/p1698189973862839.

This PR adds a resource override configuration allowing us to set the machine type used for alignment to HIGHMEM for problematic samples. We do not want to do this by default as this will significantly increase costs and the majority of samples align successfully with a STANDARD machine size


cpg_workflows/jobs/align.py
```diff
@@ -346,7 +346,12 @@ def _align_one(
     job_name = f'{job_name} {alignment_input}'
     j = b.new_job(job_name, job_attrs)
 
-    nthreads = STANDARD.set_resources(
+    if get_config()['resource_overrides']['align_use_highmem']:
```
jmarshall, line 349: Should probably use `.get('align_use_highmem')` here, so the tests can get away without a dummy entry.

cpg_workflows/defaults.toml
```diff
@@ -105,6 +105,9 @@ delete_scratch_on_exit = false
 # defaults in code are 40 for genomes, none for exomes
 #haplotypecaller_storage = 80
 
+# Use highmem machine type for alignment step
+align_use_highmem = false
```
jmarshall, line 109: I guess with `.get()`, this could be commented-out-by-default too, like the other resource_overrides entries.

jmarshall (approved):

PR #511, opened 2023-11-07
Script to find SGs needing SV calling

MattWellie:
Candidate script to generate a list of samples to start off a fresh GATK-SV batch

- give it a number of samples to aim for, and a list of projects in order of priority
- iterate through projects
  - find all samples in the project (remove excluded samples, if exclusion file is added)
  - find subset of samples with SV calls already, remove from consideration
  - add all samples in the subset yet to be called (unless that would exceed the number specified, in which case add a subset using random.sample)
- if a file of 'additional' samples is specified, add all of those too
- write to output file

Completely functionally unrelated change crowbar'd in here to ride the wave of an approved PR: Labels!

There is an issue whereby Cromwell implements strict RegEx limitations on GCP labels, which we are relying on for billing auditability. This PR:

- updates the billing labels to conform to `[a-z]([-a-z0-9]*[a-z0-9])?`
- adds the `ar-guid` label
- prefixes `ar-guid`'s value with `ar-` to prevent failure when the first digit happens to be a number
- adds the Cromwell Stage (lowercase) as a label
- lowercases the CPG ID
- sequencing_group -> `sequencing-group`

### DIFF
--- a/scripts/find_samples_needing_sv.py
+++ b/scripts/find_samples_needing_sv.py
@@ -113,7 +113,7 @@ def main(
         additional (str | None): a json file containing sequencing groups to include specifically
     """
 
-    collected_sgs = set()
+    collected_sgs: set[str] = set()
 
     # get excluded SGs if that's valid
     if exclude is not None:
--- a/scripts/find_samples_needing_sv.py
+++ b/scripts/find_samples_needing_sv.py
@@ -1,0 +1,42 @@
+#!/usr/bin/python3
+
+"""
+a script to find samples that need to be run through the structural variant pipeline
+(i.e. are not present yet in any extant callsets)
+
+usage:
+    python3 find_samples_needing_sv.py
+    -n <min_#_samples>
+    -o <output_path>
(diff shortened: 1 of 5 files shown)


scripts/find_samples_needing_sv.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/python3
+
+"""
+a script to find samples that need to be run through the structural variant pipeline
+(i.e. are not present yet in any extant callsets)
+
+usage:
+    python3 find_samples_needing_sv.py
+    -n <min_#_samples>
```
cassimons, line 9: Can you add something that explains what -n does?

cassimons (approved):

scripts/find_samples_needing_sv.py
MattWellie: updated! Also switched this to max_samples to reflect what it intends to do

violetbrina (approved):
Looks good to me. Thanks for adding the `ar-guid` in particular.

PR #512, opened 2023-11-07
Corrects ar_guid -> ar-guid

MattWellie:
silly typo from me

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.1
+current_version = 1.18.2
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


illusional (changes requested):
Can you use `try_get_ar_guid` (https://github.com/populationgenomics/cpg-utils/blob/83e0470687128a6880f5ae75dfa244f36c376650/cpg_utils/config.py#L47C5-L47C20)

You can import they `ar-guid` billing label as `AR_GUID_NAME`: https://github.com/populationgenomics/cpg-utils/blob/83e0470687128a6880f5ae75dfa244f36c376650/cpg_utils/config.py#L11

illusional (approved):
LGTMichael

PR #513, opened 2023-11-07
Vep and ar-guid fixes

MattWellie:
- Pins the cpg-utils requirement to a point post `ar-guid` (currently at 4.15.1 in the driver image, so is not updated and is broken relative to code on main)
- removes the no-stats VEP flag (see [issue](https://github.com/Ensembl/ensembl-vep/issues/1013#issuecomment-867595782) affecting [batch](https://batch.hail.populationgenomics.org.au/batches/429519/jobs/3772))

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.2
+current_version = 1.18.3
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


EddieLF (approved):
Looks good!

MattWellie:
@illusional I've pushed a bonus commit - the Dockerfile un-pins the driver image to build on, and reverts back to latest (we don't have that Hail batch issue any more in the latest driver). Is there any reason not to attach to `latest`?

illusional:
I think this is fine, we'll discuss at the software team / data office hours generally about software decencies

PR #514, opened 2023-11-08
Remove Dataproc in RD pipeline where possible

MattWellie:
Dataproc was initially required for a few key RD pipeline steps:

1. Collating VEP JSON files into a Hail Table
2. Annotating MT with VEP data
3. Creating per-dataset MT ready for Seqr
4. Translating Seqr-ready MT to an ElasticSearch index

Of these, only `4` still requires dataproc - all other features are implemented in Hail Query. This PR would remove dataproc usage where it isn't required. Not sure removing all the scripts/code is essential, we could just toggle this behaviour off by default - but all usage of dataproc in these Stages is unnecessary

---

Since the latest update to our Dataproc image, we're now able to run an up-to-date hail version inside the dataproc image. This means that we can interchangeably run DP and non-DP jobs, with files readable from both versions. Personal preference is still non-DP, as it's faster and cheaper to run, and easier to debug.

This PR has been rebased on main, and specifically covers:

- removing the dataproc version of AnnotateCohort (and the relevant wrappers/scripts to facilitate running in DP)
- remove the dataproc running of AnnotateDataset (and wrappers)
- VEP JSON -> Hail.Table is now done in QoB
- Moves around the loading of reference data objects in AnnotateCohort (this was a bit of a debugging odyssey, I'll revert as much of that as possible)
- Adds an argument when reading VCFs so that we don't run into issues writing/checkpointing (relevant Zulip thread as a comment, FYI @EddieLF)
- Checkpointing is _not_ done by default (the checkpoint method includes a allow_reuse argument, defaulting to False in the RD pipeline)


cassimons (commented):
I like the idea of this, but I am not comfortable merging until we have done an end-end test run with it. I can kick one off later on if that helps?

MattWellie:
> I can kick one off later on if that helps?

Sounds doable. Repeating VEP/AnnotateDataset/Cohort with that validation MT could be just the ticket

EddieLF:
Good writeup @MattWellie 

So the steps to take might be
1. Pin new Hail version to the dataproc AR script [dataproc.py](https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/dataproc.py#L22), so that we are running Hail 0.2.118 in dataproc.
2. Modify the [install_common.sh](https://github.com/populationgenomics/analysis-runner/blob/main/dataproc/init_scripts/install_common.sh) dataproc init script to use the latest `cpg_workflows`, `cpg_utils`, `analysis_runner`, etc.

Or, we could modify the script so that it installs the latest versions of every package? To do this, after the `pip install` lines we could add 
`pip freeze > requirements.txt`
which will save the current packages to the `requirements.txt` file with `==<version>` next to each package name.

Then update the `==` to `>=` and pipe the results to `pip install --upgrade`
`sed 's/==/>=/' requirements.txt | pip install --upgrade -r /dev/stdin`

Which should update each package to the latest available that can be found in the package index. Is this what we want?

MattWellie:
@EddieLF that approach would definitely bring us up to the latest versions of relevant packages, would result in a massive update and dependency resolution process. It would suffer from a similar issue - we wouldn't know in advance which versions are eventually used.

so long as we don't update the gnomad-methods submodule we should be Ok for the RD pipeline, the bigger concern is the large-cohorts pipeline which appears to use dataproc much more widely

MattWellie:
Considering this as on-hold, while other Dataproc issues and updates are pursued

cassimons (approved):
Tested on a validation run a LGTM! So long data proc! (almost)

PR #515, opened 2023-11-08
Cache GCS exists 2; iterdir boogaloo

MattWellie:
Reimplementation based on @illusional's #497, with a slight adjustment of how the cloud dirs are tested.

Swaps out the google storage client call for a cloudpathlib iterdir. The former was having issues for me - returning empty sets on all directories. 

Side issue is that `dirname` or `basename` calls on the GSPath objects seems to trigger a get, which is usually not permitted as a local user. Translating to a Str first solved this.

For discussion - 

The pipeline has a version of this written in which is aiming to do the same thing, but in a really inelegant way (i can say that, I wrote it), I'd be keen to swap out that whole process for this one.

In addition, how do you feel about migrating this functionality to cpg-utils instead of keeping it pipeline specific?


illusional (approved):
I thought we'd had issues with `iterdir` in the past, that it actually iterated over everything in the bucket (and maybe subdirectories) internally, and filters on the client making it quite slow - but I ran a quick test and I don't think that's the case anymore.

If this worked in your testing, happy to approve (but I think @vivbak should be across this as well).

vivbak (dismissed):

vivbak (approved):

MattWellie:
I'm pretty content to merge this..
lil test case

- https://batch.hail.populationgenomics.org.au/batches/429916 a test batch set off, cancelled after the first two stages completed
- the same batch restarted, the first two stages are not repeated (output found)

PR #517, opened 2023-11-12
Cache GCS exists redux: implementation

MattWellie:
Stacked PR based on #515, takes the LRU-cached existence checking and implements during the workflow setup phase.

Strips out all the garbage that I previously wrote in to do cached object existence tests. Also removes the test_utils script, which was only testing that manual result caching implementation. I'll run a test job and see how this goes.


illusional (approved):
Approve in principle, but wait for @vivbak's review.

vivbak (changes requested):
Can we look at the failing tests first? 
Also, it's worth noting that the plan was to always remove these checks entirely, but until then this change is fine.

MattWellie:
@MattWellie look at those tests - re-add mocking but for the methods being used

MattWellie:
@vivbak the test fails were due to this
https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/utils.py#L137

Because the cached check will work for GCS or local Paths, they both now route through that method. The cloudpathlib iterdir method doesn't like it when the parent folder doesn't exist at all (as is the case with the pytest temp file locations, e.g. `/private/var/folders/rb/rtd27jzx1sdgrp3p4ph2cjv00000gp/T/pytest-of-mwelland/pytest-18/test_get_intervals_valid_input11/intervals/1.interval_list`)

I've patched all the relevant failures to deliver the expected False/True depending on whether the test needs to exist, and added a exception catcher for the FileNotFound error - if that occurs (parent dir doesn't exist) we just print the exception text and return False.

vivbak (approved):
🙏

PR #518, opened 2023-11-13
Correct bug, repartition table

MattWellie:
I got pretty into the weeds over a pipeline failure and wasn't seeing the forest for the trees. Here's a greenfield solution that solves the root cause without sprouting into a massive rethink of the prod-pipes... ecosystem? I'm all out of horticultural metaphors. 

TL; DR: corrects a typo, retains the 're-partition my MT/Table on checkpoints

### DIFF
--- a/cpg_workflows/utils.py
+++ b/cpg_workflows/utils.py
@@ -70,27 +70,7 @@ def checkpoint_hail(
         return read_hail(path)
 
     logging.info(f'Checkpointing {path}')
-    t = t.checkpoint(path, overwrite=True)
-
-    # throw in a repartition
-    if file_name.endswith('.mt'):
-        # estimate partitions; fall back to 1 if low row count
-        current_rows = t.count_rows()
-    else:
-        current_rows = t.count()
-
-    partitions = current_rows // 200000 or 1
-
-    # no need to alter partitions, do not shuffle
-    if t.n_partitions() == partitions:
-        return t
-
-    logging.info(
-        f'Re-partitioning {current_rows} into {partitions} partitions'
-    )
-
-    # don't name partition arg, 'partitions' for Table, 'n_partitions' for MT
-    return t.repartition(partitions, shuffle=True)
+    return t.checkpoint(path, overwrite=True)
 
 
 def missing_from_pre_collected(test: set[Path], known: set[Path]) -> Path | None:
--- a/cpg_workflows/utils.py
+++ b/cpg_workflows/utils.py
@@ -38,16 +38,59 @@ def read_hail(path):
     return t
 
 
-def checkpoint_hail(t, file_name: str, checkpoint_prefix: str | None = None):
-    if checkpoint_prefix:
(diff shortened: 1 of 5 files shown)


EddieLF (approved):
Looks like this wood work!

cpg_workflows/utils.py
```diff
@@ -38,16 +38,59 @@ def read_hail(path):
...
+    path = join(checkpoint_prefix, file_name)
+    if can_reuse(path) and allow_reuse:
+        logging.info(f'Re-using {path}')
+        return read_hail(path)
+
+    logging.info(f'Checkpointing {path}')
+    t = t.checkpoint(path, overwrite=True)
+
+    # throw in a repartition
+    if file_name.endswith('.mt'):
```
cassimons, line 76: Surely re-partitioning should be optional?  eg the PA team will have MTs with very different shapes to ours so the portion logic would be different. 

I am not sure if we should be hiding extra logic here or explicitly adding it where needed in the jobs.

cpg_workflows/utils.py
```diff
@@ -38,16 +38,59 @@ def read_hail(path):
...
+    # no need to alter partitions, do not shuffle
+    if t.n_partitions() == partitions:
+        return t
+
+    logging.info(
+        f'Re-partitioning {current_rows} into {partitions} partitions'
+    )
+
+    # don't name partition arg, 'partitions' for Table, 'n_partitions' for MT
+    return t.repartition(partitions, shuffle=True)
```
cassimons, line 93: If we are going to repartition, should we do that before the checkpoint is run?

cpg_workflows/utils.py
MattWellie: I've removed the repartition logic, it is indeed a bit hidden here. This specific method is used exclusively by the seqr_loader pipeline for now, so there are no other implications, but I've removed it anyway

cpg_workflows/utils.py
MattWellie: This is a holdover from the AIP implementation, where repartitioning is done during the filtering process.

In that context, each checkpoint will have fewer and fewer entries, and repartitioning is done to prevent sparse/empty partitions occurring. To estimate the required number of partitions you need to count the rows. Counting the remaining rows causes all the hail operations to trigger and then the results get thrown away and repeated, so checkpointing first and repartitioning second in that context makes sense.

MattWellie: The one I'm not sure about is whether a Batch backend repartition will **always** include a checkpoint into tmp. There's some code I don't really understand there...
```
if hl.current_backend().requires_lowering:
    <do something resulting in a checkpoint in the batch scratch space>
```

but I haven't been able to find a definition for requires lowering

MattWellie: ANYWAY, this data will only need to be repartitioned once, so will rethink the location of this outside of the checkpointing method so it's not hidden

cassimons (approved):

PR #519, opened 2023-11-13
Allow ES-Index name registration without file checks

MattWellie:
ES Indexes are registered as names, not files. Skip pesky file checks (separate logic from the 'permit output to be missing')

As discussed on Slack


EddieLF (approved):
Good solution, the ES-Index analyses are distinct from all the other analysis types in that their output exists outside of the project buckets so making an exception here works.

PR #521, opened 2023-11-14
Reduce Metamist logging

MattWellie:
Wraps the metamist GQL query calls with a reduced-granularity logging profile. Prevents the full un-parsed metamist query response being printed into Hail logs (often a multi-MB text dump)

First (?) get_config call is set to 'print result to log', in case https://github.com/populationgenomics/cpg-utils/pull/146 is merged

### DIFF
--- a/cpg_workflows/metamist.py
+++ b/cpg_workflows/metamist.py
@@ -14,7 +14,7 @@ from cpg_utils.config import get_config
 from metamist import models
 from metamist.apis import AnalysisApi
 from metamist.exceptions import ApiException
-from metamist.graphql import gql, query, validate
+from metamist.graphql import gql, query
 
 from cpg_workflows.filetypes import (
     AlignmentInput,
@@ -24,6 +24,7 @@ from cpg_workflows.filetypes import (
     FastqPairs,
 )
 from cpg_workflows.utils import exists
+
 
 GET_SEQUENCING_GROUPS_QUERY = gql(
     """
--- a/cpg_workflows/metamist.py
+++ b/cpg_workflows/metamist.py
@@ -5,7 +5,6 @@ Helpers to communicate with the metamist database.
 import logging
 import pprint
 import traceback
-from collections import defaultdict
 from dataclasses import dataclass
 from enum import Enum
 from typing import Optional
@@ -87,6 +86,19 @@ GET_PEDIGREE_QUERY = gql(
 
 
 _metamist: Optional['Metamist'] = None
+
+
+def silent_query(query_to_run, query_params: dict | None = None):
+    """
+    Run a query, but don't log the results to the console
+    Ref: https://github.com/populationgenomics/metamist/issues/540
+    """
+
+    curr_level = logging.root.level
+    logging.getLogger().setLevel(logging.WARN)
+    result = query(query_to_run, variables=query_params)
+    logging.getLogger().setLevel(curr_level)
+    return result
 
 
 def get_metamist() -> 'Metamist':
@@ -226,7 +238,7 @@ class Metamist:
         if only_sgs and skip_sgs:
             raise MetamistError('Cannot specify both only_sgs and skip_sgs in config')
 
-        sequencing_group_entries = query(
+        sequencing_group_entries = silent_query(
             GET_SEQUENCING_GROUPS_QUERY,
             {
                 'metamist_proj': metamist_proj,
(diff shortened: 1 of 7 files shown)


cpg_workflows/workflow.py
```diff
@@ -929,7 +929,7 @@ def __init__(
                 'Workflow already initialised. Use get_workflow() to get the instance'
             )
 
-        self.dry_run = dry_run or get_config()['workflow'].get('dry_run')
+        self.dry_run = dry_run or get_config(True)['workflow'].get('dry_run')
```
vivbak, line 932: Since this is the only change where true is passed to get_config, does that mean that we'd only print the config during a dry-run?

cpg_workflows/metamist.py
```diff
@@ -89,6 +89,29 @@
...
+    Allow for a config override to log the results to the console
+
+    Args:
+        query_to_run (gql DocumentNode): Query String
+        query_params (dict): Query Parameters or None
+
+    Returns:
+        query result, with or without logging
+    """
+    if get_config()['workflow'].get('log_metamist', False):
```
vivbak, line 105: We should name this file better, but until we have a better solution, can all new parameters be added to `defaults.toml` with documentation (a comment) on what they do.

cpg_workflows/metamist.py
```diff
@@ -89,6 +89,29 @@
 _metamist: Optional['Metamist'] = None
 
 
+def silent_query(query_to_run, query_params: dict | None = None):
```
vivbak, line 92: I'm wondering if we could change the name of this. It isn't really a silent query, because if log_metamist is set then it will still be be a very noisy query.

cpg_workflows/metamist.py
```diff
@@ -15,7 +14,7 @@
 from metamist import models
 from metamist.apis import AnalysisApi
 from metamist.exceptions import ApiException
-from metamist.graphql import gql, query, validate
```
vivbak, line 18: Thanks! Wild that my linter didn't catch this if it isn't used anywhere.

vivbak (changes requested):
Thanks Matt! Logging has been something that hasn't been done well for a while. This is a great step forward.

cpg_workflows/workflow.py
vivbak: Ah sorry, I mis-read. Ignore my above comment.

MattWellie: As far as I can track back this is the first instance of get_config being called in the workflow startup process. I'm not sure if we'd want to put a True in the next one as well, in case we run with `dry_run==True`, which would make the first get_config() call the one after this.

cpg_workflows/metamist.py
MattWellie: That is a fair comment... maybe `gql_query_optional_logging`?

cpg_workflows/metamist.py
MattWellie: Added the defaults.toml entry

vivbak (approved):
Awesome! Thanks :D

vivbak:
oh wait can you look at the linter first

cpg_workflows/workflow.py
```diff
@@ -929,9 +929,9 @@ def __init__(
...
 
-        analysis_dataset = get_config()['workflow']['dataset']
+        analysis_dataset = get_config(True)['workflow']['dataset']
```
jmarshall, line 934: Won't it still print it twice for non-dry-runs?

jmarshall: Oh… only does the printing if `_config` actually got consed up.

:shrug: Okay, or could be clearer if the `self.dry_run` line was moved below the other line's block of four lines. Or really we could have `Workflow.__init__` start with `config = get_config(True)` and use that instead of calling `get_config()` a zillion times…

MattWellie: Yeah, moving the dry run would have made sense, I should have done that. I tried banking the get_config call into a dummy variable first but the linter didn't like that.

MattWellie: The get_config calls are a bit weird looking, but I prefer the consistency compared to pulling it into an object and having to pass the object around.

jmarshall: No worries. Geez linter 🙃

PR #522, opened 2023-11-15
Interim quick fix: Account for GB vs MiB in JobResource.get_java_mem_mb()

jmarshall:
We will soon rewrite this Java option facility to be more flexible, but in the interim let's correct the mathematics. While `get_mem_gb()` is measured in decimal GB, `-Xms` uses `get_java_mem_mb()` as binary MiB. Hence the correct multiplier is 1000<sup>3</sup> / 1024<sup>2</sup> rather than 1000. (At present, the JVM is sometimes being told to use more memory than Hail has allocated to the job, leading to intermittent failures and a need to use the big hammer of the `picard_mem_gb` override.)

Context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1698815009640079?thread_ts=1698788154.906359&cid=C018KFBCR1C and the surrounding thread.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.6
+current_version = 1.18.7
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -11,7 +11,7 @@ on:
         default: "test"
 
 env:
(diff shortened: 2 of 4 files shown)


MattWellie (approved):
how frustrating

vivbak (approved):
love this find though

jmarshall:
You would not believe how carefully I had to do this — on paper — to figure out exactly what the problem was, and what the right answer was!

PR #523, opened 2023-11-16
Update default variant sites table

KatalinaBobowik:
The default variants table was being pulled from `test`, rather than `main`.

For more context, this sites table has been chosen as a representative sample of variation across diverse cohorts after testing several sites tables. See slack thread here: https://centrepopgen.slack.com/archives/C04KZFM1TC7/p1679895423746189


vivbak (approved):

vivbak:
Good catch!

PR #526, opened 2023-11-17
Updated script for adjusting `@SQ` headers in existing CRAM files

jmarshall:
This re-envisions PR #206 to pull metadata from metamist via GraphQL and uses a Python job so that the helper script can be incorporated into this main script.


scripts/fix_sq_headers.py
```diff
@@ -0,0 +1,195 @@
...
+        id
+        meta
+        output
+        active
+        type
+      }
+    }
+  }
+}
+""")
```
michael-harper, line 38: Just performed this query myself and it outputs files that don't exist. This should be fine because you check the existence of files prior to performing reheadering. But is there a possibility of a file we don't want to reheader but also, don't want to delete? 

I'm wondering if it would be best to pull the latest cram analyses 

So the query would look like this (with `timestampCompleted` included):
```
query CramQuery($project: String!) {
  project(name: $project) {
    sequencingGroups {
      id
      analyses(type: {eq: "cram"}) {
        id
        meta
        output
        active
        type
        timestampCompleted
      }
    }
  }
}
```
And then sort the analyses based on `timestampCompleted`
```
sorted_analyses = sorted(
  analyses,
  key=lambda x: datetime.strptime(
      x['timestampCompleted'], '%Y-%m-%dT%H:%M:%S'
  ),
)
(snippet from Eddie's auditing scripts)
```

michael-harper: Actually, this may just capture the NAGIM crams, which if I understand correctly, don't have the issue this script is trying to solve 🙃 Leaving this here in case I'm wrong!

michael-harper (commented):
Looks good to me! Just a few questions on my end. I'm still not totally confident with config files and how they're structures and used so someone else's review may add more insight 😅

scripts/fix_sq_headers.py
```diff
@@ -0,0 +1,195 @@
...
+
+        reheader_list = []
+        for fname in cram_list:
+            path = to_path(fname)
+            reheadered_path = path.parent / 'reheadered' / path.name
+
+            if not path_exists(path):
+                print(f'{fname}: file does not exist')
+            elif path_exists(reheadered_path):
+                print(f'{reheadered_path}: reheadered file already exists')
```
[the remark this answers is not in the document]
jmarshall: Not that I'm aware of, and that `…/reheadered/…` scheme was copied from PR #206. However I don't believe that PR's script was run in earnest either.

scripts/fix_sq_headers.py
```diff
@@ -0,0 +1,195 @@
...
+
+        print(f'CRAM files to be reheadered: {len(reheader_list)}')
+
+        b = get_batch(f'Reheader {dataset} to {new_reference_path}')
+        new_reference = b.read_input(new_reference_path)
+
+        for path, newpath, filesize in reheader_list:
+            j = b.new_python_job(f'Reheader {path}')
+            j.image(image_path('samtools'))
+            j.storage(filesize * 1.2)
```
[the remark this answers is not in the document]
jmarshall: There's a small amount of additional space required on `/io` for the _.crai_ index and a few other files, but mainly this multiplier was plucked out of the air 🤷 

Fortunately these jobs are quite short-lived, so it probably doesn't matter too much…

scripts/fix_sq_headers.py
jmarshall: The script doesn't do any deleting itself yet, so it's safe from this point of view.

However [running it in test](https://batch.hail.populationgenomics.org.au/batches/430059/jobs/1) does come up with a number of seemingly faulty database records with filenames that aren't CRAM or don't exist, which the script is careful to check for. I am hoping that the data will be cleaner in main, but something like this may be a good idea.

So we might have multiple CRAM analyses for a sequencing group, and we ought to only look at the most recent ones? Actually I guess I should be paying attention to the `active` field too…

scripts/fix_sq_headers.py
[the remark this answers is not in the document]
jmarshall: This check also caters for the possibility that we've already run this script on a subset of the TOB files and are re-running it on the remainder.

vivbak (changes requested):
As per live review session :)

scripts/fix_sq_headers.py
jmarshall: Rather than look at the timestamps and active field, we've (for now) de-emphasised querying these paths from metamist in favour of just specifying a bucket path on the command line / in the config file.

jmarshall:
Okay, this is ready for re-review with an eye to it being ready for merging to **main** so that we can at least dry-run it against the main dataset. I also requested a review from @illusional in case he wants to check on the metamist updating too.

Now updated to get the list of CRAM files either by scanning a directory URL or by querying metamist, and to add an analysis record for each CRAM file that it reheaders. In dry-run mode it reads the whole CRAM file to the worker VM (because this is done by hail) but just reads and reports on the existing header; unlike normal mode, it doesn't reheader it, write it out to the `…/reheadered/…` directory, or add to metamist.

I also considered setting the Analysis's `author`, `timestamp`, `project`, and `active` fields, but leaving them out results in very suitable defaults being used, so this is fine as is.

This should typically be used with a config file like this _reheader.toml_:

```toml
[workflow]
dataset = 'tob-wgs'
access_level = 'test'
#access_level = 'full'
#dry_run = true
source = 'gs://cpg-tob-wgs-test/cram/nagim'

[references]
expected = 'Homo_sapiens_assembly38[unmasked]'
new = 'gs://cpg-common-main/references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.dict'
newset = 'Homo_sapiens_assembly38_masked'

[images]
samtools = 'australia-southeast1-docker.pkg.dev/cpg-common/images/samtools:1.18'

[hail]
#dry_run = true
billing_project = 'tob-wgs'
```

and an invocation something like

```
analysis-runner \
  --dataset tob-wgs --description 'reheadering test run' --access-level test \
  --output-dir johnm-test-reheader \
  --image australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:latest \
  --config reheader.toml --config configs/genome.toml \
  scripts/fix_sq_headers.py

```

scripts/fix_sq_headers.py
```diff
@@ -0,0 +1,229 @@
...
+            'source': f'Reheadered from {oldpath}',
+        },
+    )
+
+    aid = AnalysisApi().create_analysis(dataset, analysis)
+    print(f'Created Analysis(id={aid}, output={newpath}) in {dataset}')
+
+
+if __name__ == '__main__':
+    config = get_config(True)
```
illusional, line 161: Nitpick, can you wrap this in a function that gets called with the args?

scripts/fix_sq_headers.py
```diff
@@ -0,0 +1,229 @@
...
+
+    b = get_batch(f'Reheader {dataset} to {new_reference_path}')
+    new_reference = b.read_input(new_reference_path)
+
+    for path, newpath, filesize in reheader_list:
+        j = b.new_python_job(f'Reheader {path}')
+        j.image(image_path('samtools'))
+        j.storage(filesize * 1.2)  # Allow some extra space for index file, references, etc
+
+        j_result = j.call(do_reheader,
```
illusional, line 207: This doesn't feel like the regular black formatting. Can you pass it in using kwargs (eg: `dry_run=dry_run, ...`) for clarity?

illusional (commented):
Thanks for the tag @jmarshall! A few comments:

- This will always be overridden by the analysis-runner:
```toml
[workflow]
dataset = 'tob-wgs'
access_level = 'test'
```

- Some of these are already provided by the analysis-runner so you don't need to provide them
```toml
[images]
samtools = 'australia-southeast1-docker.pkg.dev/cpg-common/images/samtools:1.18'

[hail]
#dry_run = true
billing_project = 'tob-wgs'
```

- You don't need to specify the cpg_workflows image, as the default provided driver image contains everything you need (saves a step in the invocation

    Can you move the general inputs from the config to a CLI, so the invocation becomes something like the following. We usually use click, but some people prefer straight python argparse:

    ```bash
    analysis-runner \
      [... params] 
      scripts/fix_sq_headers.py --expected-ref <> --new-reference <> --dry-run # [... etc]
    ```

    This should also simplify the analysis-runner invocation to drop the configs

scripts/fix_sq_headers.py
jmarshall: I could, but does it really matter for a ~200 line one-shot script under some time pressure?

jmarshall:
> * This will always be overridden by the analysis-runner:
>   ```toml
>   [workflow]
>   dataset = 'tob-wgs'
>   access_level = 'test'
>   ```

Fair point.

> * Some of these are already provided by the analysis-runner so you don't need to provide them
>   ```toml
>   [images]
>   samtools = 'australia-southeast1-docker.pkg.dev/cpg-common/images/samtools:1.18'
>   ```

I didn't realise the previous 1.16.1 (or so) one was no longer the default, so I expected I would have to override this to get this one. Hmmm, if I realised updating the samtools image would affect all other production steps that ask for the `samtools` image, I might have checked it a bit more carefully…

>   Can you move the general inputs from the config to a CLI, so the invocation becomes something like the following. We usually use click, but some people prefer straight python argparse:
>   ```shell
>   analysis-runner \
>     [... params] 
>     scripts/fix_sq_headers.py --expected-ref <> --new-reference <> --dry-run # [... etc]
>   ```

I had put everything in the config file as it was an easy way to ensure the settings were logged for provenance and repeatability. And also convenient for rerunning the script on different subsets or in case of failures.

Certainly happy to convert it to a command line option interface instead, though it's not entirely consistent with the quick hack nature of this one-shot script…

illusional:
On the CLI makes it easier to see in places like the analysis runner log, you don't need to separately load a config and check the relevant paramo for the script. 

I'd push for this harder than the other nitpicks

scripts/fix_sq_headers.py
jmarshall: Done, as a byproduct of converting to use command-line options via click.

jmarshall:
Hopefully final final version of this. Now invoked using command line options, e.g. thus:

```
analysis-runner \
  --dataset tob-wgs --description 'reheadering test run' --access-level test --output-dir tmp \
  --config configs/genome.toml \
  scripts/fix_sq_headers.py \
  --ref gs://cpg-common-main/references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.dict \
  --expected 'Homo_sapiens_assembly38[unmasked]' --intended Homo_sapiens_assembly38_masked \
  gs://cpg-tob-wgs-test/cram/nagim
```

vivbak (approved):
This LGTM! I guess just a note that this doesn't seem to deactivate the old analyses. I'm assuming there's already a plan in place for them from a data management perspective, but flagging here in case.

jmarshall:
The plan at the moment would be to deactivate/delete the old analyses manually in conjunction with deleting the superseded CRAM files manually. However I guess it would not take much to expand `do_metamist_update()` to use `query_analyses()` and `update_analysis()` to set all the previously existing CRAM analysis records to `active=False` or similar.

I'll merge this as is, so that we can at least run it on **main** in dry_run mode, and we can decide if that updating should also be automated later.

PR #527, opened 2023-11-19
Use federated identities to auth to GCP

illusional:
Failed after I revoked all the service accounts for gh-images-deployer.
Successful run: https://github.com/populationgenomics/production-pipelines/actions/runs/6922877336/job/18830113488

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -10,6 +10,10 @@ on:
         description: 'Tag to use (defaults to "test")'
         default: "test"
 
+permissions:
+  id-token: write
+  contents: read
+  
 env:
   VERSION: 1.18.8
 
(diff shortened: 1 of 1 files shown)


jmarshall (approved):
Looks good. The `id:` is a bit superfluous, but I guess aids grepping across all our workflows in all our repositories.

Will we be able to delete the GH_IMAGES_DEPLOYER_JSON secret after this? (Unless it's an organisation-wide secret used elsewhere. Or if the token has been revoked anyway, maybe we've already deleted the secret too…)

illusional:
Yeah good call, I'll remove the secret from both repos :) 

> The id: is a bit superfluous
Yeah fair call, it survived the continuous copy-pasting from other repos.

PR #533, opened 2023-11-23
Incorporate sample hash into multiQC input file

### LINKED ISSUE #532: Unexpected Cross-Batch Behavior in Hail Batch: Shared Sequencing Group IDs Between Batches Leading to Job Failures
A Hail Batch ([#1](https://batch.hail.populationgenomics.org.au/batches/430116)) submission of 100 sequencing groups failed at MultiQC [CRAM] and MultiQC [GVCF]. 

The output of the jobs that failed indicate input files don't exist. The sequencing group prefixes of these input files however are not the sequencing groups submitted in the Batch's config file. They are from a secondary batch of 100 ([#2](https://batch.hail.populationgenomics.org.au/batches/430183)) sequencing groups submitted as the initial batch of 100 was nearing completion. 

As batch `#2` was progressing, the missing files were being generated and thus came into existence. Nonetheless, this does not explain why batch `#1` was looking for files related to sequencing groups from a completely separate batch with no overlapping sequencing groups.

MattWellie:
Closes #532

Incorporates the hash of the collected samples into the file name written/read for multiQC. This should permit multiple overlapping batches

### DIFF
--- a/cpg_workflows/stages/joint_genotyping_qc.py
+++ b/cpg_workflows/stages/joint_genotyping_qc.py
@@ -151,19 +151,19 @@ class JointVcfMultiQC(CohortStage):
         """
         if get_config()['workflow'].get('skip_qc', False) is True:
             return {}
-
+        h = get_workflow().output_version
         return {
-            'html': cohort.analysis_dataset.web_prefix() / 'qc' / 'jc' / 'multiqc.html',
+            'html': cohort.analysis_dataset.web_prefix()
+            / 'qc'
+            / 'jc'
+            / h
+            / 'multiqc.html',
             'json': cohort.analysis_dataset.prefix()
             / 'qc'
             / 'jc'
-            / get_workflow().output_version
+            / h
             / 'multiqc_data.json',
-            'checks': cohort.analysis_dataset.prefix()
-            / 'qc'
-            / 'jc'
-            / get_workflow().output_version
-            / '.checks',
+            'checks': cohort.analysis_dataset.prefix() / 'qc' / 'jc' / h / '.checks',
         }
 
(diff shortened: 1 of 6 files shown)


michael-harper:
This looks good to me, nice! 
I just had a question about your Slack message [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1700705642325609?thread_ts=1700696317.357829&cid=C018KFBCR1C). Can you point me to where the files get written in the workflow startup code?

MattWellie:
> Can you point me to where the files get written in the workflow startup code?

I posted an example of that in the related issue - [here](https://github.com/populationgenomics/production-pipelines/issues/532#issuecomment-1823729944). The file is written to a path in the tmp bucket, then the batch's read_input points to that tmp location. The file will only be read when the read_input result is required inside a job

cpg_workflows/stages/joint_genotyping_qc.py
```diff
@@ -151,19 +151,19 @@ def expected_outputs(self, cohort: Cohort) -> dict[str, Path]:
         """
         if get_config()['workflow'].get('skip_qc', False) is True:
             return {}
-
+        h = get_workflow().output_version
```
vivbak, line 154: Is this the best var name?

vivbak (approved):
Awesome! Thanks!

cpg_workflows/stages/joint_genotyping_qc.py
MattWellie: Nah, I really hate this name. Weirdly this was already present as `h`, and was used in the paths of all non-HTML output files. I'll improve it

PR #534, opened 2023-11-23
Remove default dataproc init script

illusional:
The script will do this by default, with support for different hail versions.

Relies on: https://github.com/populationgenomics/analysis-runner/pull/644


MattWellie:
In the linked PR it looks like the `get_init_script_from_hail_version` gets a folder, not a script. Does `--init` just run everything found in that folder?

```return f'gs://cpg-common-main/hail_dataproc/{hail_version}/'```

illusional:
Nice catch, that was definitely supposed to return a file. I'll submit a fix.

MattWellie (approved):

jmarshall:
D'oh saw that on my first pass through but then forgot about it again :-(

PR #535, opened 2023-12-03
Improve (remove) hard path building, assume metamist as source of truth. 

vivbak:
Within production pipelines there is an assumption that all gvcf and cram paths adhere to the following format: 
`self.dataset.prefix() / 'gvcf' / f'{self.id}.g.vcf.gz'` 
`self.dataset.prefix() / 'cram' / f'{self.id}.cram'` 

However, not all paths from all projects will follow this format. This is creating issues ([failing job](https://batch.hail.populationgenomics.org.au/batches/430408/jobs/1)). 

We should rely on metamist as the source of truth for these output paths (when the outputs already exist before the run is kicked off). Though, historically this has been a challenge. 

This PR aims to remove the reliance on 'hardcoded' path checking for crams and gvcfs and instead use sg.cram or sg.gvcf which are defined for each sequencing group.

### DIFF
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -140,21 +140,30 @@ def _populate_analysis(cohort: Cohort) -> None:
         # TODO: Review here
         for sequencing_group in dataset.get_sequencing_groups():
             if (analysis := gvcf_by_sgid.get(sequencing_group.id)) and analysis.output:
-                assert analysis.output == sequencing_group.make_gvcf_path().path, (
+                # assert file exists
+                assert analysis.output.exists(), (
+                    'gvcf file does not exist',
                     analysis.output,
-                    sequencing_group.make_gvcf_path().path,
                 )
-                sequencing_group.gvcf = sequencing_group.make_gvcf_path()
+                sequencing_group.gvcf = analysis.output
             elif sequencing_group.make_gvcf_path().exists():
-                sequencing_group.gvcf = sequencing_group.make_gvcf_path()
-            if (analysis := cram_by_sgid.get(sequencing_group.id)) and analysis.output:
-                assert analysis.output == sequencing_group.make_cram_path().path, (
-                    analysis.output,
-                    sequencing_group.make_cram_path().path,
+                logging.warning(
+                    f'We found a gvcf file in the expected location {sequencing_group.make_gvcf_path()},'
+                    'but it is not logged in metamist. Skipping. You may want to update the metadata and try again. '
                 )
-                sequencing_group.cram = sequencing_group.make_cram_path()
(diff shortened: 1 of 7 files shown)


cpg_workflows/inputs.py
```diff
@@ -135,25 +135,32 @@ def _populate_analysis(cohort: Cohort) -> None:
...
+                # assert file exists
+                assert analysis.output.exists(), (
+                    'cram file does not exist',
                     analysis.output,
-                    sequencing_group.make_cram_path().path,
                 )
-                sequencing_group.cram = sequencing_group.make_cram_path()
+                sequencing_group.cram = analysis.output
+
             elif sequencing_group.make_cram_path().exists():
```
illusional, line 159: Unrelated to this change, we should potentially subclass Path and override the exists method to that we can be more efficient with existence lookups.

cpg_workflows/inputs.py
```diff
@@ -135,25 +135,32 @@ def _populate_analysis(cohort: Cohort) -> None:
...
+                assert analysis.output.exists(), (
+                    'gvcf file does not exist',
                     analysis.output,
-                    sequencing_group.make_gvcf_path().path,
                 )
-                sequencing_group.gvcf = sequencing_group.make_gvcf_path()
+                sequencing_group.gvcf = analysis.output
             elif sequencing_group.make_gvcf_path().exists():
-                sequencing_group.gvcf = sequencing_group.make_gvcf_path()
+                logging.warning(
```
illusional, line 147: So to clarify this behaviour, if metamist doesn't have an entry, and the gvcf path exists, we safely skip.

cpg_workflows/stages/cram_qc.py
```diff
@@ -246,9 +250,10 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
 
 
 def _update_meta(output_path: str) -> dict[str, Any]:
-    from cloudpathlib import CloudPath
     import json
 
+    from cloudpathlib import CloudPath
```
illusional, line 255: Can we add a comment about why we import in the function?

cpg_workflows/inputs.py
vivbak: Yes that's correct. Although, as Matt points down below, we might want to be stricter here and error out?

cpg_workflows/stages/cram_qc.py
vivbak: Happy to open a ticket to add some more comments for this but I feel its outside the scope of this PR :)

MattWellie (approved):
LGTM

jmarshall (commented):
Thanks for the commentary!

PR #536, opened 2023-12-04
Add kinship>0.1 threshold for reporting Somalier ped issues to Slack

EddieLF:
The `row['relatedness']` variable stores the kinship coefficient value between each pair of samples in the Somalier outputs. 

Adding a threshold of `kin>0.1` to the condition which triggers these lines to be added to the report shared in the Slack message will prevent excessive lines that provide little meaningful information. E.g.:

> Sample1 - Sample2, provided: "unrelated", inferred: "related at unknown level", kin=0.007
> Sample3 - Sample4, provided: "unrelated", inferred: "unknown", kin=0.056


cassimons:
Will this still alert if the pair are unrelated but are expected to be related?

cassimons:
Looks good, but I think you might need to look at what falls through to the else statement now :-)

cassimons (approved):
LGTM thanks

PR #537, opened 2023-12-05
Handle existing outputs, for stage re-use

vivbak:
We previously made changes to allow the inputs of stages to defer from a rigid structure #535 

We failed to handle existing outputs in the same way. This is causing problems.
https://batch.hail.populationgenomics.org.au/batches/430489/jobs/1

Further, .crai files don't exist, and these are an expected output of the align stage. 

Also, we need to create CramPaths and GVCFPaths instead of just GSPath objects so we can access things like .index_path.


vivbak:
Looks good, https://batch.hail.populationgenomics.org.au/batches/430506

MattWellie (approved):
LGTMatt

PR #538, opened 2023-12-05
Use cpg utils get batch 2

MattWellie:
I've made a separate branch from @illusional 's <> following a particularly onerous rebase

This is an alternative to #496, and I probably should have tried a merge with main instead of a rebase for simplicity


MattWellie:
Chaotique

PR #542, opened 2023-12-08
Ensure populated CramPaths have both path and index_path

jmarshall:
Followup to PR #537: We need `sequencing_group.cram` to be a CramPath as established by the previous PR, but we also need to ensure that `sequencing_group.cram.index_path` is also set (where applicable?), which requires a suitable value to be provided to the constructor call.

There are surely better ways to query or construct this value. Is it available perhaps via the same metamist call? Is it always available, or do we need to do an `.exists()` and leave this as None if not?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.12
+current_version = 1.18.13
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
   
 env:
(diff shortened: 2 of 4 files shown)


jmarshall:
I haven't tested by hand that the `with_suffix` extension used and utils's `exists()` method do the right thing for us. However `.with_suffix('.cram.crai')` is what is already used elsewhere for these files, and the tests should exercise all this code anyway too.

jmarshall:
I assume there's nothing you need to set up first to use the caching `exists(…)` function…?

illusional:
@jmarshall, nothing needed, just throw the path at it and it should do the rest (the power of decorators)

MattWellie (approved):
I like this change. Can we chuck a version bump in

PR #545, opened 2023-12-12
Fix cpg-utils dependency

MattWellie:
4.18.3 is the version with the fix for multiple configs, require this version when installing

see https://github.com/populationgenomics/cpg-utils/pull/149


vivbak (approved):

illusional (approved):
Not trying to rock the boat, what does the actual pypi release get us, because most people use the docker image right?

PR #546, opened 2023-12-12
Update gnomad_methods submodule

jmarshall:
Update to include this fix: populationgenomics/gnomad_methods#5.


MattWellie (approved):
I've not got much context here, but I unconditionally trust that this is the right change

PR #547, opened 2023-12-12
Dataproc uses latest cpg-utils

MattWellie:
Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1702423297929589?thread_ts=1702345700.581549&cid=C030X7WGFCL


illusional:
We should be able to plainly reissue a build instead of forcing a version, if the regular build doesn't work, then let's merge this.

MattWellie:
Treated elsewhere

PR #549, opened 2023-12-13
Alter max polling interval for Cromwell

MattWellie:
Allows for an extended polling interval for Cromwell watcher jobs
Adds some per-sample jitter to the max interval so that we spread out the API burden

This task is particularly long running (~days) so this could be extended even further


cassimons (approved):
Looks good. I know it will be seconds, but would it make sense to be explicit about the unit either in the variable or doc string - or is this just standard in the code base?

MattWellie:
I've added a comment to clarify why this is happening, and that the value represents seconds

PR #551, opened 2023-12-13
Rare disease RNA pipeline - count reads

mgeaghan:
This stage performs read counting on the alignments produced by the trim and align stage. The steps involved are:

- Attempt to use the temporary BAM files generated by the trim stage in the `-tmp` bucket. If they have been removed, take the CRAM files in `-main` and convert back to bam using `samtools`.
  - The new BAMs will again be placed into the `-tmp` bucket for temporary storage.
- Run `FeatureCounts` from the `subread` package to perform read counting on the BAMs.
- Generates two output files per sample:
  - SAMPLE.count (contains the actual counts for each gene)
  - SAMPLE.count.summary (contains overal counting statistics)


cpg_workflows/jobs/bam_to_cram.py
```diff
@@ -55,3 +55,46 @@ def bam_to_cram(
...
+    )
+
+    j.declare_resource_group(
+        sorted_bam={
+            'bam': '{root}.bam',
+            'bam.bai': '{root}.bam.bai',
+        }
+    )
+
+    cmd = f'samtools view -@ {res.get_nthreads() - 1} -b {input_cram.cram} | tee {j.sorted_bam["bam"]} | samtools index -@ {res.get_nthreads() - 1} - {j.sorted_bam["bam.bai"]}'
```
MattWellie, line 97: this puts the resulting BAM file in _very_ temporary this-batch-only storage (i.e. if a job in the batch fails then this whole inflation needs to be repeated). Would it be worth copying this into the project temp (config.storage.default.temp e.g. `cpg-<whatever>-main-tmp`) and then putting in a loop that will either use or regenerate, depending on whether it already exists?

mgeaghan: Thanks for pointing that out. Turns out, I had done that, just in one of the downstream branches. I'm updating the code now so it's in this branch.

cpg_workflows/jobs/count.py
```diff
@@ -0,0 +1,222 @@
...
+        feature_type='exon',
+        attribute='gene_id',
+        strandness='reverse',  # TODO: confirm this
+        multi_mapping=False,  # TODO: determine default value
+        min_quality=None,  # TODO: determine default value
+        primary_only=True,  # TODO: determine default value
+        ignore_duplicates=False,  # TODO: determine default value
+        count_pairs=True,  # TODO: determine default value
+        both_ends_mapped=True,  # TODO: determine default value
+        both_ends_same_chr=True,  # TODO: determine default value
```
MattWellie, line 206: Are these TODOs all to do, or are they settled?

MattWellie (approved):
Approved, depending on if those TODO statements need to be altered

cpg_workflows/jobs/count.py
mgeaghan: Ah, thanks for pointing that out, I think they're good defaults, but I will also plan to update the code in a future edit to expose these settings through the config toml.

PR #552, opened 2023-12-13
Simplify Align.expected_outputs() CRAM path construction

### LINKED ISSUE #543: Queuing STRipy stage jobs breaks for sequencing groups without Pathlike CRAM paths
The recent merges https://github.com/populationgenomics/production-pipelines/pull/537 and https://github.com/populationgenomics/production-pipelines/pull/542 helped to unify our approach to storing CRAM output paths in `SequencingGroup` objects, as `CramPath` objects. 

The STRipy stage job submissions use `cram_path` and `crai_path` for the input sequencing group  and format them for input using the `CramPath` constructor [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/stripy.py#L76). However, for some some sequencing groups, the CRAM analysis output path used is not a `Pathlike` object, so the pipeline breaks. See the output of [this batch](https://batch.hail.populationgenomics.org.au/batches/430652/jobs/1) for an example of this occurring. I'm not exactly sure why some CRAM analysis outputs are formatted as `Pathlike` entries while others are just strings.

jmarshall:
Reported in #543, and in a similar `Align` inputs context from elsewhere, is the following failure:

```
  File "/production-pipelines/cpg_workflows/stages/stripy.py", line 76, in queue_jobs
    cram_path = inputs.as_path(sequencing_group, Align, 'cram')
  File "/production-pipelines/cpg_workflows/workflow.py", line 309, in as_path
    return res.as_path(key)
  File "/production-pipelines/cpg_workflows/workflow.py", line 178, in as_path
    raise ValueError(f'{res} is not a path object.')
ValueError: gs://cpg-perth-neuro-main/cram/CPG297432.cram is not a path object.
```

Messages like this are usually because `res` is a string rather than some kind of path object. However in this case (as revealed by studying the code or by `mypy`), the problem is that `sequencing_group.cram` is a `CramPath`, which — despite the name! — is not a `pathlib.Path`-like class.

Hence the minimal change to make this work would be:

```diff
-            'cram': sequencing_group.cram      or sequencing_group.make_cram_path().path,
+            'cram': sequencing_group.cram.path or sequencing_group.make_cram_path().path,
```

However in this PR I've instead taken the opportunity to refactor this to treat `cram` and `crai` together and only check whether `sg.cram` is present once, which is hopefully clearer.

We want `.path` and `.index_path` from `sequencing_group.cram` if it has been set; otherwise fall back to the generic `make_cram_path()`. Fixes #543.

### DIFF
--- a/cpg_workflows/stages/align.py
+++ b/cpg_workflows/stages/align.py
@@ -29,16 +29,13 @@ class Align(SequencingGroupStage):
         Stage is expected to generate a CRAM file and a corresponding index.
         """
         if sequencing_group.cram:
-            if sequencing_group.cram.index_path:
-                crai_path = sequencing_group.cram.index_path
-            else:
-                crai_path = None
+            cram_path = sequencing_group.cram
         else:
-            crai_path = sequencing_group.make_cram_path().index_path
+            cram_path = sequencing_group.make_cram_path()
 
         return {
-            'cram': sequencing_group.cram or sequencing_group.make_cram_path().path,
-            'crai': crai_path,
+            'cram': cram_path.path,
+            'crai': cram_path.index_path,
         }
 
     def queue_jobs(


jmarshall:
I'm disappointed that static type checking did not point out this issue when this code was introduced in PR #537. This may be due in part to the fact that (at least on my local machine's installation) mypy does not see the type annotations in _cpg_utils/\__init__.py_, instead just reporting `sequencing_group.make_cram_path().path`'s type as Any.

vivbak (approved):

PR #557, opened 2023-12-15
A couple of suggestions

MattWellie:
- optimising imports (removing unused, rearranging to please the linter)
- changes to use the cpg_utils version of get_batch
- sets job dependencies so that the align jobs don't all start up at once
- one typo
- remove `overwrite` unused argument

### DIFF
--- a/cpg_workflows/jobs/align_rna.py
+++ b/cpg_workflows/jobs/align_rna.py
@@ -186,7 +186,7 @@ def align(
             requested_nthreads=requested_nthreads,
         )
         jobs.append(j)
-        aligned_bams.append(bam.bam)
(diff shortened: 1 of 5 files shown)


mgeaghan:
@MattWellie As I mentioned in a comment in the main PR, the `.depends_on` methods shouldn't be necessary, as the jobs create new resource groups/resource files, which are used in the subsequent jobs, so hail batch should be able to infer the dependencies from those. I've tested the workflow out on the test dataset and it does work, and doesn't start all the jobs at once. If there is a better way to do this I'm happy to make some changes.

MattWellie:
Awesome! That's oversight on my part, I forgot hail was smart enough to do that

PR #561, opened 2023-12-18
gCNV sub-pr 3

MattWellie:
Adds the step for combining all single-sample intervals VCFs into a single joint-call.

- Reads each VCF and corresponding index into the batch temp directory
- Runs a fast bcftools merge on the individual VCFs (these are all called at the same sites/intervals, to this works as a naive merge)
- Inserts a python code block to read the VCF, add 2 header lines for the SVTYPE and SVLEN attributes, then annotate these values into all variant rows. These are both required for StrVCTVRE and AnnotateVcf, but are omitted from the gCNV output

Operational image for this stage is borrowed from GATK-SV (bcftools, tabix, python)

### DIFF
--- a/cpg_workflows/jobs/gcnv.py
+++ b/cpg_workflows/jobs/gcnv.py
@@ -359,3 +359,91 @@ def fix_intervals_vcf(
 
     return reheader_job
 
+
+def merge_calls(
+    b: hb.Batch, sg_vcfs: list[str], docker_image: str, job_attrs: dict[str, str], output_path: Path
+):
+    """
+    This job will run a fast simple merge on per-SGID call files
+    It then throws in a python script to add in two additional header lines
+    and edit the SVLEN and SVTYPE attributes into each row
+
+    Args:
+        b (batch):
+        sg_vcfs (list[str]): paths to all individual VCFs
+        docker_image (str): docker image to use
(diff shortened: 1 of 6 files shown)


cpg_workflows/jobs/gcnv.py
```diff
@@ -359,3 +359,91 @@ def fix_intervals_vcf(
...
+    # option breakdown:
+    # -Oz: bgzip output
+    # -o: output file
+    # --threads: number of threads to use
+    # -m: merge strategy
+    # -0: compression level
+    j.command(
+        f'bcftools merge {" ".join(batch_vcfs)} -Oz -o temp.vcf.bgz --threads 4 -m all -0'
+    )
+    j.command(fr"""
```
cassimons, line 414: nit: Cant say I am in love with inline python.

cassimons (approved):
LGTM

cpg_workflows/jobs/gcnv.py
MattWellie: I really dislike this too. I think there should be a fairly simple way of doing this as a Python job followed by a tabix job, instead of bundling both operations into one Job. Having to parse the VCF and shove in extra header lines etc. it awful at the best of times.

At least I can say that this design is consistent with the GATK-SV codebase 🙃

PR #564, opened 2023-12-18
Bump Scramble part 1 RAM

MattWellie:
Seeing some consistent failures in Scramble, possibly RAM related?

https://github.com/broadinstitute/gatk-sv/blob/main/wdl/Scramble.wdl#L21
-> 
https://github.com/broadinstitute/gatk-sv/blob/main/wdl/Scramble.wdl#L83C15-L83C27
->
https://github.com/broadinstitute/gatk-sv/blob/main/wdl/Scramble.wdl#L98

I _think_ this would override the RAM setting, and everything else would still pull from the default attributes?

### DIFF
--- a/configs/gatk_sv/config_overrides.toml
+++ b/configs/gatk_sv/config_overrides.toml
@@ -6,5 +6,5 @@
 
 # resource overrides for the GatherSampleEvidence stage
 
-[resource_overrides.GatherSampleEvidence.runtime_attr_wham]
+[resource_overrides.GatherSampleEvidence.runtime_attr_scramble_part2]
 mem_gb = 7.5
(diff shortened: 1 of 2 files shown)


configs/gatk_sv/config_overrides.toml
```diff
@@ -6,5 +6,5 @@
 
 # resource overrides for the GatherSampleEvidence stage
 
-[resource_overrides.GatherSampleEvidence.runtime_attr_wham]
+[resource_overrides.GatherSampleEvidence.runtime_attr_scramble_part2]
```
illusional, line 9: Noting that this will cause a rerun of any `wham` jobs as the memory is now different. Fine if we don't use wham, just not sure.

illusional (approved):

configs/gatk_sv/config_overrides.toml
MattWellie: We do use Wham, but I think this override didn't achieve the expected result previously so it's not been used in a while. Fingers crossed this works as expected...

PR #569, opened 2023-12-21
Update default parameters for the rare disease rna seq pipeline

mgeaghan:
Added the STAR reference in cpg-common-main to the default config toml for the rare_disease_rnaseq pipeline. Updated config calls in code to match the new toml section [references.star]


mgeaghan:
@MattWellie Before we merge the outrider pull request I would like to merge this. I had forgot to update the default toml parameters for this pipeline, which I've fixed here. I've also updated the parameters to be under `references.star` rather than just `references` so they can be separated from the regular fasta and gtf. Once we've merged this, I will update the outrider stage to be in line with these changes.

MattWellie (approved):
LGTM

PR #570, opened 2023-12-21
Annotate sex fix

MattWellie:
See https://centrepopgen.slack.com/archives/C030X7WGFCL/p1702345700581549

Test run https://batch.hail.populationgenomics.org.au/batches/431057

We were seeing failures in annotate_sex in the large-cohort pipeline's SampleQC stage. We originally traced these errors to the `gnomad-methods` submodule, but after an extensive debugging operation the point of failure looks to be the filter_intervals operation - this op is not designed for massive numbers of intervals, and was failing repeatedly when we tried to apply a filter table with ~2M rows.

@cassimons redesigned the syntax around this operation, and combining the new filter syntax with checkpointing, we can successfully complete SampleQC, at the expense of creating a few more checkpoints in tmp (these checkpoints may not be necessary given the new syntax/operations, but it's working and I'm too scared to change it now)

Debugging credit to @michael-harper @jmarshall @KatalinaBobowik @illusional

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.18
+current_version = 1.18.19
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
   
 env:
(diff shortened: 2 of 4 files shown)


michael-harper:
Adding debugging credit to @illusional as well!

michael-harper (approved):

PR #576, opened 2023-12-26
Cheeky resource bumps

### LINKED ISSUE #574: Might need more resources by default in GATK-SV
https://github.com/broadinstitute/gatk-sv/blob/dc92e9f4bc46a1ab19459b515c24c9747a73a967/wdl/MedianCov.wdl#L41
```
[resource_overrides.GatherSampleEvidence.runtime_attr_scramble_part2]
mem_gb = 7.5

[resource_overrides.EvidenceQC.wgd_build_runtime_attr]
disk_gb = 40

[resource_overrides.EvidenceQC.runtime_attr_mediancov]
disk_gb = 200
```

MattWellie:
1. Instead of adding a separate code block to each GATK-SV stage ("if this section has config overrides, apply them"), I've moved this to the abstracted "run Cromwell with these instructions" method. This will pick up any config overrides relevant to the WDL name being passed; this should achieve the same effect without any code repetition
2. Adds a few resource overrides which were required during the latest run
3. gCNV resource bump, test run in progress here https://batch.hail.populationgenomics.org.au/batches/431102

Closes #574


cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
```diff
@@ -182,13 +182,16 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
 
+        # this step runs for approximately 15 hours
         jobs = add_gatk_sv_jobs(
             batch=get_batch(),
             dataset=cohort.analysis_dataset,
             wfl_name=self.name,
             input_dict=input_dict,
             expected_out_dict=expected_d,
             labels=billing_labels,
+            cromwell_status_min_poll_interval=500,
```
illusional, line 193: Don't want to make these configurable, or at least overridable?

configs/gatk_sv/config_overrides.toml
```diff
@@ -8,3 +8,9 @@
 
 [resource_overrides.GatherSampleEvidence.runtime_attr_scramble_part2]
 mem_gb = 7.5
+
+[resource_overrides.EvidenceQC.wgd_build_runtime_attr]
```
illusional, line 12: @vivbak, just flagging resource overrides will potentially be a challenge if we move to structured configs.

illusional (approved):
LetsGambleTryMerging

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
MattWellie: That might be better... I've tried to set some reasonable intervals based on the job durations instead of loading up the config with overrides for each job.

I'll make a ticket to T-shirt size the job runtimes, then maybe have a min/max range for `small`, `medium` and `large` jobs

illusional: You could probably roughly estimate based on how many SGs / how big the fastqs are

MattWellie: From poking it in a few ways, the runtime is mostly determined by the number of tasks/sub-workflows in each stage's WDL, and the resource requirements are mainly determined by number of SGs. Both factors will have some sway

PR #577, opened 2023-12-27
Inline python is bad

### LINKED ISSUE #568: Inline python is hideous
https://github.com/populationgenomics/production-pipelines/pull/561/#discussion_r1429409763

This block of inline python should be replaced ASAP

MattWellie:
Closes #568 

Takes the grotty chunk of block-text-python-parsed-as-cli-args and reimagines as a pythonJob

Test run https://batch.hail.populationgenomics.org.au/batches/431133


cassimons (approved):
very nice :-)

Issue #578, opened 2023-12-27
GATK-SV rejects some valid Family IDs


MattWellie:
Error from the log file during ValidatePedFile

```ValueError: Invalid family ID in PED file: 'DXX-YYYY'. IDs should only contain alphanumeric and underscore characters.```

AFAIK the family ID isn't meaningful for GATK-SV, so hot-swap any failing values when generating the PED file

PR #579, opened 2023-12-27
Ped fix 578

MattWellie:
Adds a fix for #578 

GATK-SV includes a PED validation stage, which only permits alphanumeric/underscore characters in Family IDs. Not sure why.

This sandwich method takes each line in the cohort PED file, checks the family ID for conformance (if good, no action). If the ID contains a banned character it runs a regex substitution to replace banned chars with underscores. AFAIK family ID is meaningless in this process, and Sample IDs must match the VCF contents. Altering the family IDs to pass this test shouldn't have any unforeseen consequences. 

If it does have consequences... I have not foreseen them 🙃

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.19
+current_version = 1.18.20
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
   
 env:
-  VERSION: 1.18.19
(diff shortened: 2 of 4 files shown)


illusional (approved):
You can grab a PED with the internal family IDs from metamist, you can guarantee they'll be unique, and if they're already meaningless?

MattWellie:
> You can grab a PED with the internal family IDs from metamist, you can guarantee they'll be unique, and if they're already meaningless?

That's... not a bad idea. The issue here is that the samples can come from any project, and the workflow setup has already dealt with that complexity when gathering SequencingGroup objects. A new metamist query would end up being an arbitrary number of queries per cohort

PR #582, opened 2024-01-02
Outrider side branch

MattWellie:
Mostly just linting and optimising the imports (some were unused, and the order of imports fails a strict linting check)

If `can_reuse` passes, return None instead of an empty list. Both are equivalent arguments to `make_outputs`, but the return type was specified only as Job, so my linter was complaining. Adding `list` as a return type and returning the empty list would be equally valid here


mgeaghan (approved):
Thanks for these changes, I had to update them after the latest changes to the original outrider branch, just to point to the right gtf parameter in the config TOML. I'll merge these changes into the outrider branch.

PR #583, opened 2024-01-04
Fix source of input data for GATK-SV stage

MattWellie:
See: https://batch.hail.populationgenomics.org.au/batches/431434/jobs/1

I'm not really sure how this crept in, this workflow has been run end-to-end previously. 

[This](https://github.com/populationgenomics/production-pipelines/commit/46c325f93f134f691ecf581f46ccb3d629ff77ee) might be the commit that did it - we introduced the FilterGenotypes workflow, then deactivated it due to a broken Docker image, then re-added it - that final re-addition may not have been done comprehensively

Changes to the config files here are done to stay up to date with all batches which have been run

### DIFF
--- a/configs/gatk_sv/all_batch_names.toml
+++ b/configs/gatk_sv/all_batch_names.toml
@@ -2,4 +2,13 @@
 # this will be used when running:
 # - MergeBatchSites (sandwich stage)
 # - MakeCohortVCF (multisample 2)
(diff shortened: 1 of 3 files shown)


EddieLF:
So the input for `AnnotateVcf` will be updated to use the outputs from the required stage `FilterGenotypes` in the `AnnotateVcf` stage decorator, which should fix the DAG so that the output from that stage can be used as input for the annotation stage. If I'm following correctly, looks good!

EddieLF (approved):

PR #589, opened 2024-01-10
gCNV bugfix - make variant IDs unique

MattWellie:
Test batch: https://batch.hail.populationgenomics.org.au/batches/431755

Weird unforeseen issue here - When passing the VCF through SVAnnotate in GATK all the DEL variants were being removed. It seems to be that GATK-SV includes some weird semi-manual VCF parsing/translation to BED which retains only unique variant IDs...

- gCNV calls all variants in the VCF as `CHR POS ID DEL,DUP ...`, i.e. each span is called as DUP and DEL on the same row, sharing a single ID
- This was previously Split, as GATK-SV's AnnotateVcf can't handle multiallelics
- Upon splitting each row was broken into a DEL and a DUP row, with the corresponding genotypes, but the same ID
- Passing this through AnnotateVcf created annotation, but only DUPs were retained
- Working assumption is that buried in the deep dark heart of GATK-SV there's a plaintext manipulation of the variants, made unique on ID, so we only retain the variant type that comes second in the VCF for each span

This change introduces an additional amendment in the VCF; the ID is edited to be `ID_ALT` - these values become unique in the whole dataset, and annotation now completes with both variant types retained


cassimons (approved):

jmarshall:
Not new in this PR, but

```python
 headers.extend([
                        '##INFO=<ID=SVTYPE,Number=1,Type=String,Description="SV Type">\n',
                        '##INFO=<ID=SVLEN,Number=1,Type=Integer,Description="SV Length">\n']
```

SVLEN should be `Number=.` or `Number=A`, which could make a difference to downstream picard/gatk validation.

MattWellie:
> SVLEN should be `Number=.` or `Number=A`, which could make a difference to downstream picard/gatk validation.

cheers, I've prepped that into the next PR

PR #590, opened 2024-01-11
Remove build-during-pr workflow

MattWellie:
This is a hold over from an older workflow, where we would build a new image for every commit into an open PR. 

Now that we no longer do that, remove this section


nevoodoo (approved):

MattWellie:
superseded

PR #591, opened 2024-01-11
Allow main builds to main artifactory

MattWellie:
Replacement for #590 

- A while back we blocked manually-triggered builds from reaching the main artifactory
- When we want to refresh the cpg-workflows image (e.g. deps, but no code change) we don't have a way to do that

This change alters the workflow yaml:

- all builds on `main`, manually triggered or not, are equivalent; build and push to main artifactory with the tags `VERSION` and `latest`
- removes the section designed to build a new image when a new commit is added to an open PR. The workflow no longer runs at this time

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -64,7 +64,7 @@ jobs:
           docker push $DOCKER_MAIN/$IMAGE_NAME:latest
 
       - name: manually triggered build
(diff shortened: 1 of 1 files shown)


illusional (commented):
Looks good to me, can you get a review from @vivbak?

vivbak (approved):

PR #593, opened 2024-01-15
Fraser

mgeaghan:
A stage to detect aberrant splicing events with FRASER.
The stage is broken into several jobs for parallelization, which was based on the method implemented in the [Drop pipeline](https://github.com/gagneurlab/drop):

- If BAMs are not available, run `cram_to_bam`. FRASER cannot count from CRAMs, so BAMs are essential.
- Perform the read counting steps by calling `fraser_count`. This function does the following:
  - Calls `fraser_init`, which launches a job to run a short R script to create the Fraser Data Set (FDS) object required for the counting stages.
  - Calls `fraser_count_split_reads_ones_sample` for every sample in the cohort. One job is created for each sample to count split reads. The counts are saved in an RDS file which can be reused instead of re-counting the sample in the future.
  - Calls `fraser_merge_split_reads`, which is called once for the entire cohort after all split read count jobs have been run (or the existing output has been reused). This merges the counts and generates a list of splice sites which is used by the next stage
  - Calls `fraser_count_non_split_reads_one_sample` for every sample in the cohort. Each job counts the non-split reads in the sample, i.e. reads that span across a splice site. This uses the output list of splice sites from the previous stage. Because of this, all samples need to be recounted, even if the output already exists from a previous run, as the output may change by adding new samples to the cohort.
  - Calls `fraser_merge_non_split_reads` once for the entire cohort after all non-split reads have been counted. This produces a tar file containing the FDS object for the entire cohort. This tarred FDS object is loaded into the final FRASER job.
- Once the counting has been performed, FRASER is run using the tarred FDS object from the counting step. This is extracted within the FRASER job and loaded within R. The FRASER job outputs several tar files containing plots, as well as two CSVs containing the results (one for all results, the other for just significant results).


MattWellie (approved):
There's quite a few unused import in the two new files, and a few instances of `fds` in type hinting as a ResourceGroup when it's actually a ResourceFile causing inspection warnings, would be nice to clean up but nothing fatal. Apologies for taking so long to get around to this

mgeaghan:
I've updated the scripts to remove the unnecessary imports and fix the ResourceGroup/ResourceFile issue. Running a final test now just to be sure nothing has broken.

PR #594, opened 2024-01-15
Split joint call multiallelics

MattWellie:
This change introduces multiallelic VCF splitting into the RD pipeline after JointCalling

Current Flow in JointGenotyping
- Split genome into X intervals, where X is the # samples in the callset
- JointCall samples in each of those intervals separately, generating a VCF fragment
- Obtain a SitesOnly version of the VCF for each fragment
- Join all VCF fragments into a whole-genome VCF (~1TB)
- Join all sites-only VCF fragments into a whole-genome VCF (~7GB)

Current Flow in VEP
- Pass in list of sites-only fragments as argument
- Pass in whole genome sites-only VCF as argument
- Read sites-only whole genome into batch (and ignore it, at least for typical pipeline usage)
- Localise VCF fragments into batch
- Annotate each fragment
- Join results into either a HT or VCF
- If HT, index only on Locus (not alleles)

Applying VEP to VCF
- Read VEP data into HT
- Read whole genome VCF into MT
- Apply VEP annotations to MT using only Locus as key
- Split variants in Hail
- Apply VQSR to MT based on full-key

---

Suggested changes:

Honestly this might need a full design document...

1. We joint call in intervals, then create a sites-only version of each interval. The full VCFs fragments are joined into the full callset, the sites-only fragments are used as input to VEP, and are joined into the input for VQSR.
2. This adds a multiallelic splitting step on each fragment


cpg_workflows/jobs/joint_genotyping.py
```diff
@@ -92,11 +106,12 @@ def make_joint_genotyping_jobs(
         with sample_map_bucket_path.open('w') as fp:
             df.to_csv(fp, index=False, header=False, sep='\t')
 
+    # For small callsets, we don't apply the ExcessHet filtering anyway
     do_filter_excesshet = len(gvcf_by_sgid) >= 1000 and do_filter_excesshet
```
MattWellie, line 110: this line and comment was moved from a few lines down. We did
```do_filter_excesshet = len(gvcf_by_sgid) >= 1000 and do_filter_excesshet``` 
here, then later did a boolean test for 
```if len(gvcf_by_sgid) >= 1000 and do_filter_excesshet:```
which feels silly

Issue #598, opened 2024-01-18
GATK-SV samples go missing during the pipeline


MattWellie:
This only becomes evident during AnnotateDataset, when the pipeline fails because expected samples are not present in the joint-call.

Relevant batch here
https://batch.hail.populationgenomics.org.au/batches/432020

MattWellie:
This might be caused by a filtering step in the workflow that removes individual samples when generating the aggregate. I haven't tracked that down specifically, but it's the only thing that makes sense

MattWellie:
This might be a relevant object to inspect
gs://cpg-seqr-main/gatk_sv/9150271ccec913f7270b318cfd289d4d76bb69_1589/MakeCohortVcf/cleaned_SV_VCF_QC_output.tar.gz

PR #600, opened 2024-01-18
Allow workflow outputs to be written to `-analysis`

MattWellie:
Some pipeline outputs would benefit from manual review, but the Stage/dataset/cohort prefix structure doesn't currently support addresses in the analysis buckets.

This would be used for data intended for manual review, not for the majority of data generated by analysis pipelines. Issue #598 is an example of where the pipeline writes some QC objects, but they are on `-main` with the rest of the data which is a barrier to manual inspection, making this a tricky solve

### DIFF
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -422,6 +422,10 @@ class Stage(Generic[TargetT], ABC):
     def prefix(self) -> Path:
         return get_workflow().prefix / self.name
 
+    @property
+    def analysis_prefix(self) -> Path:
+        return get_workflow().analysis_prefix / self.name
+
     def __str__(self):
         res = f'{self._name}'
         if self.skipped:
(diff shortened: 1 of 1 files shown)


vivbak:
Curious r.e. the processes for deciding what can and can't be written into -analysis, and how we make these decisions, and protect against mistakes.

MattWellie:
@vivbak we don't have any formal guidance in place, and I think at times this can be quite subjective. This particular change would enable us to make that choice in future, identifying that the pipeline codebase doesn't have any notion of analysis as an output option at the moment.

The change I would like to make following this would be altering the output location of the aggregated QC stats `.tar` file from GATK-SV to analysis (it's not an HTML so web is inappropriate, and it contains a zipped file of aggregated plots so it is intended for manual review)

vivbak (approved):

PR #601, opened 2024-01-18
Fix Y Ploidy Inflation in SampleQC Stage: Allow variants_only_x_ploidy and variants_only_y_ploidy to be set in config file

michael-harper:
There was an observed imbalance in the number of inferred "X" and "XY" samples in the TenK10K Phase 1 cohort. A closer investigation revealed that Y ploidy values were inflated, ranging from 2-6 in both "X" and "XY" samples, contrary to the expected range of 0-1.

**Possible fix:**
The proposed fix involves setting `variants_only_x_ploidy` and `variants_only_y_ploidy` in the `annotate_sex()` function of the SampleQC stage to the default value of False. Currently, both parameters are set to True in the pipeline. This change ensures that reference block depths, rather than solely allele depths at variant sites, are used for calculating coverage across sex chromosomes and subsequently for sex imputation.

We have tested this fix in the test VDS, and it produces Y ploidy values of ~0 in "XX" samples and ~1 in "X" (presumably "XY") samples, aligning with expectations. The default value for these parameters in gnomad v3's SampleQC is also set to False.

This change would allow the user to set the value for `variants_only_x_ploidy` and `variants_only_y_ploidy` in the large_cohort.toml file using `inf_ploidy_using_var`

>https://centrepopgen.slack.com/archives/C018KFBCR1C/p1705539231990829?thread_ts=1704233101.883849&cid=C018KFBCR1C

### DIFF
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -53,6 +53,12 @@ pca_plot_name = 'hgdp_1kg_sites'
 # set whether PCA output labels are from given population names, or random forest
 # inferred population labels
 inferred_ancestry = false
+# Whether or not to infer ploidy using the variant data:
+# Using variant sites only for ploidy calculations causes ChrY ploidy inflation.
+# Setting `variants_only_x_ploidy` and `variants_only_y_ploidy` to `False` (default value) causes
+# function to use reference blocks for ploidy calculations
+# https://centrepopgen.slack.com/archives/C018KFBCR1C/p1705539231990829?thread_ts=1704233101.883849&cid=C018KFBCR1C
+inf_ploidy_using_var = true
 
 # Parameters for sample QC. The default values for soft filtering taken from gnomAD QC:
 # https://macarthurlab.org/2019/10/16/gnomad-v3-0, with `max_n_snps` and 
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -58,7 +58,7 @@ inferred_ancestry = false
 # Setting `variants_only_x_ploidy` and `variants_only_y_ploidy` to `False` (default value) causes
 # function to use reference blocks for ploidy calculations
 # https://centrepopgen.slack.com/archives/C018KFBCR1C/p1705539231990829?thread_ts=1704233101.883849&cid=C018KFBCR1C
-inf_ploidy_using_var = true
+inf_ploidy_using_var = false
 
 # Parameters for sample QC. The default values for soft filtering taken from gnomAD QC:
 # https://macarthurlab.org/2019/10/16/gnomad-v3-0, with `max_n_snps` and 
--- a/cpg_workflows/large_cohort/relatedness.py
+++ b/cpg_workflows/large_cohort/relatedness.py
@@ -139,7 +139,9 @@ def _compute_sample_rankings(ht: hl.Table) -> hl.Table:
     @return: table ordered by rank, with the following row fields:
         `rank`, `filtered`
     """
-    use_only_variants = get_config()['large_cohort'].get('inf_ploidy_using_var', False)
+    use_only_variants = get_config()['large_cohort']['pca_background'].get(
+        'inf_ploidy_using_var', False
+    )
     if use_only_variants:
         ranking_column = 'var_data_chr20_mean_dp'
     else:
--- a/cpg_workflows/large_cohort/relatedness.py
+++ b/cpg_workflows/large_cohort/relatedness.py
@@ -139,10 +139,7 @@ def _compute_sample_rankings(ht: hl.Table) -> hl.Table:
     @return: table ordered by rank, with the following row fields:
         `rank`, `filtered`
     """
-    use_only_variants = get_config()['large_cohort']['pca_background'].get(
-        'inf_ploidy_using_var', False
-    )
-    if use_only_variants:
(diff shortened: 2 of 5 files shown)


cpg_workflows/large_cohort/sample_qc.py
```diff
@@ -121,14 +122,15 @@ def impute_sex(
             logging.info(f'count post {name} filter:{vds.variant_data.count()}')
 
     # Infer sex (adds row fields: is_female, var_data_chr20_mean_dp, sex_karyotype)
+    inf_ploidy_using_var = get_config()['large_cohort'].get('inf_ploidy_using_var')
```
illusional, line 125: This will return `None` if the config value isn't specified - do we want a default False here, or to raise an exception if the value isn't supplied?

configs/defaults/large_cohort.toml
```diff
@@ -53,6 +53,12 @@ pca_plot_name = 'hgdp_1kg_sites'
 # set whether PCA output labels are from given population names, or random forest
 # inferred population labels
 inferred_ancestry = false
+# Whether or not to infer ploidy using the variant data:
+# Using variant sites only for ploidy calculations causes ChrY ploidy inflation.
+# Setting `variants_only_x_ploidy` and `variants_only_y_ploidy` to `False` (default value) causes
+# function to use reference blocks for ploidy calculations
+# https://centrepopgen.slack.com/archives/C018KFBCR1C/p1705539231990829?thread_ts=1704233101.883849&cid=C018KFBCR1C
+inf_ploidy_using_var = true
```
hopedisastro, line 61: shouldn't this be False? ie we `variant_only...` to be set to False.

cpg_workflows/large_cohort/sample_qc.py
michael-harper: I think the general consensus in the thread was to have the default value of `False` though this is still dependent on how setting the variables to false effects results across whole cohorts

configs/defaults/large_cohort.toml
michael-harper: Correct

cpg_workflows/large_cohort/relatedness.py
```diff
@@ -125,17 +132,25 @@ def flag_related(
...
-    the better.
+    the better. The ranking is based on either 'var_data_chr20_mean_dp' or
+    'autosomal_mean_dp' depending on the 'inf_ploidy_using_var' configuration.
 
-    @param ht: table with a `var_data_chr20_mean_dp` and `filters` fields.
+    @param ht: table with either `var_data_chr20_mean_dp` or `autosomal_mean_dp` and `filters` fields.
     @return: table ordered by rank, with the following row fields:
         `rank`, `filtered`
     """
+    use_only_variants = get_config()['large_cohort']['pca_background'].get(
```
jmarshall, line 142: `_compute_sample_rankings` may be called from other contexts, now or in the future.

I would rather determine which column name to use by introspecting `ht`. Either by checking the global annotations via something like `hl.eval(ht[variants_only_x_ploidy])` (and checking `variants_only_y_ploidy` too) or perhaps by checking whether a `var_data_chr20_mean_dp` column exists. The former is probably better.

michael-harper: I've changed it to the following in the latest commit: 
```python
if 'var_data_chr20_mean_dp' in ht.row:
        ranking_column = 'var_data_chr20_mean_dp'
    else:
        ranking_column = 'autosomal_mean_dp'
```

If also occurs in `add_soft_filters` inside `sample_qc.py`:
```python
if 'var_data_chr20_mean_dp' in ht.row:
        autosomal_coverage_colname = 'var_data_chr20_mean_dp'
    else:
        autosomal_coverage_colname = 'autosomal_mean_dp'
    ht = add_filter(
        ht,
        ht[autosomal_coverage_colname] < cutoffs['min_coverage'],
        'low_coverage',
    )
```

cpg_workflows/large_cohort/relatedness.py
```diff
@@ -83,9 +83,16 @@ def flag_related(
 ) -> hl.Table:
     """
     Rank samples and flag samples to drop so there is only one sample per family
-    left, with the highest rank in the family.
-
-    `sample_qc_ht` has to have a `filters` and `var_data_chr20_mean_dp` columns.
+    left, with the highest rank in the family. The ranking is based on either
+    'var_data_chr20_mean_dp' or 'autosomal_mean_dp' depending on the
```
jmarshall, line 87: Nice that the introspection is as tidy as `'foo' in ht.row`! Probably should update this comment to say “depending on which is present in the table” or so.

michael-harper: Good call, I've changed it to the following:
```test
Rank samples and flag samples to drop so there is only one sample per family
left, with the highest rank in the family. The ranking is based on either
'var_data_chr20_mean_dp' or 'autosomal_mean_dp' depending on which is present
in the `sample_qc_ht` table. The presence of either column is determined by
the 'inf_ploidy_using_var' parameter in the config toml.
```

hopedisastro:
Hey @jmarshall is this OK to merge into `main` now?

jmarshall (approved):
Yes, from the code point of view I think this is fine to merge into **main** now.

cassimons (approved):
Looks good to go from my POV.

PR #602, opened 2024-01-19
586 Multiallelic splitting

### LINKED ISSUE #586: Split Multiallelic Variants in main pipeline
Our main pipeline calls variants in a multiallelic way, then retains that multiallelic representation through to the end of the workflow, instead of splitting. Partially because of this we don't do any normalisation of variant representations.

- Joint Called VCF is multiallelic
- VQSR runs on this multiallelic representation and includes splitting variants as part of the standard workflow. When VQSR results are annotated onto the original data, this is done after splitting the callset.
- VEP runs on the multiallelic representation
- We read the multiallelic VCF into a MatrixTable, then join VEP annotations using locus as the only key (instead of locus, alt which would be typical). Only then is the data split into atomic alt alleles, and VQSR annotations are added

This is not working great in VEP - we have a recurrent failure to annotate a given fragment of the VCF. This also occurred on the Broad dataset, which prompted an upgrade to VEP 110 (solved the issue for a while, but we're seeing it again now).

This annotation problem goes away if the VCF is split into atomic variant representations. I'm yet to make a bug ticket for Ensembl with the failing VCF(s), but I attempted that previously and they didn't have a solution... just asked us to update our annotation sources, and we're already up to date.

# Q?

Is there a reason not to split our joint-calling VCF immediately after calling, and proceed with atomic variants throughout the pipeline?

## TODO 

- ~matt to gather various code sections for where this splitting would occur~
- ~matt to gather code sections where we currently rely on multiallelic variant representations when joining data from the 3 sources (VCF, VEP, VQSR)~
- discuss with team - is there a benefit to our current process I'm not seeing? Is there a reason not to change?

see https://batch.hail.populationgenomics.org.au/batches/431187/jobs/1

## Codey bits

Needs to be a compound (locus, alt) key [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/vep.py#L340)

Join [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/seqr_loader.py#L49) on full key, not just locus 

Split [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/seqr_loader.py#L51) needs to be removed, unsure if locus_old and alleles_old are relevant if splitting doesn't occur...

Looks like VQSR table is already keyed correctly, and annotation is applied post-splitting [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/seqr_loader.py#L66). My earlier criticism of VQSR is invalid...

An additional stage should follow JointGenotyping, or a splitting job should be included in the JointGenotyping stage [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/joint_genotyping.py#L24)

MattWellie:
Closes #586 

Proposal [here](https://docs.google.com/document/d/1kMEopvFpi6ZdP1QFWanIQ7vJu7P0D0xn_49_VFv80Io)

The relevant changes match the proposal in the linked document:

* during joint genotyping-per-interval, create a sites-only version of each fragment (for use in VQSR in fragments, and also as a whole VCF)
* also generate a version of each VCF with multiallelics split (to be joined, becoming the final callset VCF)
* ALSO generate a sites-only version of each split fragment (for use in VEP)


cpg_workflows/jobs/joint_genotyping.py
```diff
@@ -166,6 +191,36 @@ def make_joint_genotyping_jobs(
...
+            siteonly_j, siteonly_j_vcf = add_make_sitesonly_job(
+                b=b,
+                input_vcf=split_fragment,
+                output_vcf_path=out_split_sitesonly_vcf_part_paths[idx],
+                job_attrs=(job_attrs or {}) | dict(part=f'{idx + 1}/{scatter_count}'),
+            )
+
+            # depend on the previous job
+            if siteonly_j:
+                siteonly_j.depends_on(jobs[-1])
```
cassimons, line 221: nit: I am not sure how I feel about the use of jobs[-1] in this context.

cpg_workflows/jobs/joint_genotyping.py
```diff
@@ -509,3 +564,50 @@ def add_make_sitesonly_job(
...
+    # this is only for a single fragment, so expect it to be small
+    j.storage(get_config()['workflow'].get('splitting_gb', '10Gi'))
+
+    j.declare_resource_group(
+        output_vcf={'vcf.gz': '{root}.vcf.gz', 'vcf.gz.tbi': '{root}.vcf.gz.tbi'}
+    )
+
+    # bcftools norm splits multiallelic variants
+    # -m -any: split all multiallelic variant types
+    # -N: don't left-align indels (NO Normalisation)
```
cassimons, line 598: I would be keen to discuss this at some point. I know it would require some further rejig of our pipelines, but it still feels like normalisation would make sense for us.

cpg_workflows/query_modules/seqr_loader.py
```diff
@@ -88,14 +83,9 @@ def annotate_cohort(
 
     logging.info('Annotating with clinvar and munging annotation fields')
     mt = mt.annotate_rows(
-        AC=mt.info.AC[mt.a_index - 1],
-        AF=mt.info.AF[mt.a_index - 1],
+        AC=mt.info.AC[0],  # TODO not sure about these two
```
cassimons, line 86: Do these need to go up into where the split happens now?

MattWellie: I need to check the behaviour here, the splitting should mean these are either single numbers, or lists of length 1

cpg_workflows/jobs/joint_genotyping.py
MattWellie: It's safer to depend on all prior jobs, and Hail can do much bigger dependency graphs than this. Will update

cpg_workflows/stages/joint_genotyping.py
```diff
@@ -67,17 +71,27 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
         vcf_path = self.expected_outputs(cohort)['vcf']
         siteonly_vcf_path = self.expected_outputs(cohort)['siteonly']
         scatter_count = joint_calling_scatter_count(len(cohort.get_sequencing_groups()))
+        # vcf framents, stripped of genotypes
```
EddieLF, line 74:
```suggestion
        # vcf fragments, stripped of genotypes
```

cpg_workflows/stages/joint_genotyping.py
```diff
@@ -67,17 +71,27 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+            to_path(
+                self.expected_outputs(cohort)['split_part_pattern'].format(idx=idx)
+            )
+            for idx in range(scatter_count)
+        ]
 
         jc_jobs = joint_genotyping.make_joint_genotyping_jobs(
             b=get_batch(),
             out_vcf_path=vcf_path,
+            out_split_sitesonly_vcf_part_paths=out_split_vcf_part_paths,
```
EddieLF, line 92: In the `make_joint_genotyping_jobs` changes in the `jobs/joint_genotyping`, two new arguments are added:
```
out_split_vcf_part_paths: list[Path] | None = None,
out_split_sitesonly_vcf_part_paths: list[Path] | None = None,
```

But only the latter is passed in here, does this check out? Is it because we want to trigger the `# if requested, make a site-only VCF for this split fragment (for VEP)` condition? And that we would only have a `out_split_vcf_part_paths` argument to pass in if the split vcf had already been made?

MattWellie: Sorry, this is a bit convoluted. I confused myself a bit.

This process makes all the fragments separately and joins them up, with the code offering to save the end product, the fragments, or both. The files that we want at the end of this process are:

1. The whole VCF with split multiallelics and genotypes (for combining annotations -> Seqr)
2. The whole VCF without splitting, with no genotypes for VQSR
3. The fragments which are both split and no genotypes for VEP

All these VCF fragments are generated, presence in this list just determines which are generated in tmp, and which are persisted to GCP. I think it's correct to:

1. Not persist unsplit sites-only fragments (VQSR re-splits the whole sites-only VCF based on different logic. I think doing that is dumb, but the topic for another PR)
2. Persist split VCF with genotypes for annotateCohort later on
3. Persist the split & sites-only VCF fragments for feeding into VEP

MattWellie: I've made a bunch of corrections to bring it in line with this intent ^ which it was definitely not doing before

MattWellie:
Test batches keep failing when re-aggregating all the VEP data
https://batch.hail.populationgenomics.org.au/batches/432509

MattWellie:
Latest batch with the highmem worker patch has succeeded - requires evaluation
https://batch.hail.populationgenomics.org.au/batches/433214?q=&last_job_id=50

MattWellie:
Closing as irrelevant in an RD_Combiner world. Open to re-evaluating later

cpg_workflows/query_modules/vep.py
```diff
@@ -333,9 +333,11 @@ def vep_json_to_ht(vep_result_paths: list[str], out_path, vep_version: str):
...
-    chrom = ht.vep.seq_region_name
-    ht = ht.annotate(locus=hl.locus(chrom, start))
-    ht = ht.key_by(ht.locus)
+    ht = ht.annotate(
+        locus=hl.locus(
+            ht.vep.seq_region_name, hl.parse_int(ht.vep.input.split('\t')[1])
+        ),
+        alleles=ht.vep.allele_string.split('/'),
+    )
+    ht = ht.key_by(ht.locus, ht.alleles)
```
MattWellie, line 342: This has come back to haunt me

Issue #500, opened 2023-11-02
Bugs in memory specification for Java jobs


jmarshall:
The way we specify memory requirements for Java jobs needs an overhaul:

* `get_java_mem_mb()` mixes up decimal GB and binary MiB in its arithmetic, leading to intermittently allocating a larger Java heap than hail has been specified to allow;
* Memory overrides that act via `j.memory()` do not update the `resource` hence do not influence the `‑Xms`/etc settings produced by `get_java_mem_mb()`;
* Java commands (variously invoked via `java …`, `bazam …`, `picard …`, and `gatk …`) use a mixture of `‑Xms` (starting heap size) and `‑Xmx` (maximum heap size).

Slack thread for context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1698788154906359

jmarshall:
We applied PR #522 back in mid November to fix the arithmetic mix up. PR #603 adds new methods to overhaul the remaining infelicities.

jmarshall:
Fixed now that PR #603 has been merged.

PR #603, opened 2024-01-21
Higher-level abstraction to replace `JobResource.get_java_mem_mb()`

jmarshall:
This addresses (most of) #500 by introducing `java_mem_options()` to encapsulate both `-Xms` and `-Xmx` options, and also `java_gc_thread_options()` to simplify another repeated `-XX` pattern.

We could add `get_…` to the names of these methods, but unlike the other getters that return an individual property, these produce a complete option string — so they can be thought of as being conceptually different.

The only real behaviour change here is in the `markdup()` memory override, in which when you set `resource_overrides.picard_mem_gb` the Java heap now actually takes advantage of all the memory you've given it.

There are still several snippets that hard-code various lowish memory values for `-Xmx` etc which could be brought into the fold.

### DIFF
--- a/cpg_workflows/jobs/gcnv.py
+++ b/cpg_workflows/jobs/gcnv.py
@@ -104,10 +104,9 @@ def collect_read_counts(
     # set highmem resources for this job
     job_res = HIGHMEM.request_resources(ncpu=2, storage_gb=10)
     job_res.set_to_job(j)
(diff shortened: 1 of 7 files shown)


cassimons (approved):
Looks good to me.

vivbak (approved):

PR #604, opened 2024-01-23
Remove 'v' Prefix from VDS Version in Large Cohort Pipeline for Better Output Version Handling

michael-harper:
This Pull Request (PR) addresses an issue in the large_cohort pipeline where specifying an `output_version` in the configuration file doesn't allow for both reusing a previous VDS version and specifying a new `sample_qc_ht` version prefix.

The large_cohort pipeline has a parameter to specify the VDS version, but it prepends a 'v' to this version. This becomes a problem when the previous VDS was specified by `output_version`, because a VDS with a name like `v1-0.vds` won't exist.

To solve this issue, the PR removes the prepending of 'v' to the `vds_version`. This allows for the correct referencing of VDS files when they are named using the `output_version` scheme, and enables the generation of a new `sample_qc_ht` version while reusing a specified `output_version` VDS.

This change enhances the flexibility of the large_cohort pipeline, making it possible to compare new `SampleQC` results with a previous `sample_qc_ht `output, while also using the corresponding VDS from the previous run.

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -18,8 +18,6 @@ class Combiner(CohortStage):
     def expected_outputs(self, cohort: Cohort) -> Path:
         if vds_version := get_config()['workflow'].get('vds_version'):
             vds_version = slugify(vds_version)
-            if not vds_version.startswith('v'):
-                vds_version = f'v{vds_version}'
 
         vds_version = vds_version or get_workflow().output_version
         return cohort.analysis_dataset.prefix() / 'vds' / f'{vds_version}.vds'


hopedisastro:
LGTM

vivbak (commented):
@michael-harper do you have links to the batch where you encountered this failure

hopedisastro:
@vivbak see here: https://batch.hail.populationgenomics.org.au/batches/432286/jobs/1

vivbak (commented):
Ah looks like when we discussed this last, we made the decision to add the `v` prefix in front of everything. Although that doesn't look like it was actioned (and it looks like I am responsible for that one yikes). 
https://centrepopgen.slack.com/archives/C018KFBCR1C/p1678334879933399 

I'd be keen to quickly make sure analysts are all good with scrapping this `v` before we go ahead as the last discussion we had was to keep the v.

illusional:
Dismissing my review, I think you've got better people to provide better feedback.

vivbak:
I'd also like to suggest as an alternative if we don't have support to scrap the v:
- Ensure vds_version is specified on future runs. We can enforce this. 
- Rename the .vds to match the desired naming convention instead (add a v). This will also unblock you.

hopedisastro:
@vivbak for the second option: renaming the VDS, which repository should we add the shell script to?

vivbak:
@hopedisastro I don't have a great answer for this, and worth asking in a data standup. But I feel the best option is here https://github.com/populationgenomics/transfer-private/tree/main/bioheart

michael-harper:
@vivbak I'm in favour of scrapping the 'v' and using `vds_version`. If I'm understanding correctly, it provides greater flexibility to re-analysis like in this case where we want to use previous output for a different stage where we will produce a new version of that stages output. This way, when running data for the first time, one can use just `output_version` and produce outputs with the specified output version prefix. Then subsequent analysis has the flexibility to take previous outputs and produce new analyses.

amyminiter:
@vivbak We are re-examining this PR in the context of https://github.com/populationgenomics/transfer-private/pull/56 (tl;dr: rename folders in bioheart buckets to disambiguate; delete files that were produced in debugging processes). 

This PR from January removes code that reinserts a v into the file path. Can you please re-examine this PR and advise if it is necessary/desirable to remove this segment of code in the Combiner stage? Thanks! ✋

amyminiter:
@vivbak and to confuse us even more, documentation for the LC pipeline states that the v should in fact be present 

image
 
Link to LC pipeline [Miro board](https://miro.com/app/board/uXjVPPN9TaM=/?share_link_id=822927026069).

amyminiter:
@vivbak

vivbak:
> Can you please re-examine this PR and advise if it is necessary/desirable to remove this segment of code in the Combiner stage?

@amyminiter, it just depends on whether the PA team decides to scrap the V or not. As per the thread I linked above, the last decision on the matter that I was involved in, was to keep the v. So if that's the call, we close this PR. 

It's important to note, the Miro board was a reflection of the code at the time, not the other way around, so it IMO it doesn't really matter what that documentation says, so long as its updated accordingly.

MattWellie:
@michael-harper I think this change has been superseded given @Alexander-Stuckey's refactor of the combiner stage - now this VDS version is exclusively taken from config, with no editing

PR #607, opened 2024-01-29
Precompute coverage mt

MattWellie:
When we call annotate_sex, we do it without a `coverage_mt`. This leads to a behaviour where the gnomad_methods will go off and generate the pre-computed coverage MT, then call the version of the method which requires that input [here](https://github.com/hail-is/hail/blob/5fde6f2a8cf0d520769efd6f9ea2db10a033f976/hail/python/hail/vds/methods.py#L459).

Effectively this is an attempt to bridge the gap between the method calls, generating that MT using the same code as is present in [the naive impute_sex_chromosome_ploidy](https://github.com/hail-is/hail/blob/5fde6f2a8cf0d520769efd6f9ea2db10a033f976/hail/python/hail/vds/methods.py#L531-L631) method, and checkpointing the results in a recoverable way.

The method [here](https://github.com/hail-is/hail/blob/5fde6f2a8cf0d520769efd6f9ea2db10a033f976/hail/python/hail/vds/methods.py#L595) does checkpoint partway through into Batch temp space, but then carries out a bunch of operations on the returned object without checkpointing. This implementation explicitly checkpoints the final object prior to returning, which may help with memory issues we're seeing in dataproc. Not sure if this change is sufficient.

### DIFF
--- a/cpg_workflows/large_cohort/sample_qc.py
+++ b/cpg_workflows/large_cohort/sample_qc.py
@@ -21,8 +21,8 @@ def run(
     out_sample_qc_ht_path: Path,
     tmp_prefix: Path,
 ):
-    if can_reuse(out_sample_qc_ht_path, overwrite=True):
-        return hl.read_table(str(out_sample_qc_ht_path))
+    # if can_reuse(out_sample_qc_ht_path, overwrite=True):
+    #     return hl.read_table(str(out_sample_qc_ht_path))
 
     ht = initialise_sample_table()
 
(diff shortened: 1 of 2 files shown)


cpg_workflows/large_cohort/sample_qc.py
```diff
@@ -82,6 +83,86 @@ def initialise_sample_table() -> hl.Table:
...
+                         f'table with single key of type interval<locus>, found table with key: {key_dtype}')
+
+    rg = vds.reference_data.locus.dtype.reference_genome
+    par_boundaries = []
+    for par_interval in rg.par:
+        par_boundaries.append(par_interval.start)
+        par_boundaries.append(par_interval.end)
+
+    # segment on PAR interval boundaries
+    calling_intervals = hl.segment_intervals(calling_intervals, par_boundaries)
```
hopedisastro, line 117: I think PAR regions are already excluded from the calling_intervals ht?

MattWellie: If you're confident, then we can remove that section. If there's any chance PAR. regions remain this should be a free operation. I was just copy pasting 😅

hopedisastro:
The code runs in a stand-alone script with `test` data: https://batch.hail.populationgenomics.org.au/batches/432418 :)

michael-harper (approved):
Outside of the clarification on PAR regions it looks good. Bringing forward the calculation of coverage of variants and checkpointing prior to `annotate_sex` hopefully removes the OOM issues

vivbak (changes requested):
Quick pause since it's approved but we will not be merging these changes at present.

MattWellie:
Provisionally closing this - @hopedisastro has run this process with and without the additional code, and it may be a dataproc issue rather than a pre-computation issue? Or there may be some other confounding issue. Either way, keeping this branch alive, but the PR closed

PR #609, opened 2024-01-31
Remove dataproc from SampleQC stage

michael-harper:
Similar changes to [#PR514](https://github.com/populationgenomics/production-pipelines/pull/514).

Dataproc was proving difficult to debug and recent efforts to debug demonstrated an ability to perform `SampleQC` stage functions without the need for dataproc see [here](https://github.com/populationgenomics/sv-workflows/blob/main/str/helper/sex_ploidy/sex_inference_stand_alone.py) and relevant Slack thread [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1706563905131719?thread_ts=1704233101.883849&cid=C018KFBCR1C).


cpg_workflows/large_cohort/sample_qc.py
```diff
@@ -8,20 +8,22 @@
...
 def run(
-    vds_path: Path,
-    out_sample_qc_ht_path: Path,
-    tmp_prefix: Path,
+    vds_path,
+    out_sample_qc_ht_path,
+    tmp_prefix,
 ):
     if can_reuse(out_sample_qc_ht_path, overwrite=True):
-        return hl.read_table(str(out_sample_qc_ht_path))
```
MattWellie, line 24: So... this condition was never triggered (can_reuse with overwrite=True always returns False), but can we check why this would return a Table object? That won't be possible if this is run as a job, but also wasn't possible AFAIK if this is run in Dataproc..

cpg_workflows/large_cohort/sample_qc.py
```diff
@@ -8,20 +8,22 @@
...
 
 
 def run(
-    vds_path: Path,
-    out_sample_qc_ht_path: Path,
-    tmp_prefix: Path,
+    vds_path,
+    out_sample_qc_ht_path,
+    tmp_prefix,
 ):
```
MattWellie, line 24: can you workshop this a bit? The type hinting here should identify these are `str` types, which means that you can remove a couple of `str(X)` transformations of these variables below, but `tmp_prefix` needs to be run through `to_path()` for this kind of syntax to work: `sqc_ht_path = tmp_prefix / 'sample_qc.ht'`

cpg_workflows/large_cohort/sample_qc.py
michael-harper: It looks like all of `sample_qc` was committed as one file (see PR [here](https://github.com/populationgenomics/production-pipelines/pull/136)) so I can't see why the decision to return a Table object was made. Also possibly the last change to functionality of `can_reuse` in this script is by Kat [here](https://github.com/populationgenomics/production-pipelines/pull/342)

cpg_workflows/stages/large_cohort.py
```diff
@@ -1,16 +1,19 @@
...
-    stage,
     CohortStage,
     StageInput,
     StageOutput,
     get_workflow,
+    stage,
 )
-from .genotype import Genotype
+
 from .. import get_batch
```
MattWellie, line 15: 🤔 technically this is a possible import, but it's not an input from source. This should be an import from `cpg_utils.hail_batch`

cpg_workflows/stages/large_cohort.py
```diff
@@ -52,18 +55,21 @@ def expected_outputs(self, cohort: Cohort) -> Path:
...
-            depends_on=inputs.get_jobs(cohort),
+        j = get_batch.new_job(
+            'Sample QC', (self.get_job_attrs() or {}) | {'tool': 'hail query'}
+        )
+        j.image(image_path('cpg_workflows'))
+        j.command(
+            query_command(
+                sample_qc,
+                sample_qc.run.__name__,
+                str(inputs.as_path(cohort, Combiner)),
```
MattWellie, line 68:
```suggestion
                inputs.as_str(cohort, Combiner),
```

Should work as an alternative here

MattWellie (approved):
One dodgy import, but it looks good. Would be good to have some test results to prove it works, I'm optimistic

MattWellie:
@vivbak this has now been run with and without dataproc in test, which uncovered something interesting... Points of comparison:

1. an old SampleQC result, using dataproc (`gs://cpg-bioheart-test/large_cohort/3-0/sample_qc.ht`)
2. a new SampleQC result, not using dataproc (`gs://cpg-bioheart-test/large_cohort/no_dp_sampleqc/sample_qc.ht`)
3. a new SampleQC result, using dataproc again (`gs://cpg-bioheart-test/large_cohort/back_in_dataproc/sample_qc.ht`, main branch)

1 & 2 were **not** the same here; the reason was sexes assigned to participants (all `0` in **1.**, TOB-WGS samples were populated in **2.**). I assumed that this stemmed from the sex updates applied yesterday, but they were only applied to main-project data. I re-ran SampleQC on the same input, but from the main branch, and got an identical result between **2.** & **3.**.

I think I've tracked the reason to [this](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/inputs.py#L181) - when running in test mode (access level) we still pull pedigree data from the main project. This would require an amendment like the ones in `metamist.py` [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/metamist.py#L243-L245). I think that's outside the scope of this change, but just wanted to flag it. Will reference in a related issue.

vivbak:
Hey @michael-harper :) How did we go about testing these changes again?

MattWellie:
> How did we go about testing these changes again?

Latest attempt at testing was by running the Combiner & SampleQC stages with and without dataproc (run on main code and this branch). I compared the output Hail Table with a HT -> TSV, followed by a straight diff between the two (no changes). I'd be happy for someone else to repeat/check my working 😬

michael-harper:
@vivbak @MattWellie I also attempted to view each table visually and couldn't find any differences though this is definitely not a scientific approach. I've asked on Zulip [here](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Compare.20Hail.20tables) for guidance.

michael-harper:
Update: in the Zulip chat linked above there is a function, looks like there is an undocumented function ([_same](https://github.com/hail-is/hail/blob/70ae91b801a920450db48f64ef226b38f0583098/hail/python/hail/table.py#L4381)) the Hail team uses for internal testing (see the Zulip chat [here](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Compare.20Hail.20tables))

Here's my testing using the function:

```python
no_dp_sqc_ht = hl.read_table('gs://cpg-bioheart-test/large_cohort/no_dp_sampleqc/sample_qc.ht')
dp_sqc_ht = hl.read_table('gs://cpg-bioheart-test/large_cohort/back_in_dataproc/sample_qc.ht')
version3_sqc_ht = hl.read_table('gs://cpg-bioheart-test/large_cohort/3-0/sample_qc.ht')
```

image

So it looks like the removal of dataproc hasn't effected the `sample_qc.ht`. The `version3_sqc_ht` is before the y-ploidy changes where we were only using variant sites to calculate sex chromosome ploidies (we are now using variant sites for X and reference blocks for Y) and so the `autosomal_mean_dp` field doesn't exist in this Hail table.

Note that we can't compare the contents of the tables in `version3_sqc_ht` to the current dataproc/no-dataproc tables because the Y-ploidy numbers are way off in `version3_sqc_ht`.

vivbak (approved):
Great job @michael-harper! 

FYI @KatalinaBobowik

PR #613, opened 2024-02-10
gCNV annotate cohort

MattWellie:
Adds the equivalent to AnnotateCohort but for gCNV results

Most of this is new code, some tweaks to previous stages:

1. Fixing of `Number=1` in the added VCF headers to `Number=.` (spotted by @jmarshall)
2. `get_expr_for_xpos` in the SV script now also takes a hail Struct (an update to the type inference, usage here is passing a Struct with the same attributes as a Hail Locus, so behaviour is unchanged) - some linting changes in this file
3. The SV MT generated has a better name (prefixed with `SV` to differentiate from the small variant analysis results. We could also move this to a separate folder instead of `dataset bucket/mt`)
4. Correction to the output names in the Strvctvre Stage - a borked string replacement was generating final files called `XXX.vCf.gz.vcf.bgz`. Also adds metamist registration to these output files

All the rest is adding the AnnotateCohort functionality for gCNV data, emulating the SV equivalent in our pipeline, but based on [this](https://github.com/populationgenomics/seqr-loading-pipelines/blob/master/luigi_pipeline/lib/model/gcnv_mt_schema.py#L16)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.20
+current_version = 1.19.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
   
 env:
-  VERSION: 1.18.20
+  VERSION: 1.19.0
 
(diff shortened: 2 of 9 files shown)


cassimons (approved):
Looks good thanks Matt!

PR #614, opened 2024-02-11
gCNV dataset stages

MattWellie:
Adds the final stages of the gCNV pipeline - AnnotateDataset-equivalent, and MT -> ElasticSearch

I wasn't able to test this final step in the test project, afaik because the acute-care project doesn't have permission to see the relevant seqr ES credentials. We can probably fudge those things to generate some data here and see if we can ingest it

### DIFF
--- a/cpg_workflows/query_modules/seqr_loader_cnv.py
+++ b/cpg_workflows/query_modules/seqr_loader_cnv.py
@@ -63,6 +63,9 @@ def annotate_cohort_gcnv(
     # reimplementation of
     # github.com/populationgenomics/seqr-loading-pipelines..luigi_pipeline/lib/model/sv_mt_schema.py
     mt = mt.annotate_rows(
+        contig=mt.locus.contig.replace('^chr', ''),
+        start=mt.locus.position,
+        pos=mt.locus.position,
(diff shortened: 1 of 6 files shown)


cpg_workflows/query_modules/seqr_loader_cnv.py
```diff
@@ -185,6 +165,88 @@ def annotate_cohort_gcnv(
...
+    mt = read_hail(mt_path)
+
+    mt = mt.annotate_rows(
+        genotypes=hl.agg.collect(
+            hl.struct(
+                sample_id=mt.s,
+                gq=mt.CNQ,
+                cn=mt.CN,
+                end=mt.end,
+                num_exon=hl.missing('int32'),
```
cassimons, line 196: Can we get num_exon from anywhere? It is one of the most useful values for filtering so it would be great to have it.

cassimons (commented):
Looks good. Looking at the test data with Ed yesterday it seemed we still need to tweak some of the fields before it is good to go to prod?

MattWellie:
Yup, there's still something dodgy going on with a couple of fields. I'll try and crack that this week

PR #615, opened 2024-02-12
Auto-restrict some datasets from generating ES indexes

MattWellie:
(where projects are known to not be in Seqr)

Not sure if this should be present in public config files, or moved to the private repo


cassimons:
Can we move this to the private repo please?

MattWellie:
meh, there are other workarounds

PR #616, opened 2024-02-19
Rebatch all GATK-SV SG IDs

MattWellie:
Adds a new script which can load all the individual dataframes from the EvidenceQC stages, filters to currently live samples, then feeds this into the batch generation process we already have.


MattWellie:
Tacked on some config overrides which I've been using lately to get workflows to complete

EddieLF (approved):

PR #618, opened 2024-02-20
Move DenseSubset from Dataproc to Batch

vivbak:
We're currently having issues with the DenseSubset stage. It's failing with little explanation as to why. 

But, we don't need to use Dataproc for the DenseSubset stage anymore. 

This change moved it to a Query on Batch implementation.


vivbak:
Testing here https://batch.hail.populationgenomics.org.au/batches/433500

vivbak:
Test successful: https://batch.hail.populationgenomics.org.au/batches/433503/jobs/1

KatalinaBobowik (approved):

PR #619, opened 2024-02-20
Add logs to help debug _is_reusable

vivbak:
We're getting a failure in the LC pipeline, but it's not reproducible in test. 

We'd like to gain further insight on what's happening in the _is_resuable function.

https://batch.hail.populationgenomics.org.au/batches/433515/jobs/1 

> the following expected outputs for target Cohort("**redacted**", 2 datasets) do not exist: gs://cpg-**redacted**-main/vds/1-0.vds

PR #620, opened 2024-02-20
Add logs to show VDS and MT existence checks are for _SUCCESS files

vivbak:
Did you know the existence checks for REUSE files are based on the existence of the _SUCCESS file? Cause I didn't. 

May no-one question their sanity like this again. 

Previous issue, https://batch.hail.populationgenomics.org.au/batches/433515/jobs/1 was because the _SUCCESS file didn't exist in the .vds folder

### DIFF
--- a/cpg_workflows/utils.py
+++ b/cpg_workflows/utils.py
@@ -37,10 +37,10 @@ def read_hail(path):
 
 
 def checkpoint_hail(
-        t: hl.Table | hl.MatrixTable,
-        file_name: str,
-        checkpoint_prefix: str | None = None,
-        allow_reuse=False
+    t: hl.Table | hl.MatrixTable,
+    file_name: str,
+    checkpoint_prefix: str | None = None,
+    allow_reuse=False,
 ):
     """
     checkpoint method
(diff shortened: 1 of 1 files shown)


jmarshall:
Hail tables and VDSes are written as multiple files in a directory hierarchy, which can't be done atomically. So testing whether one of these has been written completely requires designating one of the files to always be written last, so that implementations know which file to check. Hail does this the best-practices way of having that file be a 0-length flag file, and calls it the fairly clear `_SUCCESS`. TL;DR as a Hail nerd I did know that; mostly nothing outside Hail should need to know that, but this function (`exists_not_cached()`) needs to check table/VDS basic validity so contains that hail-internal information. And it has enough comments to clarify this, if you had somehow known in advance to look there…

So the problem in batch 433515 was that the hail object had not been fully written out (or incompletely copied from somewhere else?), so was being correctly categorised as not existing.

Nevermind the background; what about the PR?

To be honest, I don't love this. The logs are cluttered enough already, and ideally we wouldn't need to expose this Hail implementation detail in them. However logging these details would have saved you time (and sanity!) while debugging this issue. So I would be happy with this as `logging.debug(…)` instead of `logging.info(…)` so that these details would not appear by default but could be switched on when needed. However `exists_not_cached()` already has a `logging.debug(f'Checked {path} … [exists|missing]…')` statement further down that would show the exact `…/_SUCCESS` path checked. So IMHO this PR is not needed; instead a way of activating the existing `logging.debug(…)` is needed, probably a config setting to set the initial logging level (which currently seems to be always INFO) if we don't already have one.

(I've made a similar comment on yesterday's #619 as well, which IMHO would usefully be level DEBUG rather than INFO too.)

vivbak:
Good point! #619 was a little frantic, because we were under a time crunch & needed to figure out what was wrong. 

Closing this for now. Our logs have always been an area that needs significant improvement. Not verbose enough when we need them, and far too verbose every other time.

jmarshall:
> Not verbose enough when we need them, and far too verbose every other time.

Now that is very accurate and wise 🤣

PR #621, opened 2024-02-20
Remove dataproc from Combiner stage

michael-harper:
This PR is part of a larger effort to migrate away from reliance on Dataproc in the Large Cohort pipeline. 

This PR moves Combiner to a query-on-batch implementation


michael-harper:
This successfully produced an output at: `gs://cpg-bioheart-test/vds/no_dp_combiner.vds`

Comparison between vds versions in the bioheat-test bucket was performed inside a Jupyter notebook:

image

Note: the `back_in_dataproc.vds` file was generated from the SampleQC dataproc removal and represents a vds created from within dataproc, but from the version of the pipeline that had the y-ploidy issues updated (while `3-0.vds` was before these fixes). This shouldn't effect the vds (hence why they're all 'the same') but for explanatory purposes, that is the origin of the file.

Looks like the reference blocks and variant data are the same between dataproc and non-dataproc Combiner outputs

vivbak (approved):
Well done 🙌

PR #626, opened 2024-02-21
Remove dataproc from MakeSiteOnlyVcf

michael-harper:
This pull request aims to address concerns regarding the dependency on Dataproc for the MakeSiteOnlyVcf stage of large cohort processing. See this Slack thread [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1708485526044469).

Moving this job away from non-preemptible machines within Dataproc to pre-emptible machines in QoB raises concerns about the risk of process failure due to the potentially extended processing times for large datasets. Nonetheless, it seems possible to remove dataproc from this stage.

Note: An alternative is to move towards Hail's ability to conduct parallel exports during VCF merging that would help nullify the pre-emption issue. Although integrating parallel exports into the pipeline may be complex, it could significantly optimise the process and eliminate the need for Dataproc.
**Update**: @MattWellie has created a proof-of-concept for this approach #629 !


michael-harper:
**Note**: The output Hail Tables of dataproc (dp) vs. no-dataproc (no-dp) are different. The `AS_FS` field appears to be different between dp and no-dp. At first it was thought that this was due to a stochastic nature of the FisherStrand variable however, when comparing two no-dp `site_only.ht` files (see screenshot below cell [4]) there appears to be no difference (the `no_path_no_dp_site_only_ht` was a result of a second no-dp test to test whether `Path` or `str` variables had to be passed - turns out `str` variables are required!). 

Screenshot 2024-02-28 at 9 01 50 am

I've included the full output of the comparison as a txt file
[MakeSiteOnlyVcf_dp_vs_nodp_comparison.txt](https://github.com/populationgenomics/production-pipelines/files/14426263/MakeSiteOnlyVcf_dp_vs_nodp_comparison.txt)

cpg_workflows/stages/large_cohort.py
```diff
@@ -20,8 +20,8 @@ class Combiner(CohortStage):
     def expected_outputs(self, cohort: Cohort) -> Path:
         if vds_version := get_config()['workflow'].get('vds_version'):
             vds_version = slugify(vds_version)
-            if not vds_version.startswith('v'):
-                vds_version = f'v{vds_version}'
+            # if not vds_version.startswith('v'):
```
vivbak, line 23: Could we please revert this in this PR? All good if we end up going for this change, but it's outside the scope and should be introduced in a separate PR :)

vivbak (changes requested):

michael-harper:
Running again to check if same inputs generate different outputs. This is to check if the AS_FS field has a random component.

Two batches here: 

    https://batch.hail.populationgenomics.org.au/batches/438609/jobs/1

    https://batch.hail.populationgenomics.org.au/batches/438612/jobs/1 

The output of the above batches are the same, suggesting that the AS_FS field does not change due to randomness and the change in output detected is likely due to Dataproc and non-Dataproc usage.
Screenshot 2024-04-02 at 10 20 53 AM


(Note: both run2_no_dp_siteonly and run2_no_dp_siteonly2 use the same inputs, also note that these outputs are the same as gs://cpg-bioheart-test/large_cohort/3-0/no_dp_site_only.ht)

michael-harper:
We are content with the current differences between using Dataproc and not using Dataproc. The difference seems to be due to the way math is happening (see [Zulip chat here](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Difference.20between.20dataproc.20and.20non-dataproc.20output/near/433256063)). This change can be merged

vivbak (approved):

PR #628, opened 2024-02-26
gCNV Joint Genotyping

MattWellie:
Lots changed here, I'll try and make a summary...

## Rationale

The previous incarnation of gCNV calls through to annotated joint call was using the wrong data - I erroneously thought that the Intervals VCF was the correctly joint genotyped source of information. Instead the workflow has a whole extra set of steps as shown in [this WDL](https://github.com/broadinstitute/gatk/blob/master/scripts/cnv_wdl/germline/joint_call_exome_cnvs.wdl#L113).

This meant restarting from the per-sample call generation, and shoving in a few more steps. That's what this change is doing.

#### Stuff that happens all over

- the batch instance is no longer passed to the job-generating methods, instead the `get_batch()` method is used to accomplish the same
- altered output file name to no longer contain the `cohort.name` - most/all data will be processed under the `seqr` project, so all files would have seqr as a name
- where multiple values are obtained from prior stages (e.g. `inputs.as_path(cohort, Stage, 'key')`, these are obtained once as a dict, and each key is then pulled out. This should avoid re-triggering the logic to find prior stage outputs.

#### DeterminePloidy

- output is stored in regular buckets, not tmp - the joint-genotyping stage now requires these files, and the (relatively) cheap final stages can't be repeated without these files which are expensive to reproduce

#### GermlineCNV

- output files stored in main bucket, same reasoning applies

#### GermlineCNVCalls

- fetching input from prior stages is done using the Cohort, not the Dataset. This was not an issue previously as both were Acute-Care.

#### PrepareVcfsForMerge

- Step completely removed, it was adapting the wrong VCFs, and is not not necessary

#### GCNVJointSegmentation

- Pulls in the Segment VCFs from all individual SG IDs, and triggers the joint-segmentation algorithm. This was done using a dummy Pedigree (IDs inside the files are from prod, file names and metamist entries are from test. GATK failed when the pedigree with prod IDs mismatched the file content)

#### RecalculateClusteredQuality

- Pulls in the clustered results and re-scores the quality for each individual's calls based on the cohort context. This is essentially a second pass through PostProcessGCNVCalls with some additional arguments.
- `Missing`: this step produces a QC file, with the intention that only QC PASS'es are used when generating the joint call. Based on this test run all samples failed 🙃 so we should reconsider the thresholds in use there. If we generate some QC results that make sense there, this should be the end of a first workflow - Hail can't make decisions about which samples to pass into the VCF merging state based on QC results that are generated within the workflow, so we would have to terminate here, and run a second workflow to parse the QC results, and use that data when generating the joint call.

#### FastCombineGCNVs

- Mostly unchanged, except for the source VCF used when merging

---

#### postprocess_calls

- Restructured so that two different stages can call this method - one naively to generate VCFs from the sharded Tar files, and once to recalculate qualities in a joint-call context. If the intervals, clusters, and QC file args are provided, the second behaviour is triggered
- The 3 additional arguments are added to the GATK tool command-line args, and read into the batch as appropriate
- The output directory is generated using `mkdir` - I don't know why or how, but the Tar'chive unpacking seems to delete/obfuscate the Hail Batch outputs folder, so the workflows crashed having nowhere to write to.
- The job `command`s are added individually instead of as a single list of commands. The reasoning here is kinda dumb - [this conditional behaviour](https://github.com/populationgenomics/production-pipelines/blob/a3b2150f226ff4c595eba5ff1fbfa58e57e37bd0/cpg_workflows/jobs/gcnv.py#L377-L398) creates the QC file for each sample, then writes that file out to GCP. The hail-batch resource files are only created once the `"{job.output}"` text is included in a call to `job.command`. The previous structure was elegant, but meant that `j.qc_file` wouldn't exist until all commands are sent to Hail. That means that this conditional loop would have to happen twice:
    - if qc file: generate qc stats
    - send command strings to Hail Batch
    - if qc file - retrieve the qc output file
- which undermines the elegance of the original solution

#### run_joint_segmentation  & joint_segment_vcfs

- Runs the joint segmentation operation in GATK, but as a 2-step. If the total number of samples is greater than X (from config), produce an intermediate round of merges
- Step 1: split the group of SG IDs into groups of length X, then produce one merged VCF for each
- Step 2: merge all the intermediate merge files

### update_vcf_attributes

- This method previously compensated for 'missing values' in the combined VCF (mostly missing because I was using the wrong input data. 
- Behaviours which are still required (to comply with Strvctvre more than anything else:
   - add SVLEN (header entry exists, but value is not populated in variant.INFO...)
   - generate a unique ID for each variant (the bcftools merge creates a huge list of ;-delimited IDs, not all of which are relevant to the specific variant)
   - Reassembles the INFO string, plus SVLEN

---

I've re-run this whole pipeline from scratch, with the latest job [here](https://batch.hail.populationgenomics.org.au/batches/433927) (I cancelled it once I spotted that the MT outputs would be registered in metamist wrongly as `es-index`es)

Apologies to all for the general grottiness of this PR

https://batch.hail.populationgenomics.org.au/batches/433950

### DIFF
--- a/configs/defaults/gcnv.toml
+++ b/configs/defaults/gcnv.toml
@@ -1,12 +1,12 @@
 [workflow]
 name = 'gcnv'
-intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.v1.interval_list'
+#intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.v1.interval_list'
 exclude_intervals = ['chrM']
 # Aim for 10-50Mbp genomic coverage per shard
 interval_shards = [['chr1'], ['chr2', 'chr3'], ['chr4', 'chr5'], ['chr6', 'chr7'], ['chr8', 'chr9', 'chr10'], ['chr11', 'chr12'], ['chr13', 'chr14', 'chr15'], ['chr16', 'chr17'], ['chr18', 'chr19', 'chr20'], ['chr21', 'chr22', 'chrX', 'chrY']]
-ploidy_priors = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv'
+#ploidy_priors = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv'
 allosomal_contigs = ['chrX', 'chrY']
-num_samples_per_scatter_block = 2
+num_samples_per_scatter_block = 5
 
 [images]
 gatk_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv/gatk:2023-07-13-4.4.0.0-43-gd79823f9c-NIGHTLY-SNAPSHOT"
--- a/configs/defaults/gcnv.toml
+++ b/configs/defaults/gcnv.toml
@@ -23,3 +23,4 @@ sv_pipeline_docker = "australia-southeast1-docker.pkg.dev/cpg-common/images/sv/s
 # github.com/broadinstitute/gatk-sv/blob/main/inputs/templates/test/AnnotateVcf/AnnotateVcf.json.tmpl#L4-L8
 external_af_population = ['ALL', 'AFR', 'AMR', 'EAS', 'EUR']
 external_af_ref_bed_prefix = 'gnomad_v2.1_sv'
+noncoding_bed = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/noncoding.sort.hg38.bed'
--- a/configs/defaults/gcnv.toml
+++ b/configs/defaults/gcnv.toml
@@ -4,13 +4,13 @@ intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.
 exclude_intervals = ['chrM']
 # Aim for 10-50Mbp genomic coverage per shard
 interval_shards = [['chr1'], ['chr2', 'chr3'], ['chr4', 'chr5'], ['chr6', 'chr7'], ['chr8', 'chr9', 'chr10'], ['chr11', 'chr12'], ['chr13', 'chr14', 'chr15'], ['chr16', 'chr17'], ['chr18', 'chr19', 'chr20'], ['chr21', 'chr22', 'chrX', 'chrY']]
-#ploidy_priors = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv'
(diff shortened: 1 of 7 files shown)


MattWellie:
TODO remove that dummy pedigree

configs/defaults/gcnv.toml
```diff
@@ -4,9 +4,14 @@ intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.
 exclude_intervals = ['chrM']
 # Aim for 10-50Mbp genomic coverage per shard
 interval_shards = [['chr1'], ['chr2', 'chr3'], ['chr4', 'chr5'], ['chr6', 'chr7'], ['chr8', 'chr9', 'chr10'], ['chr11', 'chr12'], ['chr13', 'chr14', 'chr15'], ['chr16', 'chr17'], ['chr18', 'chr19', 'chr20'], ['chr21', 'chr22', 'chrX', 'chrY']]
-ploidy_priors = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv'
+#ploidy_priors = 'gs://cpg-common-main/references/gatk-sv/hg38/v0/sv-resources/resources/v1/hg38.contig_ploidy_priors_homo_sapiens.tsv'
```
cassimons, line 7: Can we just rm this if we don't need it anymore?

configs/defaults/gcnv.toml
```diff
@@ -4,9 +4,14 @@ intervals_path = 'gs://cpg-common-main/references/hg38/v0/exome_calling_regions.
...
 allosomal_contigs = ['chrX', 'chrY']
 
+# https://github.com/broadinstitute/gatk/blob/cfd4d87ec29ac45a68f13a37f30101f326546b7d/scripts/cnv_cromwell_tests/germline/cnv_germline_case_scattered_workflow.json
```
cassimons, line 10: Am I reading it correctly that this is the config used for testing? Could that be why the max events thresholds are so low (they are set for a single downsampled chrome)?

cassimons (approved):
Looks great thanks Matt

configs/defaults/gcnv.toml
MattWellie: Yup! There were two files added to this WF's config - one is interval regions which should be relative to this exact cohort's exome capture, but this looks value was the same as the default in the analysis-runner config so I've removed from this WF-specific config

configs/defaults/gcnv.toml
MattWellie: That's what I suspect... all the files and thresholds look subsampled/super low, so I don't think they make sense to adopt directly. I'm happy to continue for now just ignoring the QC check until we can get a line on sensible exome-wide values

cassimons: I wonder if we should ping the Broad guys and ask if they have prod config we should use?

PR #633, opened 2024-03-01
gCNV joint call dataset stages

MattWellie:
Corrected version of https://github.com/populationgenomics/production-pipelines/pull/614

This version follows on from the joint-called CNV data instead of the wrongly processed Intervals VCFs

Currently loaded into AC Exomes in Seqr staging, could do with prettying up, but the results at least load this time around

Most of the MtToEsCNV stage is common to MtToEs and MtToEsSv, so that should probably be abstracted out into a varianttype-independent Dataproc wrapper. Feels outside of the scope here, but a candidate for a tidy-up PR to shed some lines

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.20.0
+current_version = 1.21.0
 commit = True
 tag = False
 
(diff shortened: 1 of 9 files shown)


cassimons (approved):
LGTM and the test data in staging looks excellent 🙌

PR #635, opened 2024-03-06
Custom Cohorts in Production Pipelines 

vivbak:
Depends on https://github.com/populationgenomics/metamist/pull/615

Support Custom Cohorts in Production Pipelines!


cpg_workflows/defaults.toml
```diff
@@ -1,4 +1,7 @@
 [workflow]
+# Cohorts to use as inputs.
+# input_cohorts = []
+
 # Datasets to load inputs. If not provided, datasets will be determined
 # automatically based on the input provider implementation:
 #input_datasets = []
```
MattWellie, line 7: does cohorts-as-input negate the need for input_datasets, or is there a transitional period where both are accepted? I remember some discussion on this, but not sure where we ended up

MattWellie: Same comment applies for this whole list of arguments; skip_sgs, only_sgs, etc.

cpg_workflows/inputs.py
```diff
@@ -26,20 +26,41 @@ def create_cohort() -> Cohort:
...
     Add datasets in the cohort. There exists only one cohort for the workflow run.
     """
-    analysis_dataset_name = get_config()['workflow']['dataset']
-    dataset_names = get_config()['workflow'].get('input_datasets', [analysis_dataset_name])
-    skip_datasets = get_config()['workflow'].get('skip_datasets', [])
+    config = get_config()['workflow']
+    analysis_dataset_name = config['dataset']
+    custom_cohort_ids = config.get('input_cohorts', [])
+    input_datasets = config.get('input_datasets', [])
+    skip_datasets = config.get('skip_datasets', [])
```
MattWellie, line 33:
```suggestion
    from cpg_utils.config import config_retrieve
    
    config = config_retrieve(['workflow'])
    analysis_dataset_name = config_retrieve(['workflow', 'dataset'])
    custom_cohort_ids = config_retrieve(['workflow', 'input_cohorts'], [])
    input_datasets = config_retrieve(['workflow', 'input_datasets'], [])
    skip_datasets = config_retrieve(['workflow', 'skip_datasets'], [])
```

preference for the use of `config_retrieve` instead of `get_config()[key][...]`. `config_retrieve` will fail descriptively if the nested keys aren't present, and incorporates a default value return.

cpg_workflows/defaults.toml
MattWellie: I just saw the deprecation message below, so I guess for a while we're swapping input_datasets -> dataset_names behind the scenes.

cpg_workflows/inputs.py
```diff
@@ -26,20 +26,41 @@ def create_cohort() -> Cohort:
...
+
+    # Additional logic to support cohorts + datasets as inputs. In future, cohorts will only be supported.
+    if custom_cohort_ids and input_datasets:
+        raise ValueError('Cannot use both custom_cohort_ids and input_datasets in the same workflow')
+
+    if custom_cohort_ids:
+        # TODO: Handle more than one cohort
+        sgs_by_dataset = get_metamist().get_sgs_by_project_from_cohort(custom_cohort_ids[0])
```
MattWellie, line 41: assuming this is still a placeholder until multi-cohort runs exist? It would be good to have the process fail out if `custom_cohort_ids` is longer than 1 entry until this is supported

cpg_workflows/metamist.py
```diff
@@ -55,6 +55,40 @@
...
+                sample {
+                    project {
+                        name
+                    }
+                    externalId
+                    participant {
+                        id
+                        externalId
+                        reportedSex
+                        meta
```
MattWellie, line 77: I recently made a change [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/metamist.py#L42) and in inputs.py [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/inputs.py#L42-L43) to facilitate grabbing phenotypes for use in Exomiser. Can we chuck those back in?

both on line 42, how cool is that

MattWellie: https://github.com/populationgenomics/production-pipelines/commit/001e92e4d2b8c3b6c6dc39052c90953c9f1e4d53 was the PR, most of the changes were in Exomiser, a little bit in metamist/inputs and corresponding changes to get the tests to pass

MattWellie (commented):
Sorry this is a lot of commenting for what is ultimately a good thing that I like

cpg_workflows/defaults.toml
vivbak: We will be supporting both for an interim period while we make the transition across :)

MattWellie (approved):
Looked good yesterday. I need to check up the SGs by cohort unit tests to check that phenotypes are popping in correctly. I believe they're parsed from the graphql result with an empty default, so I might add a value to the mocked result to check for

cpg_workflows/inputs.py
```diff
@@ -26,20 +26,43 @@ def create_cohort() -> Cohort:
...
+        if custom_cohort_ids:
+            sgs = sgs_by_dataset[dataset_name]
+        else:
+            sgs = get_metamist().get_sg_entries(dataset_name)
+
+        for entry in sgs:
             metadata = entry.get('meta', {})
             update_dict(metadata, entry['sample']['participant'].get('meta', {}))
-
-            # phenotypes are managed badly here, need a cleaner way to get them into the SG
```
jmarshall, line 42: The code hasn't changed here; does the TODOesque comment still apply?

cpg_workflows/metamist.py
```diff
@@ -206,6 +259,21 @@ def __init__(self) -> None:
...
         self.aapi = AnalysisApi()
 
+    def get_sgs_by_project_from_cohort(self, cohort_id: str) -> dict:
+        """
+        Retrieve sequencing group entries for a cohort.
+        """
+        entries = query(GET_SEQUENCING_GROUPS_BY_COHORT_QUERY, {'cohort_id': cohort_id})
+
+        # Create dictionary keying sequencing groups by project
+        # {project_id: [sequencing_group_1, sequencing_group_2, ...]}
```
jmarshall, line 269:
```suggestion
        # {project_id: [sequencing_group_1, sequencing_group_2, ...], ...}
```

cpg_workflows/metamist.py
```diff
@@ -197,6 +232,24 @@ def parse(data: dict) -> 'Analysis':
         return a
 
 
+def sort_sgs_by_project(response_data) -> dict:
+    """
+    Create dictionary organising sequencing groups by project
+    {project_id: [sequencing_group_1, sequencing_group_2, ...]}
+    """
+    result_dict: dict[str, Any] = {}
```
jmarshall, line 240: Is this always going to be `dict[str, list[str]]` or could the dict values really be other kinds of `Any`?

jmarshall (approved):
Some extremely trivial comments, otherwise LGTM.

cpg_workflows/metamist.py
vivbak: Yeah good point, this is very much a sign of me being lazy

cpg_workflows/inputs.py
vivbak: Ooops, deleting this was most definitely a mistake.

PR #637, opened 2024-03-07
Adjust pipeline for quality filtered samples

### LINKED ISSUE #636: Correctly process samples filtered out during the GATK-SV pipeline
The GATK-SV pipeline routinely fails at the end, during the 'generate cohort subset' stage. This stage takes the joint call, and extracts out the samples in each separate cohort from the joint-called whole. This takes a list of sample IDs, and routinely fails with something like

```
Exception: Found 8/669 samples in the subset set that do not matching IDs in the variant callset.
IDs that aren't in the callset: {'CPGN', 'CPGN+1', ...}
```

I was not sure where this was creeping in, so I was tweaking config files to get the joint-calls through. I've since identified the issue:

1. FilterBatch generates lists of outliers for each separate batch, and grooms those out of the batch callsets
2. These lists of excluded samples are present in a file like `gs://cpg-seqr-main/gatk_sv/a7f4438075b14b7353065b78461a41256da665_241/FilterBatch/outliers.samples.list`
3. When generating the filtered batch VCF, these samples are specifically excluded. 

This is solvable by adding a new stage, reading the excluded sample ID list for each separate batch, combining them into a single file, and registering the file/data in metamist so we can easily access this data in future. 

We should also pass this list of all excluded samples to the cohort subsetting Stage(s), so that we stop trying to subset the joint- call to a group of samples which are not contained within that joint-call.

---

Issues (@vivbak not sure if you have an idea here)

1. We have Stage wrappers that generate analysis entries for files created in each Stage. These contain a list of all the contained sample IDs. At the moment we are creating analysis entries for the final joint-call and project-specific ES indexes, and registering them against all SG IDs in the cohort. Due to this filtering, not all the SG IDs are actually present in the joint-call, so this metadata is misleading.

MattWellie:
Closes #636

Issue:

- Samples are filtered during the GATK-SV pipeline for quality reasons (no dispute there). The list of excluded identifiers per batch is stored in GCP, but we have so far not read that file or done anything with the contents.
- The excluded samples then are 'silently' filtered from the joint-calls (e.g. by a `bcftools -S ^<ID list>`, so the SG IDs in the cohort/dataset becomes disconnected from the SG IDs remaining in the data/VCFs)
- Stage output files are registered as metamist analysis entries, linked to all the SG IDs in the cohort - including SG IDs/samples removed from the VCFs/MTs. This creates misleading analysis entries: "SG ID CPGXX should be included in this joint-call, the metamist entry claims they are present, why can't we find their data in Seqr"
- Fatal: SG ID subsetting (from cohort-level joint call to dataset-level joint call) fails, because not all required SG IDs are present

Solution:

- Create a Stage-independent file path (`gs://<BUCKET>/gatk_sv/<DATE>/combined_exclusion_list.txt`) so that we can reference this file in both stages and metamist annotators
- Collect all per-batch files containing SG IDs which were removed for quality reasons. This is done using a `gcloud object compose` - this is a `cat` between cloud objects without having to localise them first. This is written to the path from the previous point.
- Read this file during the AnnotateDataset stage when subsetting - we parse the list of all filtered IDs and adjust the expectation of samples present in the callset, rather than failing because we can't find quality-filtered samples.
- Add the name/path of this file to the metamist registration calls (the `update_analysis_meta` Callables, with a specific key a specific key `remove_sgids`) - during the metamist file registration take the list of SG IDs generated by `target.get_sequencing_group_ids()`, read in the excluded SG ID file, and do a set subtraction - this should make the analysis object's `sequencing_group_ids` accurate to the data files.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.19.0
+current_version = 1.20.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
   
 env:
-  VERSION: 1.19.0
+  VERSION: 1.20.0
 
 jobs:
   docker:
--- a/configs/gatk_sv/config_overrides.toml
+++ b/configs/gatk_sv/config_overrides.toml
@@ -19,6 +19,8 @@ disk_gb = 200
 
 [resource_overrides.MakeCohortVcf]
 run_module_metrics = false
+[resource_overrides.MakeCohortVcf.runtime_override_plot_qc_per_family]
+mem_gb = 32
 
 # resource overrides for indirect workflows
 
--- a/cpg_workflows/jobs/seqr_loader.py
+++ b/cpg_workflows/jobs/seqr_loader.py
@@ -7,12 +7,11 @@ from hailtop.batch.job import Job
 
 from cpg_utils import Path
 from cpg_utils.config import get_config
-from cpg_utils.hail_batch import image_path, query_command
+from cpg_utils.hail_batch import get_batch, image_path, query_command
 from cpg_workflows.query_modules import seqr_loader
 
 
 def annotate_dataset_jobs(
-    b: Batch,
     mt_path: Path,
     sequencing_group_ids: list[str],
     out_mt_path: Path,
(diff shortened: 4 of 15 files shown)


cassimons:
How do we communicate the failed samples to the anayests and can we export any stats on why they failed?

Eventually, this would be good to plumb into the hypothetical QC event module in metamist. Until then, can we post to slack in a similar way we do with multiQC fails?

cassimons (approved):
LGTM. It would be good to solve the notification issue but I don't think that need to be in this PR.

PR #639, opened 2024-03-11
Parallel exports gcloud compose [prototype]

MattWellie:
Replaces #629 

This upgrades the previous solution by incorporating the `gcloud storage objects compose` functionality to concatenate the separate fragments in GCP without needing to localise first.

For what it's worth, this same logic could be used to replace the current 'gather VCFs from fragments' [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/jobs/vcf.py#L64) (so long as the files are Block-GZipped, which they are)

### DIFF
--- a/cpg_workflows/jobs/make_vcf_from_hail.py
+++ b/cpg_workflows/jobs/make_vcf_from_hail.py
@@ -108,7 +108,6 @@ def make_sites_only(obj: hl.MatrixTable | hl.Table) -> hl.MatrixTable | hl.Table
     LOGGER.info('No further processing applied, assuming VDS')
     return obj
 
-    raise ValueError(f'Unexpected hail object type {type(obj)}')
 
 
 def create_vcf_from_hail(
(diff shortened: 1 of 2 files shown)


jmarshall:
I looked at using `gcloud compose` in the gCNV pipeline at one point, but lost interest when I saw the limit was 32 items in one composition. Is that an issue here? I guess it's enough for e.g. one-shard-per-chromosome for humans…

MattWellie:
> lost interest when I saw the limit was 32 items in one composition. Is that an issue here? I guess it's enough for e.g. one-shard-per-chromosome for humans…

I've mitigated this with a rolling merge - start off with the list of all fragments, condense that in blocks of 30, then keep condensing the list until you end up with a single product

cpg_workflows/jobs/make_vcf_from_hail.py
```diff
@@ -0,0 +1,282 @@
...
+    temp_chunk_prefix_num = 1
+    condense_job = None
+    while len(chunk_paths) > 1:
+        condense_temp = join(temp, f'temp_chunk_{temp_chunk_prefix_num}')
+        temp_chunk_prefix_num += 1
+        chunk_paths, condense_job = compose_condense_fragments(
+            chunk_paths,
+            condense_temp,
+            depends=condense_job
+        )
```
MattWellie, line 250: @jmarshall

MattWellie:
Better demo of this in the new RD combiner pipeline

PR #640, opened 2024-03-11
Various tweaks relating to metamist

MattWellie:
1. Removes the no-logging query wrapper, now pointless since the implementation of https://github.com/populationgenomics/metamist/pull/650
2. Removes the Enum for analysis types in cpg_workflows - this is not viable now that new analysis types can be created at will in metamist. This list is out of date and not worth maintaining
3. Removes all uses of the above that I can find

This is unbelievably low priority, I just found some dodgy things

### DIFF
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -8,9 +8,9 @@ from cpg_utils.config import get_config, update_dict
 
 from cpg_workflows.filetypes import CramPath, GvcfPath
 
-from .metamist import AnalysisType, Assay, MetamistError, get_metamist, parse_reads
-from .targets import Cohort, PedigreeInfo, SequencingGroup, Sex
-from .utils import exists
+from cpg_workflows.metamist import Assay, MetamistError, get_metamist, parse_reads
+from cpg_workflows.targets import Cohort, PedigreeInfo, SequencingGroup, Sex
+from cpg_workflows.utils import exists
 
 _cohort: Cohort | None = None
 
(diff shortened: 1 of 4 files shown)


cpg_workflows/metamist.py
```diff
@@ -151,38 +125,6 @@ def parse(name: str) -> 'AnalysisStatus':
         return {v.value: v for v in AnalysisStatus}[name.lower()]
 
 
-class AnalysisType(Enum):
```
illusional, line 154: You don't want to keep this definition - make it a clear process for how people define new analysis types in production pipelines?

MattWellie: I don't know if this makes sense in its current form - this enumeration would track something in code that can be changed with an API call. If we add new entries using the API, we'd have to create a corresponding change to this Enum. 

Upon startup, an API call to find the currently existing analysis types for the given metamist instance could be used so that we don't try and post/create an analysis type that's unacceptable. That could take the place of this Enum so that the interface between the two wouldn't require code changes to remain in sync. Just spitballin' here

cpg_workflows/inputs.py
```diff
@@ -146,7 +146,7 @@ def _populate_analysis(cohort: Cohort) -> None:
                     analysis.output,
                 )
                 sequencing_group.gvcf = GvcfPath(path=analysis.output)
-            elif exists(sequencing_group.make_gvcf_path()):
+            elif exists(sequencing_group.make_gvcf_path().path):
```
jmarshall, line 149: Is this (and the similar change on line 165) a bug fix?

These things are a GvcfPath and CramPath respectively, which (despite the names) are not is-a-kind-of-Path 😢  so getting at their `path` member (which actually is-a-Path-or-str) is warranted.

Adding `.path` is probably clearer, though deep in the cloudpathlib guts of `exists()` there is an explicit `str(…)` which may well have made this work. Hence question: did you observe this code not firing — thus buggy behaviour — or is this just a clarification?

MattWellie: Screenshot 2024-04-03 at 1 40 39 PM

This popped up in my IDE as a warning - the argument being provided is an invalid type, so I'm guessing either this somehow works on the string/.path representation, or we never exercise this code at runtime

jmarshall: Interesting… So the remaining question is why mypy didn't complain about this at the time… 🤔

PR #643, opened 2024-03-12
Update MyPy version

vivbak:
Pulling in Matt's commit to fix https://github.com/populationgenomics/production-pipelines/issues/641 

Note, the pre-commit-config is due for a huge overall. This will be addressed in a later PR (following custom-cohorts).


MattWellie:
@vivbak that's on me - I forgot to mention that the updated linter revealed some linting failures which previously sailed under the radar

MattWellie (approved):

vivbak:
Haha just figured it out and pulled across!

PR #645, opened 2024-03-13
Update cpg-utils and analysis-runner paths after refactor

illusional:
This PR updated cpg_utils and analysis-runner paths to coincide with the release of https://github.com/populationgenomics/cpg-utils/pull/152. Note, that cpg-utils PR was _mostly_ backwards compatible. Context: https://centrepopgen.slack.com/archives/C04M1G5HLM9/p1713308700232179

There are a few changes worth highlighting:

- when reference_path moved from cpg_utils.hail_batch to config, it became a str, not a Path
- Dataproc and Cromwell were moved from analysis-runner to cpg-utils.
- There were improvements to dataproc + cromwell, which meant they don't load configs on module load, so we can move those imports towards the top.


MattWellie:
holymoly

MattWellie:
https://github.com/populationgenomics/production-pipelines/blob/f9035d29f3516b05ab84ace72343c5c510132e1d/cpg_workflows/jobs/picard.py#L99

casting as a string here should fix the test, but this should probably be string-ified all the way back for consistency
```-I {b.read_input(str(source_intervals_path))} \```

MattWellie:
@vivbak @illusional is this still being reviewed? Would be shiny to update soon 😶 
I've stacked #696 on top of this, updating GATK-SV to the new syntax/imports

vivbak (approved):

PR #654, opened 2024-03-18
Reduce requested storage for seqr_loader jobs based on workflow sequencing type

EddieLF:
Changes to the requested storage for various jobs in the seqr_loader pipeline. Some jobs are seriously over-requesting storage and this will hopefully bring down some of that overhead.

Updates the function `storage_for_joint_vcf` so that it uses the `get_config()['workflow']` dict, which has the `sequencing_type` field. The `storage_for_joint_vcf` function is used by the `gather_vcfs` jobs: `"Merge X {site-only} VCFs"`, and the `tabix_vcf` jobs: `"Tabix sorted VCF"`. 

If the sequencing type is exome, the requested disk storage is reduced to 0.1GB per sequencing group. If the sequencing type is genome or otherwise, the storage per sequencing group has been reduced from 4GB to 1.5GB, or 1GB if sites-only.

~Also updates the VQSR workflow to use a `'small_disk'` as opposed to `'huge_disk'` when applying the VQSR threshold for SNPs. The `apply_recalibration_snps` is used by the jobs: `"VQSR: ApplyVQSR SNPs"` which does not seem to need a huge disk size to complete successfully.~

For example, in the recent exome seqr load jobs (2202 sequencing groups):

| Job | Requested Storage | Peak Storage Use | Job link |
|--------|--------|--------|--------|
| Joint genotyping: Merge 600 VCFs | 9.2TB | 86GB | [Job](https://batch.hail.populationgenomics.org.au/batches/433456/jobs/2955) |
| Joint genotyping: Tabix sorted VCF | 9.2TB | 33GB | [Job](https://batch.hail.populationgenomics.org.au/batches/433456/jobs/2956) |
| VQSR: Merge 600 site-only VCFs | 2.3TB | 1.5GB | [Job](https://batch.hail.populationgenomics.org.au/batches/433456/jobs/4163) |
| VQSR: Tabix sorted VCF | 2.3TB | <1GB | [Job](https://batch.hail.populationgenomics.org.au/batches/433456/jobs/4164) |
| VQSR: ApplyVQSR SNPs, x/600 | 525GB | <1GB | [Job](https://batch.hail.populationgenomics.org.au/batches/433456/jobs/4089) | 

And the recent genome seqr load jobs (3549 sequencing groups):

| Job | Requested Storage | Peak Storage Use | Job link |
|--------|--------|--------|--------|
| Joint genotyping: Merge 1200 VCFs | 14.9TB | 3.1TB | [Job](https://batch.hail.populationgenomics.org.au/batches/433937/jobs/4) |
| Joint genotyping: Tabix sorted VCF | 14.9TB | 1.2TB | [Job](https://batch.hail.populationgenomics.org.au/batches/433937/jobs/5) |
| VQSR: Merge 1200 site-only VCFs | 3.7TB | 22GB | [Job](https://batch.hail.populationgenomics.org.au/batches/434576/jobs/2403) |
| VQSR: Tabix sorted VCF | 3.7TB | 8.5GB | [Job](https://batch.hail.populationgenomics.org.au/batches/434576/jobs/2404) |
| VQSR: ApplyVQSR SNPs, x/1200 | 525GB | 7.5GB | [Job](https://batch.hail.populationgenomics.org.au/batches/434576/jobs/1249) | 

I think these initial reductions should be sufficient to significantly reduce the storage request overheads without disrupting any workflows.

### DIFF
--- a/cpg_workflows/resources.py
+++ b/cpg_workflows/resources.py
@@ -337,14 +337,18 @@ def joint_calling_scatter_count(sequencing_group_count: int) -> int:
 
 
 def storage_for_joint_vcf(
-    sequencing_group_count: int | None, site_only: bool = True
+    sequencing_group_count: int | None, site_only: bool = True, job_attrs: dict | None = None
 ) -> int | None:
     """
     Storage enough to fit and process a joint-called VCF
     """
     if not sequencing_group_count:
         return None
-    gb_per_sequencing_group = 1
-    if not site_only:
-        gb_per_sequencing_group = 4
+    if job_attrs.get('sequencing_type') == 'exome':
+        gb_per_sequencing_group = 0.1
+    else:
+        gb_per_sequencing_group = 1.0
+        if not site_only:
+            gb_per_sequencing_group = 1.5
+        
     return gb_per_sequencing_group * sequencing_group_count
--- a/cpg_workflows/resources.py
+++ b/cpg_workflows/resources.py
@@ -338,12 +338,14 @@ def joint_calling_scatter_count(sequencing_group_count: int) -> int:
 
 def storage_for_joint_vcf(
     sequencing_group_count: int | None, site_only: bool = True, job_attrs: dict | None = None
-) -> int | None:
+) -> float | None:
     """
     Storage enough to fit and process a joint-called VCF
     """
     if not sequencing_group_count:
         return None
+    if job_attrs is None:
+        job_attrs = {}
     if job_attrs.get('sequencing_type') == 'exome':
         gb_per_sequencing_group = 0.1
     else:
--- a/cpg_workflows/resources.py
+++ b/cpg_workflows/resources.py
@@ -337,16 +337,14 @@ def joint_calling_scatter_count(sequencing_group_count: int) -> int:
 
 
 def storage_for_joint_vcf(
-    sequencing_group_count: int | None, site_only: bool = True, job_attrs: dict | None = None
+    sequencing_group_count: int | None, site_only: bool = True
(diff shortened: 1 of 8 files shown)


cpg_workflows/resources.py
```diff
@@ -337,14 +337,20 @@ def joint_calling_scatter_count(sequencing_group_count: int) -> int:
 
 
 def storage_for_joint_vcf(
-    sequencing_group_count: int | None, site_only: bool = True
-) -> int | None:
+    sequencing_group_count: int | None, site_only: bool = True, job_attrs: dict | None = None
```
MattWellie, line 340: the job_attrs dict should be fine here, but we only use it for sequencing type which should always be present in `get_config()['workflow']['sequencing_type']` without the need to pass any objects around

MattWellie (approved):
Looks good!

MattWellie:
Could probably do with a bump version as well

cassimons:
These changes look good, but these functions are used by the LC pipeline correct? If so we need to put it past the PA team before merging.

EddieLF:
For the time being I have reverted the changes to the storage requests for VQSR jobs, because these are used by the large_cohort pipeline. Until we can confirm whether or not these changes are suitable for their jobs too then we will just change the resource allocations for the vcf jobs which are only used by the seqr_loader pipeline. I'll add a bumpversion too.

cpg_workflows/resources.py
EddieLF: Good shout

cassimons (approved):

PR #657, opened 2024-03-20
WC -l causing job failures

MattWellie:
See https://batch.hail.populationgenomics.org.au/batches/436642/jobs/373

The final bash section of this job seems to fail when the number of PASS events is 0. There are 2 failing jobs in 2 different batches, both appear to be the only 0-pass samples. I can't really find a good explanation for this.

I can't recreate this failure locally...

```bash
─➤  touch afile.txt
─➤  THIS=$(grep "something" afile.txt | grep foo | wc -l)
─➤  echo $?
0
─➤  echo $THIS
         0
```

The exit code is fine, and the leading spaces are an apple-specific thing (see [here](https://apple.stackexchange.com/questions/370366/why-is-wc-c-printing-spaces-before-the-number))

Here I'm having a go at passing the output through xargs, which should succeed and prevent failure.

### DIFF
--- a/cpg_workflows/jobs/gcnv.py
+++ b/cpg_workflows/jobs/gcnv.py
@@ -418,6 +418,8 @@ def postprocess_calls(
             f"""
         #use wc instead of grep -c so zero count isn't non-zero exit
         #use grep -P to recognize tab character
+        VARIANTS=$(zgrep '^[^#]' {j.output['segments.vcf.gz']})
+        echo "Num Variants: $VARIANTS"
         NUM_SEGMENTS=$(zgrep '^[^#]' {j.output['segments.vcf.gz']} | grep -v '0/0' | grep -v -P '\t0:1:' | grep '' | wc -l)
         NUM_PASS_SEGMENTS=$(zgrep '^[^#]' {j.output['segments.vcf.gz']} | grep -v '0/0' | grep -v -P '\t0:1:' | grep 'PASS' | wc -l)
         if [ $NUM_SEGMENTS -lt {max_events} ]; then
@@ -429,6 +431,7 @@ def postprocess_calls(
         else
             echo "EXCESSIVE_NUMBER_OF_EVENTS" >> {j.qc_file}
         fi
+        cat {j.qc_file}
         """
         )
         get_batch().write_output(j.qc_file, qc_file)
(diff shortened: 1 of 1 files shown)


jmarshall:
I'm still looking at this, but here's a theory:

* `grep foo` will fail if it sees no /foo/ lines
* the snippet probably has `pipefail` set, so this will cause that whole line to fail

(however I haven't reproduced this whole theory locally yet)

Fixing it with `xargs` is pretty obscure, and if my theory is correct won't fix it anyway.

I'd suggest replacing the `grep foo | wc -l` with an `awk` incantation.

MattWellie:
> * `grep foo` will fail if it sees no /foo/ lines
> * the snippet probably has `pipefail` set, so this will cause that whole line to fail

Yup, this was it. I didn't have pipefail set in local testing. I've changed this codeblock to un-set pipefail for this counting section, as a count of 0 is a realistic possibility.

cassimons (approved):
I love the smell of awk in the morning

cpg_workflows/jobs/gcnv.py
```diff
@@ -414,12 +414,13 @@ def postprocess_calls(
...
         max_pass_events = get_config()['workflow']['gncv_max_pass_events']
         # do some additional stuff to determine pass/fail
+        # flake8: noqa
         j.command(
             f"""
         #use wc instead of grep -c so zero count isn't non-zero exit
         #use grep -P to recognize tab character
-        NUM_SEGMENTS=$(zgrep '^[^#]' {j.output['segments.vcf.gz']} | grep -v '0/0' | grep -v -P '\t0:1:' | grep '' | wc -l)
-        NUM_PASS_SEGMENTS=$(zgrep '^[^#]' {j.output['segments.vcf.gz']} | grep -v '0/0' | grep -v -P '\t0:1:' | grep 'PASS' | wc -l)
+        NUM_SEGMENTS=$(zcat {j.output['segments.vcf.gz']} | awk '!/^#/ && !/0\/0/ && !/\t0:1:/ {{count++}} END {{print count}})
```
jmarshall, line 422:
```suggestion
        NUM_SEGMENTS=$(zcat {j.output['segments.vcf.gz']} | awk '!/^#/ && !/0\/0/ && !/\t0:1:/ {{count++}} END {{print count}}')
```

jmarshall (approved):
Dammit flake8 :smile:

LGTM too modulo a quote that the next job failure will probably tell you about too (if I'm reading it right), and the "use wc / grep -P" comments can go now too.

PR #661, opened 2024-03-25
Issue 660 - non-deterministic unpacking of SGID data

### LINKED ISSUE #660: gCNV generating wrong contents
https://batch.hail.populationgenomics.org.au/batches/437477/jobs/1

https://github.com/populationgenomics/production-pipelines/blob/1390a68b31c1bcc4474ab57b6914eb7c75b36db6/cpg_workflows/stages/gcnv.py#L165

MattWellie:
Closes #660 

... see #660 

This creates a new first stage for gCNV, taking the list of all SGIDs in the cohort and writing to a JSON file

Stages in the workflow which have a need to unpack a list of SGIDs in consistent order now read this file and unpack/set indices relative to this fixed order. This JSON of ordered SGIDs is set for the pipeline hash, so new sample list == new sample order.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.22.0
+current_version = 1.22.1
 commit = True
 tag = False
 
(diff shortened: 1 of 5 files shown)


cassimons (approved):
Looks great thanks.

PR #662, opened 2024-03-25
Correct Exomiser find-affected behaviour

MattWellie:
This was tough to test in `test` where nobody has HPO terms (and no-one can hear you scream)


cassimons (approved):
Love test more and more each day...

PR #673, opened 2024-04-05
Remove blacklisted intervals with picard

EddieLF:
So the idea here is that we pass the source intervals file `wgs_calling_regions.hg38.interval_list` to the `get_intervals` function in `picard.py`, alongside the `scatter_count` variable. The function then splits the interval list into scatter_count fragments, which are then each allocated to the various joint calling jobs.

A new function `blacklist_intervals` would take the source intervals file, a blacklist intervals file, and apply the blacklist to the source by using picard `IntervalListTools` with the `SUBTRACT` action, as described in the doc: https://gatk.broadinstitute.org/hc/en-us/articles/360036897212-IntervalListTools-Picard

This would create a source_intervals file with the blacklisted regions removed. This new file could then be handed to get_intervals as normal, alongside the scatter count. 

Essentially, this would add one extra step / job where the intervals file gets modified before its used to create the joint genotyping jobs.

### DIFF
--- a/cpg_workflows/jobs/picard.py
+++ b/cpg_workflows/jobs/picard.py
@@ -18,6 +18,48 @@ from cpg_workflows.resources import (
 from cpg_workflows.utils import can_reuse, exists
 
 
+def blacklist_intervals(
+    b: hb.Batch,
+    job_attrs: dict | None = None,
+    source_intervals_path: Path | None = None,
+    blacklist_intervals_path: Path | None = None,
+) -> Job | None:
+    """
+    Add a job that subtracts the blacklisted intervals from the source intervals.
+
+    @param b: Hail Batch object,
+    @param job_attrs: attributes for Hail Batch job,
+    @param source_intervals_path: path to source intervals to use in joint calling.
+        Would check for config if not provided.
+    @param blacklist_intervals_path: path to blacklist intervals to exclude. 
+        Would check for config if not provided.
+
+    The job calls picard IntervalListTools to subtract the blacklist interval list
+    from the source interval list.
+    """
+    sequencing_type = get_config()['workflow']['sequencing_type']
+    source_intervals_path = source_intervals_path or reference_path(f'broad/{sequencing_type}_calling_interval_lists')
+    blacklist_intervals_path = blacklist_intervals_path or reference_path('broad/blacklist_intervals') #TODO add this to config (default?)
+    
+    j = b.new_job(
+        f'Subtract blacklist intervals before scattering intervals for {sequencing_type} calling',
+        attributes=(job_attrs or {}) | dict(tool='picard IntervalListTools'),
+    )
+    j.image(image_path('picard'))
+    STANDARD.set_resources(j, storage_gb=4, mem_gb=2)
+    
+    cmd = f"""
+    mkdir $BATCH_TMPDIR/out
+
+    picard -Xms1000m -Xmx1500m \
+    IntervalListTools \
+    ACTION=SUBTRACT \
+    INPUT={b.read_input(str(source_intervals_path))} \
+    SECOND_INPUT={b.read_input(str(blacklist_intervals_path))} \
+    OUTPUT=$BATCH_TMPDIR/out/subtracted.interval_list
+    """
+    j.command(command(cmd))
+
 def get_intervals(
     b: hb.Batch,
     scatter_count: int,


cpg_workflows/jobs/picard.py
```diff
@@ -18,6 +18,48 @@
...
+        Would check for config if not provided.
+    @param blacklist_intervals_path: path to blacklist intervals to exclude. 
+        Would check for config if not provided.
+
+    The job calls picard IntervalListTools to subtract the blacklist interval list
+    from the source interval list.
+    """
+    sequencing_type = get_config()['workflow']['sequencing_type']
+    source_intervals_path = source_intervals_path or reference_path(f'broad/{sequencing_type}_calling_interval_lists')
+    blacklist_intervals_path = blacklist_intervals_path or reference_path('broad/blacklist_intervals') #TODO add this to config (default?)
```
MattWellie, line 42: the standard one here would be `reference_path('hg38_telomeres_and_centromeres')`?

MattWellie: once it gets successfully copied...

cpg_workflows/jobs/picard.py
```diff
@@ -18,6 +18,48 @@
...
+    
+    cmd = f"""
+    mkdir $BATCH_TMPDIR/out
+
+    picard -Xms1000m -Xmx1500m \
+    IntervalListTools \
+    ACTION=SUBTRACT \
+    INPUT={b.read_input(str(source_intervals_path))} \
+    SECOND_INPUT={b.read_input(str(blacklist_intervals_path))} \
+    OUTPUT=$BATCH_TMPDIR/out/subtracted.interval_list
```
MattWellie, line 59: this will work for the batch, but might be more future proof if we create an output path incorporating the values of the source and subtracted intervals path - save that into long term storage. That way we've got a permanent record of the A-B result, and future runs with the same intervals won't need to recreate this

MattWellie (commented):
It would be dreamy if an optional blacklist can be supplied to the interval scattering command we already have - I'l not sure from the docs whether SCATTER_COUNT to break up the output and ACTION subtract to sort out the inputs are compatible arguments

cpg_workflows/jobs/picard.py
EddieLF: Yeah good thought. If we can create it once and nicely document that it's the standard `wgs_calling_regions.hg38.interval_list` intervals list with the `telomeres_and_centromeres` list subtracted from it then we can just reference it directly without deriving it each time a batch is run.

EddieLF:
OK so I experimented with the picard interval tools locally.

First convert the telomeres_and_centromeres.bed file into an interval_list file using [`BedToIntervalList`](https://gatk.broadinstitute.org/hc/en-us/articles/360037593251-BedToIntervalList-Picard):
```
java -jar picard.jar BedToIntervalList \
I=hg38.telomeresAndMergedCentromeres.bed \
O=hg38.telomeresAndMergedCentromeres.interval_list \
SD=wgs_calling_regions.hg38.interval_list
```

Then, using the subtract action with the newly created `telomeres_and_centromeres.interval_list` with `IntervalListTools`
```
java -jar picard.jar IntervalListTools \
ACTION=SUBTRACT \
I=wgs_calling_regions.hg38.interval_list \
SI=hg38.telomeresAndMergedCentromeres.interval_list \
O=new.interval_list
```

This command created the attached `interval_list` file, which should be the original source intervals list `wgs_calling_regions.hg38.interval_list` minus the telomeres and centromeres 

[new.interval_list.txt](https://github.com/populationgenomics/production-pipelines/files/14879032/new.interval_list.txt)
(Uploaded with `.txt` extension because GitHub didnt like the `.interval_list` file)

It does seem to have removed the intervals correctly, but it's added a new column? In between the `+` and the `.` in the intervals part of the file, there is now a column with `interval-xx intersection`. 
Not sure if this is impactful. The original interval_list had 3,725 lines, the new one has 3,669 lines.

EddieLF:
At the suggestion of @cassimons, we can just add the `ACTION=SUBTRACT` and `SECOND_INPUT=hg38_telomeres_and_centromeres.interval_list` to the existing `picard.get_intervals()` function. This way, each time the intervals are generated, the telomeres and centromeres are excluded. This requires an addition to the default configs with the path to the intervals file to exclude. 

I've added a very ugly `if` - `else` block to the picard job, which will determine if the command runs with the subtract action and the secondary input, based on whether or not the exclude intervals file is present. There's probably a nicer way to do this.

I've also updated the joint genotyping stage and jobs to take an exclude intervals file as possible inputs. By default this will be none, and only at the picard job will the exclude intervals file be read from the default config `references.hg38_telomeres_and_centromeres`. 

If this general process seems ok, then we will need to get the file at `gs://cpg-common-main/references/hg38/v0/hg38.telomeresAndMergedCentromeres.bed` converted into an `.interval_list` file like I describe in the previous comment.

EddieLF:
Updated with the context that the exclusion interval_list will live at `reference_path('hg38_telomeres_and_centromeres_intervals/interval_list')` once [this is merged](https://github.com/populationgenomics/references/pull/51) - this will also fix the tests failing in this PR.
If the interval_list is present, add in the `-ACTION SUBTRACT` and `-SI {exclude.interval_list}` lines to the `IntervalListTools` command. 

Also updated the command syntax, based on a warning from Picard suggesting the new syntax
```
********** NOTE: Picard's command line syntax is changing.
**********
********** For more information, please see:
**********
https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition)
**********
********** The command line looks like this in the new syntax:
**********
**********    IntervalListTools -ACTION SUBTRACT -SCATTER_COUNT 5 -SUBDIVISION_MODE INTERVAL_SUBDIVISION -UNIQUE true -SORT true -BREAK_BANDS_AT_MULTIPLES_OF 100000 -I wgs_calling_regions.hg38.interval_list -SI hg38.telomeresAndMergedCentromeres.interval_list -OUTPUT ./out
**********
```

cassimons:
Ed, as this is a largish change, can we run this end to end in test validation cohort before merge?

EddieLF:
Batch run on this branch, for 3 WGS samples in validation-test, scatter count = 60, with centromeres and telomeres excluded - https://batch.hail.populationgenomics.org.au/batches/442181
**Took 1h54m**
The outputs were renamed in GCP to preserve them in case we want to compare them with the standard outputs generated with no excluded regions.

Batch run on main branch, same 3 WGS samples in validation-test, scatter count = 60, no excluded regions - https://batch.hail.populationgenomics.org.au/batches/442208
**Runtime TBD**

EddieLF:
## **Change Summary**
 - Added gnomAD v3's hg38 telomeres and centromeres interval_list file to cpg-common-main/references 
 - Updated Picard get_intervals function to subtract the telomeres and centromeres from the source intervals 
   - This affects the intervals used by the stages: `Genotype`, `JointGenotyping`, `VEP`, `VQSR`
 - Updated Picard get_intervals command to use new syntax, and corresponding test functions in `test_picard` module

### Results:
My previous comment links to a batch testing these changes on 3 genomes in `validation-test` for the `JointGenotyping`, `VEP`, `VQSR` stages, before an `AnnotateCohort` mt is created. There was about a 5% reduction in overall time taken to run through these stages.
| Output | [No exclusions](https://batch.hail.populationgenomics.org.au/batches/442208) | [Exclusions](https://batch.hail.populationgenomics.org.au/batches/442181) |
|--------|--------|--------|
| `JointGenotyping/full.vcf.gz` | 560.5 MiB | 546.43 MiB |
|`Vqsr/siteonly.vcf.gz` | 729.18 MiB | 707.46 MiB |
| `Vep/vep.ht/` | 1.13 GiB | 1.11 GiB |
| `AnnotateCohort/cohort.mt/` | 3.06 GiB | 3.02 GiB |

### Next steps
I just realised  the `Genotype` stage during the `haplotype_caller` job hasn't been tested yet. This is now being run, so we can see the changes through a workflow containing all impacted stages `Genotype` -> `JointGenotyping` -> `VEP` -> `VQSR`. See: https://batch.hail.populationgenomics.org.au/batches/442714

Aside from that, I'm not sure how we should actually test these outputs. In the table above, we can see that there is a reduction in size of the outputs when the exclusions are applied. I'm not sure how we can quantify exactly what this means in terms of impact on analysis and variant identification. If we are confident these intervals should be excluded, then so long as the `Genotype` stages run fine in the above batch, these changes can probably be merged.

cassimons:
Hey Ed great work! Provided the full test run ([442724](https://batch.hail.populationgenomics.org.au/batches/442724)) completes successfully, I am happy for this to be merged based on these results. Ideally, we would A/B test with the validation pipeline - but I am not sure that is in working order at the moment and the main thing I wanted to see was it running end to end with a sensible reduction in variants (which you have shown).

🙌

cpg_workflows/jobs/picard.py
MattWellie: I still fully believe that these interval lists should be stored in long term memory, but looking at it now, the output prefixes are all in temp buckets anyway. I'll make an issue, but no reason to hold back this change

MattWellie (approved):
keen4it

PR #677, opened 2024-04-06
Reblock gvcf storage option

EddieLF:
Adds resource_overrides config option for Postproc gvcf job. 

Prompted by [a recent failed job](https://batch.hail.populationgenomics.org.au/batches/440331/jobs/1) which ran out of storage. The HaplotypeCaller output for this sample is really large and it needs more than the default storage to finish the Genotype stage.

### DIFF
--- a/cpg_workflows/jobs/genotype.py
+++ b/cpg_workflows/jobs/genotype.py
@@ -322,7 +322,8 @@ def postproc_gvcf(
     # meaning we have more than enough disk (265/8=33.125G).
     # Enough to fit a pre-reblocked GVCF, which can be as big as 10G,
     # the reblocked result (1G), and ref data (5G).
-    job_res = STANDARD.set_resources(j, ncpu=2, storage_gb=20)
+    storage_gb = get_config()['workflow']['resources']['Genotype'].get('postproc_gvcf_storage', 20)
+    job_res = STANDARD.set_resources(j, ncpu=2, storage_gb=storage_gb)
 
     j.declare_resource_group(
         output_gvcf={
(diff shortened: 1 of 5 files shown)


cassimons (changes requested):
This is a new config correct? If so can you please add it into cpg_workflows/defaults.toml with a description. Just leave it commented out if it is not set by default.

Also, as this stage is used by PA all changes need to be reviewed by them as well so I have added Katie.

cpg_workflows/jobs/genotype.py
```diff
@@ -322,7 +322,11 @@ def postproc_gvcf(
     # meaning we have more than enough disk (265/8=33.125G).
     # Enough to fit a pre-reblocked GVCF, which can be as big as 10G,
     # the reblocked result (1G), and ref data (5G).
-    job_res = STANDARD.set_resources(j, ncpu=2, storage_gb=20)
+    if (postproc_gvcf_storage := get_config()['resource_overrides'].get('postproc_gvcf_storage')) is not None:
```
katiedelange, line 325: Slightly rusty, so ignore me if I'm off here, but I believe this can be simplified to
`storage_gb = get_config()['resource_overrides'].get('postproc_gvcf_storage', 20)`?

(Similar to the scatter_count construction in line 47)

EddieLF: Thanks Katie, this is cleaner that what I had

EddieLF:
@cassimons added the option to the `cpg_workflows/defaults.toml` config + a bumpversion for the deploy

cassimons (approved):
LGTM

PR #678, opened 2024-04-08
Remove metrics from GB workflow

MattWellie:
Metrics are not required during standard running - off by default.

Also moves these standard overrides into the workflow default - the override will become running, instead of disabling

### DIFF
--- a/configs/defaults/gatk_sv_multisample_2.toml
+++ b/configs/defaults/gatk_sv_multisample_2.toml
@@ -54,3 +54,7 @@ username = 'seqr'
 # Load ElasticSearch password from a secret, unless SEQR_ES_PASSWORD is set
 password_secret_id = 'seqr-es-password'
 password_project_id = 'seqr-308602'
+
+[resource_overrides.MakeCohortVcf]
+run_module_metrics = false
+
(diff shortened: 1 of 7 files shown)


cpg_workflows/stages/gatk_sv/gatk_sv_multisample_1.py
```diff
@@ -554,6 +554,12 @@ def expected_outputs(self, cohort: Cohort) -> dict:
             'metrics_file_genotypebatch': 'metrics.tsv',
         }
 
+        # if we don't run metrics, don't expect the outputs
+        # on by default in the WDL file, so expect this to run unless overridden
+        if override := get_config()['resource_overrides'].get('MakeCohortVcf'):
```
EddieLF, line 559:
```suggestion
        if override := get_config()['resource_overrides'].get('GenotypeBatch'):
```

EddieLF (approved):
perfect

configs/defaults/gatk_sv_multisample_2.toml
```diff
@@ -54,3 +54,7 @@ username = 'seqr'
 # Load ElasticSearch password from a secret, unless SEQR_ES_PASSWORD is set
 password_secret_id = 'seqr-es-password'
 password_project_id = 'seqr-308602'
+
+[resource_overrides.MakeCohortVcf]
+run_module_metrics = false
+
```
EddieLF, line 60:
```suggestion
```

PR #679, opened 2024-04-09
Remove startup config-using variables

MattWellie:
Removes config access on module import - replaced with lru_cache methods which will return the same content (generated once per run startup)

Replaces the logic to determine which Datasets have ES indexes generated with a try-except to fetch the secret as a string. Datasets without the required config entry or permission to read the secret will fail, exception caught, and the stage will be skipped for that dataset only


MattWellie:
I'll try and shove some fixes for the failure-to-find-seqr-secrets too

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
```diff
@@ -45,12 +42,23 @@
...
-EXCLUSION_FILE = join(
-    get_config()['storage']['default']['default'],
-    'gatk_sv',
-    RUN_DATETIME,
-    'combined_exclusion_list.txt',
-)
+
+
+@lru_cache(maxsize=1)
+def get_exclusion_filename() -> str:
```
cassimons, line 48: This feels heavy. Will this be an unusual usage or are we going to end up with needing to define getters everywhere? If so could we bake in some tooling to get_config somehow to handle this situation?

MattWellie: The way that I'd written this initially (plus the mito changes) was the only instance of generating a variable at the top level of a file. A getter isn't really necessary, it's just building a file path so it could be done manually instead of calling this method.

John was having issues running local tests, given that this makes some assumptions about the config content, and this ran when the module was imported in main.py. This should have been fine with a config stub, but it's still rough design putting this variable assignment where it was going to be generated immediately even if it's never used.

cassimons (commented):
I am happy to approve this if it is the direction we want to go, but to be honest I don't like the idea of having to define getters everywhere - just feels a bit ick.

Are there any other options to skin this cat (eg can we make get_config lookups lazy somehow?). @jmarshall thoughts?

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
jmarshall: Most of these things just build up these filenames inside the methods, so the `get_config()` call doesn't happen until the particular pipeline is run. But these ones have it as a variable at the top-level, so the calls happen at import time in _main.py_ — which explodes for me and others on local tests. This is what Cas and I talked about in _mito.py_ many months ago, but you and I then forgot about it again :smile:

I guess these two files have them as top-level variables because the (sets of) filenames are wildly used so it's easier than repeating the definitions…

jmarshall:
TBH I don't love it either.

I suppose another approach would be to derive the stage classes from a new class that has the config-based paths as class variables. Either by interposing it between CohortStage and CombineExclusionLists etc, or by using multiple inheritance (ick).

Might be worth a try…

cassimons:
I know we only do this a couple of times in the codebase atm so handling them as edge cases is fine and we can just merge this to make the issue go away.

But in the longer term setting constants from config at the top level seems a reasonable pattern to support so it would be nice to think of a clean way to allow this. Or is the problem better solved at the other end and we just need better support for stubbing out configs that can be used locally without causing start-up issues? 

Either way, we probably need some test to detect this across the repo so it just does not pop up when someone attempts to run locally.

jmarshall:
You're quite right that setting constants from the constants in config and doing so at the top level is a pretty reasonable pattern.

This works with the default _default.toml_ and the CPG_CONFIG_PATH settings, because those get activated before _main.py_ imports the various stages and sets up `WORKFLOWS` — because mostly by luck they get activated when `from cpg_workflows import defaults_config_path` happens, which is above there.

But it doesn't work with `main.py … --config foo.toml`, because that doesn't get added to the path until later. So you don't even have a chance to put the settings that SV2 and mito need in _foo.toml_, because it gets read too late.

Holy cow 💡 

What if we just move the imports and `WORKFLOWS` down to inside `def main(…)` so that **they** happen later?!

MattWellie:
> But it doesn't work with `main.py … --config foo.toml`, because that doesn't get added to the path until later.

I think this is missing the difference between test running and prod running. In practice we pass the config argument to analysis runner, not to main.py. The value set as CPG_CONFIG... is already a combined config, so IMO in test runs we should also be making that contain all combined elements

jmarshall:
So you're saying remove `@click.option('--config')` because it doesn't work. And happily production runs via analysis runner don't use it.

MattWellie:
Sounds rational to me. With the default config, workflow default, analysis-runner provided default, AND this there's just a wild number of seperate config entry points

jmarshall:
> I think this is missing the difference between test running and prod running.

In production practice, we pass the config argument to analysis runner, not to main.py. But in (some forms of) local test running, we don't use analysis runner. In those use cases, the `--config` option is a useful facility.

It turns out to be easy to make it work with small changes in _main.py_ and no adjustments needed to the top-level constants in _gatk_sv_multisample_2.py_ and _mito.py_. Stand by for an imminent alternative PR…

cassimons:
Sounds like a good solution to me.

PR #680, opened 2024-04-09
Remove pylint switches & comments

MattWellie:
Independent issue - we've removed pylint as a tool, but not removed the relevant pylint comments


vivbak (approved):
Thanks @MattWellie!

PR #683, opened 2024-04-09
Change Docker build

MattWellie:
# Dockerfile upgrade

This is a low priority change, but would produce a neater build:

## Issue

The current Dockerfile is:

```
FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest

RUN pip install metamist
COPY README.md .
COPY setup.py .
COPY cpg_workflows cpg_workflows
COPY seqr-loading-pipelines/hail_scripts hail_scripts
COPY gnomad_methods/gnomad gnomad
RUN pip install .
```

- We're building on a non-specific version of the driver image, it would be more specific to update that from `latest` to a specific tag. I'm happy to stick with latest as that is pretty low friction to bring through new updates, and we don't update the driver image frequently. If we move to versioned releases, this should be a proper pin.
- Runs an un-pinned install of metamist, which the driver image already contains. Guaranteed to do nothing.
- Copies in code files and submodules, inc. the whole cpg_workflows codebase. This will invalidate the Docker cache, meaning that all layers after this are guaranteed to run.
- Right at the end we do a pip install of the package, including all dependencies. This is the longest running part of the build, and will be repeated every time.

## Proposal

- Add all requirements to the `requirements.txt` and `-dev` files, including pinned versions
- These files are parsed by `setup.py`, so an environment setup is the same using either file.
- Copy these files into the image first, and install all dependencies.
- Then copy the code and run a pip install.

reasoning: 

- Dependencies will change less often than the code, so we preserve more layers from previous builds by installing dependencies prior to copying in code.
- This keeps parity between the requirements.txt file and the setup.py install, a developer can install all requirements without installing the package specifically.
- The setup.py file contains a version number, so it _will_ change with each commit. By installing requirements from the setup.py file we force a reinstall, generating a new docker image layer, even if the requirements are unchanged.

I don't think this is a massive change, none of these layers are particularly long-running... but each new layer we create in the artifactory has an attached storage cost, and I think this produces a faster-running, more storage conscious image result.


vivbak (commented):
Happy with this in general, once everything stops having a whine!

MattWellie:
> Happy with this in general, once everything stops having a whine!

NGL I'm impressed I failed all 3 workflow runs

MattWellie:
n.b. not much of this will be super meaningful; the docker cache isn't shared well on GCP nodes, and the build cache probably isn't used well on github CI servers either, but I think this is still a rational thing to do.

PR #687, opened 2024-04-11
Delay cpg_workflows.stages.* imports until after main() starts

jmarshall:
Some of our workflow stages define top-level constants based on config entries accessed via `get_config()` — which is quite reasonable as these are indeed **constants**, and especially where an entry is used multiple times in a pipeline, defining it once at the top-level is a very appropriate pattern.

However this means that config files set via _main.py_'s `--config` option aren't accessible, because the config has already been read during the imports, before `main()` has run. We don't use this in production (where configs are set via analysis runner), but it is a useful knob when dry-run testing locally.

Fix this by moving the stage imports into a new `get_workflows()` function, which delays importing them until the function is called, and finalise the config path before calling `get_workflows()` to initialise `WORKFLOWS`. (~~The config handling now needs an extra wart to handle the existing `list` special case.~~ This now removes the previous `list` special case, which was just an obscure way of saying `--list-workflows`.)

This makes (most of) PR #679 unnecessary.

You're going to want to view this PR's changes with the “Hide whitespace” setting on!

### DIFF
--- a/main.py
+++ b/main.py
@@ -11,54 +11,57 @@ import coloredlogs
 from cpg_utils import to_path
 from cpg_utils.config import set_config_paths
 from cpg_workflows import defaults_config_path
-from cpg_workflows.stages.aip import CreateAIPHTML, GenerateSeqrFile, ValidateMOI
-from cpg_workflows.stages.cram_qc import CramMultiQC
-from cpg_workflows.stages.exomiser import ExomiserSeqrTSV, RunExomiser
-from cpg_workflows.stages.fastqc import FastQCMultiQC
-from cpg_workflows.stages.fraser import Fraser
-from cpg_workflows.stages.gatk_sv.gatk_sv_multisample_1 import (
-    FilterBatch,
-    GenotypeBatch,
-    MergeBatchSites,
-)
-from cpg_workflows.stages.gatk_sv.gatk_sv_multisample_2 import MtToEsSv
-from cpg_workflows.stages.gatk_sv.gatk_sv_single_sample import CreateSampleBatches
-from cpg_workflows.stages.gcnv import AnnotateCohortgCNV, AnnotateDatasetCNV, MtToEsCNV
-from cpg_workflows.stages.gvcf_qc import GvcfMultiQC
-from cpg_workflows.stages.happy_validation import (
-    ValidationHappyOnVcf,
-    ValidationMtToVcf,
-    ValidationParseHappy,
-)
-from cpg_workflows.stages.large_cohort import AncestryPlots, Frequencies, LoadVqsr
-from cpg_workflows.stages.mito import MitoReport
-from cpg_workflows.stages.outrider import Outrider
-from cpg_workflows.stages.seqr_loader import AnnotateDataset, DatasetVCF, MtToEs
-from cpg_workflows.stages.stripy import Stripy
 from cpg_workflows.workflow import StageDecorator, run_workflow
 
-WORKFLOWS: dict[str, list[StageDecorator]] = {
-    'aip': [ValidateMOI, CreateAIPHTML, GenerateSeqrFile],
-    'exomiser': [RunExomiser, ExomiserSeqrTSV],
-    'pre_alignment': [FastQCMultiQC],
-    'seqr_loader': [
-        DatasetVCF,
-        AnnotateDataset,
-        MtToEs,
-        GvcfMultiQC,
-        CramMultiQC,
-        Stripy,
-        MitoReport,
-    ],
-    'validation': [ValidationMtToVcf, ValidationHappyOnVcf, ValidationParseHappy],
-    'large_cohort': [LoadVqsr, Frequencies, AncestryPlots, GvcfMultiQC, CramMultiQC],
-    'gatk_sv_singlesample': [CreateSampleBatches],
-    'gatk_sv_multisample_1': [FilterBatch, GenotypeBatch],
-    'gatk_sv_sandwich': [MergeBatchSites],  # stage to run between FilterBatch & GenotypeBatch
-    'gatk_sv_multisample_2': [MtToEsSv],
-    'rare_disease_rnaseq': [Outrider, Fraser],
-    'gcnv': [AnnotateCohortgCNV, AnnotateDatasetCNV, MtToEsCNV],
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -11,54 +11,57 @@
...
+    )
+    from cpg_workflows.stages.gatk_sv.gatk_sv_multisample_2 import MtToEsSv
+    from cpg_workflows.stages.gatk_sv.gatk_sv_single_sample import CreateSampleBatches
+    from cpg_workflows.stages.gcnv import AnnotateCohortgCNV, AnnotateDatasetCNV, MtToEsCNV
+    from cpg_workflows.stages.gvcf_qc import GvcfMultiQC
+    from cpg_workflows.stages.happy_validation import (
+        ValidationHappyOnVcf,
+        ValidationMtToVcf,
+        ValidationParseHappy,
+    )
```
MattWellie, line 36: I think these shrunk to one line, with the 120 char black limits

main.py
```diff
@@ -11,54 +11,57 @@
...
+    from cpg_workflows.stages.mito import MitoReport
+    from cpg_workflows.stages.outrider import Outrider
+    from cpg_workflows.stages.seqr_loader import AnnotateDataset, DatasetVCF, MtToEs
+    from cpg_workflows.stages.stripy import Stripy
+
+    return {
+        'aip': [ValidateMOI, CreateAIPHTML, GenerateSeqrFile],
+        'exomiser': [RunExomiser, ExomiserSeqrTSV],
+        'pre_alignment': [FastQCMultiQC],
+        'seqr_loader': [
```
MattWellie, line 47: maybe here too?

main.py
```diff
@@ -109,6 +112,24 @@ def main(
     fmt = '%(asctime)s %(levelname)s (%(name)s %(lineno)s): %(message)s'
     coloredlogs.install(level='DEBUG' if verbose else 'INFO', fmt=fmt)
 
+    if workflow and workflow != 'list':
+        wfl_conf_path = to_path(__file__).parent / f'configs/defaults/{workflow}.toml'
+        assert wfl_conf_path.exists(), wfl_conf_path
+        wfl_conf_path = [str(wfl_conf_path)]
+    else:
+        wfl_conf_path = []
```
MattWellie, line 120: It's not really part of this change, but I only just noticed this... Why is accepting `workflow=='list'` an option? There's a specific `--list-workflows` argument, so we're supporting this behaviour in two different ways?

Can we change the order of execution to:

1. load args
2. if list-workflows: do that, then quit
3. if list-last-stages: do that, then quit
4. all other behaviours require a valid workflow to be set

jmarshall: Yes, it's ridiculous isn't it.

What I did was make the smallest change that doesn't change behaviour. One could make a slightly less small change that's more elegant, or remove the `list` special case.

:shrug: Probably just remove the special case. Surely no `--list-*` options are used in production, and I for one just look at the code in _main.py_ to see what's available, 'cos that's less painful than running it!

MattWellie: I think the logic for list-workflows/list-endpoints is interesting, but it anticipates someone might pick up this codebase and use them to have a casual browse of the features. 

In reality the users of this pipeline are also developers of it, with a detailed knowledge of the workflows...

And also list-workflows also prints the last stages, so I'm not sure what utility having two options aimed at the same use-case was supposed to provide. I'd be OK with dropping both, but I'll make an issue for it and it can be a 10 second 'any of y'all use this option' discussion at the next prod-pipes meeting

jmarshall: I think if you make an issue about whether we should have these options, that will already be 10x how much attention they are worth. IMHO we should just leave these options' half dozen lines as they are and forget about them again, as they're not hurting anything.

jmarshall: Cleaned up the code a bit by removing the particularly ridiculous `list` special case.

MattWellie:
This mitigates the issue solved in #679, but doesn't really solve it. Top-level imports in optional modules are still a bad design choice, so I think we should consider both changes.

MattWellie:
Ok, I've thought about this a bit more and I'm not a fan -

- in production running it's crucial that the driver job and child jobs can all have the same config. This doesn't support that use case: unlike analysis-runner, which combines config, writes it to GCP, then sets the result as an environment variable for the duration of the run ([here](https://github.com/populationgenomics/analysis-runner/blob/30efba7299e481fd734d4142691317fec972cf46/server/main.py#L145), [env](https://github.com/populationgenomics/analysis-runner/blob/30efba7299e481fd734d4142691317fec972cf46/server/main.py#L224), we pass this to child jobs through `copy_common_env`), this change allows for any additional configs to be merged, and the results will be discarded after the driver runtime concludes, so it should not be permitted in production where we're aiming for reproducibility. My objection to this PR is that the pipeline is designed for cloud infrastructure, and this solution is a workaround to make it work locally - IMO the tests should emulate the way this works in production rather than find work arounds.
- This approach also builds in a PEP8 violation ([here](https://docs.astral.sh/ruff/rules/import-outside-top-level/)) which we aren't catching because the rule is not stable within Ruff, so it's not usually applied.

For local testing purposes (in AIP) I've used [conftest.py](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/test/conftest.py#L21) to load a specific test config prior to test execution, which is a solid way of providing a config stub, but I think that could still fail in the current situation where variables are created from config during initial import (which #679 aims to treat). I'm still not 100% clear if the issue we're trying solve here is running tests locally, or something else.

jmarshall:
Once again: This PR fixes a currently extant production-pipelines bug, namely that `--config` doesn't work.

Your complaint is that the `--config` option, added in the first month of prod-pipes development, shouldn't exist — because it would not be appropriate to use it in production. Please try to be clear that, for the most part, what you are not a fan of is this small piece of original _main.py_ functionality, not this PR per se.

The issue we're trying to solve here is what I would think is a common user story:

> As a pipeline developer, while developing a new prod-pipes pipeline, I'd like to dry-run it locally to exercise my new Python code and see the jobs and commands generated, in a quick-turnaround development environment.

To date, I have used `main.py --config …` for that use case. However the option can be given up as broken and setting `CPG_CONFIG_PATH` used instead, and I for one have lost interest in this conversation.

PR #693, opened 2024-04-17
added environment attr to the docker workflow

nevoodoo:
## Description
In a bid to introduce better security permissions, we have replaced the use of SA keys to workload identity providers. This change will also implement the use of environments to ensure deployments are going through the right channels.

## Changes
- [x]  Added `environment` attribute to the `docker` workflow

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -20,6 +20,7 @@ env:
 jobs:
   docker:
     runs-on: ubuntu-latest
+    environment: production
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


illusional (changes requested):
For this one, and the images repo, we'll need to come up with some way that people can run this workflow and push an unreviewed image to a the images-dev repo. Maybe we need to create a second set of credentials `gh-images-dev-deployer` that can be activated instead?

What do you think?

nevoodoo:
> For this one, and the images repo, we'll need to come up with some way that people can run this workflow and push an unreviewed image to a the images-dev repo. Maybe we need to create a second set of credentials `gh-images-dev-deployer` that can be activated instead?
> 
> What do you think?

Hmmm, I think we could potentially add their branch to the protected branches set temporarily for the push as a work around perhaps?

illusional (approved):
Looks great to me!! FYI @vivbak, @jmarshall, @MattWellie

MattWellie:
Delicious

PR #697, opened 2024-04-18
Standalone PCA dataproc runner

hopedisastro:
Basic PCA runner that 
- subsets the VDS to the sites table 
- runs hwe_normalized_pca

option to specify sample file to filter sample IDs in the VDS prior to running PCA

outputs the loadings, scores, and eigenvalues as TSV, TSV, and txt files respectively. 

Interim solution to get some plots going; will delete scripts after filtering is incorporated into the LC pipeline.

### DIFF
--- a/cpg_workflows/dataproc_scripts/pca_runner.py
+++ b/cpg_workflows/dataproc_scripts/pca_runner.py
@@ -1,0 +1,39 @@
+import csv
+
+import hail as hl
+from cpg_utils.hail_batch import output_path
+from cpg_utils import to_path
+
+
+def pca_runner(vds_path, sample_id_file_path):
+    vds = hl.vds.read_vds(str(vds_path))
+    mt = vds.variant_data
+
+    mt = mt.annotate_entries(GT=hl.vds.lgt_to_gt(mt.LGT, mt.LA))
+    if sample_id_file_path:
+        with to_path(str(sample_id_file_path)).open('r') as f:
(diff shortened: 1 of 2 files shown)


hopedisastro:
This should run ok in `main` - error we are seeing in Batch: https://batch.hail.populationgenomics.org.au/batches/443935/jobs/2 is because we're requesting it to compute 10 components with 5 samples in test (reports 2 samples found because I tested the filtering samples option). PCA needs at least as many samples as the number of components to be computed.

cpg_workflows/dataproc_scripts/pca_runner.py
```diff
@@ -0,0 +1,40 @@
...
+        mt = mt.filter_cols(hl.literal(sample_ids).contains(mt.s))
+
+    # subset to sites table
+    sites_table = "gs://cpg-common-main/references/ancestry/pruned_variants.ht"
+    qc_variants_ht = hl.read_table(sites_table)
+    qc_variants_ht = qc_variants_ht.key_by('locus')
+
+    mt = mt.filter_rows(hl.is_defined(qc_variants_ht[mt.locus]))
+
+    mt = mt.checkpoint(output_path('checkpoint.mt', 'tmp')
```
hopedisastro, line 26:
```suggestion
    mt = mt.checkpoint(output_path('checkpoint.mt', 'tmp'))
```

PR #698, opened 2024-04-18
Cleans up list workflows

### LINKED ISSUE #688: Wrangle list-workflows CLI arguments
For a prod-pipes discussion - 

There are a few [arguments](https://github.com/populationgenomics/production-pipelines/blob/main/main.py#L74-L85) and [special cases](https://github.com/populationgenomics/production-pipelines/blob/main/main.py#L114-L118) for printing the workflows currently enabled inside cpg-workflows, and/or the terminal stages of those workflows. AFAIK this isn't used, and we should probably just remove the special case handling for `workflow == 'list'` etc.

MattWellie:
Closes #688 (If this is not required, #688 should also be rejected as won't do. Feel free to, I promise I won't cry)

Changes the order of main.py argument processing:

- if list_workflows:
  - List workflows (and end points), then quit
  
- list_workflows doesn't require the `workflow` param to be set, but it should be set for all other purposes, so after this return, assert a valid workflow is named

- if list_last_stages:
  - list last stages for the current workflow (this feels redundant with list_workflows already providing this information), then quit

- For all other uses, assert that a valid workflow is defined, and continue as usual

This is a super low priority change, I don't think either of these flags are used by anybody, but maintaining a special case of `--workflow list` is odd.


MattWellie:
This will be discarded when we fracture off individual workflows

PR #699, opened 2024-04-18
Specify columns to be dropped after unioning background datasets

jmarshall:
Another experiment to fix unioning differing background datasets, as a followup to PR #685:

For data generated at different times, the tables being unioned in `add_background()` may have slightly differing sets of columns. Using `unify=True` continues to a final union table with mismatched columns a mixture of values and missing.

This may confuse later processing, so add another config item to enable configuring such columns to be dropped from the final table.

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -24,6 +24,7 @@ def add_background(
     """
     sites_table = get_config()['references']['ancestry']['sites_table']
     allow_missing_columns = get_config()['large_cohort']['pca_background'].get('allow_missing_columns', False)
+    drop_columns = get_config()['large_cohort']['pca_background'].get('drop_columns')
     qc_variants_ht = hl.read_table(sites_table)
     dense_mt = dense_mt.select_cols().select_rows().select_entries('GT', 'GQ', 'DP', 'AD')
     for dataset in get_config()['large_cohort']['pca_background']['datasets']:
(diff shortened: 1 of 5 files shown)


cpg_workflows/large_cohort/ancestry_pca.py
```diff
@@ -57,6 +58,9 @@ def add_background(
         # combine dense dataset with background population dataset
         dense_mt = dense_mt.union_cols(background_mt)
         sample_qc_ht = sample_qc_ht.union(ht, unify=allow_missing_columns)
+
+    if drop_columns:
+        sample_qc_ht = sample_qc_ht.drop(drop_columns)
```
michael-harper, line 63: I think `get_config()['large_cohort']['pca_background'].get('drop_columns')` returns the list (i.e. `['autosomal_mean_dp']`).

If this is the case then `sample_qc_ht.drop()` will need unpacking (see [here](https://hail.is/docs/0.2/hail.MatrixTable.html#hail.MatrixTable.drop))

So should become:
`sample_qc_ht = sample_qc_ht.drop(*drop_columns)`

```suggestion
        sample_qc_ht = sample_qc_ht.drop(*drop_columns)
```

michael-harper (commented):
Just one change for `.drop()`

cpg_workflows/large_cohort/ancestry_pca.py
jmarshall: Thanks; you're right, of course. I misread that documentation as saying it would DWIM if given a list…

michael-harper (approved):

PR #701, opened 2024-04-22
Debug add_background ancestry

michael-harper:
The `background_vds` comprising `hgdp-1kg` background variant and reference information is in excess of 1.6TiB and possibly causing OOM issues when running in Dataproc (though the logs are unable to confirm this is the issue). Checkpointing after calling `to_dense_mt()` has remedied the issue. Also added a checkpoint before calling `add_background()` to check if background has already been added to datasets.

Checkpointing to the `tmp` directory.

For context, see unsuccessful batch [here](https://batch.hail.populationgenomics.org.au/batches/443967/jobs/2) and successful batch run in test [here](https://batch.hail.populationgenomics.org.au/batches/444291/jobs/2) (though still with `background_vds` copied from main to test).

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -1,5 +1,6 @@
 import logging
 import pickle
+from random import sample
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


KatalinaBobowik (approved):
LGTM

PR #709, opened 2024-04-29
Add status reporting to default configs

MattWellie:
I previously had the status reporting flag added to `use_for_all_workflows.toml`, which was necessary in all GATK/gCNV workflows.

Related to https://github.com/populationgenomics/images/pull/137 I stopped using that config file for gCNV, so results have not been registered in metamist.

This change moves the status reporter flag into default config files for all GATK-SV and gCNV workflows.


EddieLF (approved):
➕ 👍

PR #712, opened 2024-04-29
High memory workers in Combiner stage of Large Cohorts

michael-harper:
The Combiner stage was recently taken off of Dataproc (https://github.com/populationgenomics/production-pipelines/pull/621). 

Outputs from this stage when not using Dataproc were tested for consistency, however only on test-level data. Running on main has produced OOM issues that may be remedied by using high memory workers. Currently, there is a config parameter `highmem_workers` available but is only retrieved when setting up a [dataproc_job](https://github.com/populationgenomics/production-pipelines/blob/a2cdfa4de9d6b178fd285eaca0114ddec1bddb26/cpg_workflows/large_cohort/dataproc_utils.py#L103). This PR retrieves the same parameter for use in the `query_command` function that consequently uses the parameter as a batch override via `init_batch`.

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -31,6 +31,8 @@ class Combiner(CohortStage):
 
         j = get_batch().new_job('Combiner', (self.get_job_attrs() or {}) | {'tool': 'hail query'})
 
+        init_batch_args = {'worker_memory': 'highmem'} if get_config()['workflow'].get('highmem_works') else None
+
         j.image(image_path('cpg_workflows'))
         j.command(
             query_command(
@@ -39,6 +41,7 @@ class Combiner(CohortStage):
                 str(self.expected_outputs(cohort)),
                 str(self.tmp_prefix),
                 setup_gcp=True,
+                init_batch_args=init_batch_args,
             ),
         )
         return self.make_outputs(cohort, self.expected_outputs(cohort), [j])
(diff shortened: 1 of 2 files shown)


hopedisastro:
prev. discussion on QoB OOMing out: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1707187236506529

hopedisastro:
Alternative solution - increase driver cores in init_batch(): 
https://github.com/populationgenomics/tob-wgs/pull/102
https://centrepopgen.slack.com/archives/C018KFBCR1C/p1658811940454569?thread_ts=1658208231.938539&cid=C018KFBCR1C

cpg_workflows/stages/large_cohort.py
```diff
@@ -31,6 +31,8 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
 
         j = get_batch().new_job('Combiner', (self.get_job_attrs() or {}) | {'tool': 'hail query'})
 
+        init_batch_args = {'worker_memory': 'highmem'} if get_config()['workflow'].get('highmem_works') else None
```
MattWellie, line 34: do you want to add `highmem_works` to the default large_cohort config?

MattWellie (approved):

cpg_workflows/stages/large_cohort.py
vivbak: ^ second this

vivbak (approved):

cpg_workflows/stages/large_cohort.py
michael-harper: Added!

PR #713, opened 2024-04-29
Run Exomiser 14

MattWellie:
This implements Exomiser 14, married up with these PRs.

- https://github.com/populationgenomics/references/pull/55
- https://github.com/populationgenomics/images/pull/142

Based on developer input we're dropping CADD and REMM for default running, though code to implement these is still present in the exomiser_13 method if we need the template for the future. I've segregated the input and output folders for exomiser 13/14 so that we don't accidentally mix up results.

Test run:
https://batch.hail.populationgenomics.org.au/batches/444875/jobs/1

### DIFF
--- a/cpg_workflows/jobs/exomiser.py
+++ b/cpg_workflows/jobs/exomiser.py
@@ -155,7 +155,7 @@ def make_phenopackets(family_dict: dict[str, list[SequencingGroup]], out_path: d
         # arbitrarily select a proband for now
         proband = affected.pop()
 
-        hpo_term_string = proband.meta['phenotypes'].get(HPO_KEY, '')
+        hpo_term_string = proband.meta['phenotypes'].get(HPO_KEY, 'HP:0000520')
 
         hpo_terms = hpo_term_string.split(',')
 
--- a/cpg_workflows/jobs/exomiser.py
+++ b/cpg_workflows/jobs/exomiser.py
@@ -329,7 +329,7 @@ def run_exomiser_14(content_dict: dict[str, dict[str, Path | dict[str, Path]]]):
         job.image(image_path('exomiser_14'))
 
         # unpack references, see linux-install link above
-        job.command(f'unzip {core_group} {pheno_group} -d {exomiser_dir}/data')
+        job.command(f'unzip "{core_group}" "{pheno_group}" -d "{exomiser_dir}/data"')
 
         job.command(f'echo "This job contains families {" ".join(family_chunk)}"')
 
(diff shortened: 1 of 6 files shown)


cpg_workflows/jobs/exomiser.py
```diff
@@ -166,10 +162,20 @@ def make_phenopackets(family_dict: dict[str, list[SequencingGroup]], out_path: d
...
 
 
-def run_exomiser_batches(content_dict: dict[str, dict[str, Path | dict[str, Path]]]):
+def run_exomiser_13(content_dict: dict[str, dict[str, Path | dict[str, Path]]]):
     """
-    run the exomiser batch
-    what a wild type hint
+    run the families through exomiser 13
+    this is a bit of a monster, but it's a good example of how to chunk jobs
+    that previous line was written by GitHub CoPilot
```
vivbak, line 169: hahahahah

cpg_workflows/jobs/exomiser.py
```diff
@@ -166,10 +162,20 @@ def make_phenopackets(family_dict: dict[str, list[SequencingGroup]], out_path: d
...
+    This version implements Exomiser 13, with default references + CADD and REMM
+    This is not recommended by the developers of Exomiser, but if we decide to
```
vivbak, line 171: all of your comments are just making me laugh

cpg_workflows/stages/exomiser.py
```diff
@@ -93,8 +94,10 @@ class CreateFamilyVCFs(DatasetStage):
 
     def expected_outputs(self, dataset: Dataset) -> dict[str, Path]:
         family_dict = find_families(dataset)
+        exomiser_version = config_retrieve(['workflow', 'exomiser_version'], 14)
```
vivbak, line 97: Would it be more legible to pull the magic number here and declare the version at the top of the file?

vivbak (approved):
LGTM!

cpg_workflows/jobs/exomiser.py
```diff
@@ -79,6 +74,7 @@ def family_vcf_from_gvcf(family_members: list[SequencingGroup], out_path: str) -
     # -0 to replace missing with WT (potentially inaccurate, but predictable parsing in exomiser)
     # --threads 4 to use 4 threads
     # -Oz to write a compressed VCF
+    # type: ignore
```
EddieLF, line 77: Is this suppressing some mypy nonsense? I thought `# type: ignore` was to ignore mypy errors on a specific line

cpg_workflows/stages/exomiser.py
MattWellie: I'm conscious not to introduce any more on-startup config reading, like the ones potentially solved [here](https://github.com/populationgenomics/production-pipelines/pull/691). Putting in a global method or a cached getter to retrieve this would reduce the number of times the value is grabbed, but it's just a dict value with no processing so I'm going with the inelegant but simple solution...

Unless that's me misunderstanding your comment

cpg_workflows/jobs/exomiser.py
MattWellie: It's confusing mypy nonsense that I'm only getting locally, the CI passes here... I'm removing this line

Issue #717, opened 2024-05-02
Remove hefty analysis meta


MattWellie:
Slack: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1714624645731339

## TL; DR

We're shoving a lot of stuff into analysis entry `meta`, and it's making metamist really unhappy

### Hefty Analyses

@illusional has identified some MASSIVE analysis entries. Part of the reason appears to be sections of code where we transcribe entire QC reports into the analysis meta:

- https://github.com/populationgenomics/production-pipelines/blob/7a3a1c1349b236a6df64f42c3b0ffb4dca1a910c/cpg_workflows/stages/joint_genotyping_qc.py#L113-L120
- https://github.com/populationgenomics/production-pipelines/blob/7a3a1c1349b236a6df64f42c3b0ffb4dca1a910c/cpg_workflows/stages/cram_qc.py#L216-L223
- https://github.com/populationgenomics/production-pipelines/blob/7a3a1c1349b236a6df64f42c3b0ffb4dca1a910c/cpg_workflows/stages/gvcf_qc.py#L99

So... we should stop doing that...

### SG IDs

We take all the SG IDs relevant to an analysis and write them into the analysis meta. This is done for all Cohort and Dataset stages:

- [Cohort Stage](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/targets.py#L204)
- [Dataset Stage](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/targets.py#L389)

These methods are called as part of the analysis entry creation process: [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/workflow.py#L589)

So... stop doing that as well. If we need SG IDs to be present, there is already a top-level field `sequencing_group_ids` designed to hold that information. Once we fully adopt custom cohorts we can accomplish this using the Custom Cohort ID/Name, so that these will be super lightweight. The Analysis entry already has an attribute ready for cohort_ids: [here](https://github.com/populationgenomics/metamist/blob/4ae0883d5031f9d49d3edecce24e4f887e1d431e/models/models/analysis.py#L28)

illusional:
FWIW, the analysis object can already be linked against multiple cohort_ids, so also best to avoid putting these in the meta: 
https://github.com/populationgenomics/metamist/blob/4ae0883d5031f9d49d3edecce24e4f887e1d431e/models/models/analysis.py#L98-L106
[...]

PR #718, opened 2024-05-02
Issue 717 analysis meta

MattWellie:
Partially Addresses #717 

Sorry y'all, tagging you both on a fairly early stage PR. 2 commits here (so far):

1. removes the transcription of full QC reports into the meta
2. pops the SG ID list off before writing the analysis entry meta - this is done right at the end, just in case SG IDs are re-populated by any of the update_meta callables. The job attribute dicts will still contain the SG IDs, but they might be used for billing purposes and I don't want to interfere with that. 

Shoved in a bunch of get_config -> config_retrieve with defaults. Unrelated to this issue, but y'know... Passes the time


MattWellie:
Pertinent to #728, all the `_update_meta` methods which directly create a CloudPath instance are flagged as MyPy failures

```
Cannot instantiate abstract class
"CloudPath" with abstract attributes "drive", "is_dir", "is_file", "mkdir" and
"touch"
```

fixable, ignorable, but if we're going to stop doing this anyway we should probably delete these methods that are writing whole QC reports into the analysis.meta

illusional (approved):
LGTMichael

PR #721, opened 2024-05-03
Descriptive sv names

MattWellie:
I should definitely put a test case in for this


cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
```diff
@@ -531,18 +520,53 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+        return self.make_outputs(cohort, data=expected_output, jobs=[pyjob, bcftools_job])
 
 
 @stage(
-    required_stages=FilterGenotypes,
+    required_stages=SpiceUpSVIDs,
     analysis_type='sv',
     analysis_keys=['annotated_vcf'],
-    update_analysis_meta=_sv_annotated_meta,
+    update_analysis_meta=lambda x: {'remove_sgids': EXCLUSION_FILE},
```
cassimons, line 569: I never think of using a lambda! So much better.

cassimons (approved):
Looks great thanks :-)

cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py
MattWellie: I really hope this works 😅

PR #724, opened 2024-05-06
Adjusting Driver Cores and Memory Allocation for Improved Performance and Cost Reduction

michael-harper:
This PR proposes adjustments to the configuration of driver cores and memory allocation to prevent OOM issues.

[Slack thread](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1714941648403249)

Background:
The investigation initiated from a post on [Zulip](https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/QoB.20out-of-memory), where the distinction between worker jobs and driver jobs became apparent. Upon examining the failed job, it was confirmed to be a driver job. Furthermore, there is flexibility to assign driver memory and cores independently of worker memory.

Proposal:

- Currently, the worker memory setting is 'highmem,' which, unfortunately, did not address the failure. However, it did result in a notable cost reduction despite the same execution time.
- It is proposed to quadruple the number of driver cores and assign them as 'highmem', while retaining the high worker memory configuration.
- This adjustment aims to align the driver configuration more closely with the settings used in Dataproc, specifically targeting n1-highmem-8 instances.

The proposed changes were conducted in the following links. The driver jobs now have an allotted 27.9Gb of memory.
Batch link [here](https://batch.hail.populationgenomics.org.au/batches/445134), with worker jobs [here](https://batch.hail.populationgenomics.org.au/batches/445135).

Should the default `large_cohort.toml` be changed to have the following defaults in addition to `highmem_workers = true`?:
```toml
[workflow]
highmem_driver = true
driver_cores = 4
```


cpg_workflows/stages/large_cohort.py
```diff
@@ -31,7 +31,15 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
 
         j = get_batch().new_job('Combiner', (self.get_job_attrs() or {}) | {'tool': 'hail query'})
 
-        init_batch_args = {'worker_memory': 'highmem'} if get_config()['workflow'].get('highmem_workers') else None
+        init_batch_args = {}
+        config = get_config()['workflow']
```
vivbak, line 35: Could we use the new config_retrieve function instead of get_config

cpg_workflows/stages/large_cohort.py
```diff
@@ -31,7 +31,15 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+
+        if config.get('highmem_workers'):
+            init_batch_args['worker_memory'] = 'highmem'
+        if config.get('highmem_drivers'):
+            init_batch_args['driver_memory'] = 'highmem'
+        if 'driver_cores' in config:
+            init_batch_args['driver_cores'] = config['driver_cores']
```
vivbak, line 43: Could we just add a logging.warning message here if none of these values are specified, so the user knows to ensure they're included in the config if they're getting OOM errors.

configs/defaults/large_cohort.toml
```diff
@@ -6,6 +6,8 @@ status_reporter = 'metamist'
 # define reference fasta. If not specified here, the reference file is pulled from [references.broad][ref_fasta]
 ref_fasta='gs://cpg-common-main/references/hg38/v0/dragen_reference/Homo_sapiens_assembly38_masked.fasta'
 highmem_workers = true
+highmem_drivers = true
```
vivbak, line 9: I think this default behaviour is up to the PA team. If you are inherently going to need highmem for larger cohorts, then it makes sense to keep it as the default. I guess the consideration is whether this is overkill/drives up cost for smaller test runs.

vivbak (approved):

configs/defaults/large_cohort.toml
michael-harper: I'll remove this as the default behaviour given how infrequently we run complete cohorts. I've put it to the rest of PA to discuss what the default behaviour should be. Thanks @vivbak !

cpg_workflows/stages/large_cohort.py
```diff
@@ -31,7 +33,19 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
 
         j = get_batch().new_job('Combiner', (self.get_job_attrs() or {}) | {'tool': 'hail query'})
 
-        init_batch_args = {'worker_memory': 'highmem'} if get_config()['workflow'].get('highmem_workers') else None
+        init_batch_args = {}
```
MattWellie, line 36:
```suggestion
        init_batch_args: dict[str, str | int] = {}
```

suggesting this to make mypy happy

PR #727, opened 2024-05-06
Read_input requires a String

MattWellie:
How did pre-commit not find this? My IDE is now showing this as an error, but the CI which runs MyPy had no issues 🤔

Screenshot 2024-05-06 at 7 42 23 PM


vivbak (approved):

jmarshall:
Turns out pre-commit did not detect this because it didn't know what `get_batch()` returns. That can be usefully fixed by #728, but see also populationgenomics/cpg-utils#166's proposal.

PR #729, opened 2024-05-07
Precommit mypy increment

MattWellie:
Down to 119 mypy failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ To run the Seqr Loader workflow on the dataset called `validation`, create a con
 [workflow]
 input_datasets = ['validation']
 sequencing_type = 'genome'
-create_es_index_for_datasets = ['validation']
 skip_sgs = [
     'CPGXXXX',  # eg: low coverage (8x)
 ]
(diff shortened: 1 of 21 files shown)


jmarshall:
I can't review this without explanations of **why** you're making the changes you're making. In particular:

* Replacing instances of cpg_utils's `Path` by `(pathlib.Path, CloudPath)` seems like a backwards step. What is the reason for it?

* Adding `# type: ignore` annotations should be an absolute last resort IMHO. What is error being reported here, and why can we not reasonably refactor it away?

MattWellie:
1. See https://github.com/python/mypy/issues/12155, `error: Parameterized generics cannot be used with class or instance checks` - mypy (rightly or wrongly) doesn't support using our `Path` (a union of `CloudPath | pathlib.Path`) in `isinstance` checks
2. Most of the `type: ignore` statements are around Hail ResourceGroups - there's no way I can see to solve `error: Value of type "Resource" is not indexable` for named ResourceGroups, Hail does some job-creation-time magic to swap out `job.output['vcf.gz']` for a filepath, and the entity `job.[arbitrary_name]` can't really have a fixed type.

Other instances of the latter are around expected_outputs - when returning a `dict[str, str | Path]`, mypy fails with incompatible type on all usage of the contents unless the expected type of each is literally `str | Path`. We could pull out each value individually and assert a specific type, but... there's 120 more errors to address and only so many hours in the day.

jmarshall:
As you say, there are a lot of such instances, and it would not make development any more fun if we have to annotate our new code similarly.

I think we should look at selectively disabling some type checking (e.g. perhaps saying that whatever returns the expect_outputs returns them as Any; or nixing this plan to expose cpg_utils to mypy) rather than carry on down this road of putting ignore directives everywhere.

PR #733, opened 2024-05-07
BAM to CRAM conversion stage

EddieLF:
A stage to convert BAMs to CRAMs and write them to the bucket under `pacbio/cram`. Utilises the existing bam_to_cram job.

Also adds a new default config for the long read workflow - `configs/defaults/seqr_loader_long_read.toml`. 
This config includes a specific analysis type for outputs from the BamToCram stage:
```
[workflow]
bam_to_cram_analysis_type = 'pacbio_cram'
```

This analysis type will be picked up by the stage and created analyses will be stored under this type. Without this config option, the analysis type will be `cram`.  
The output path with be `gs://cpg-dataset-main/pacbio/cram/CPGxxxxx.cram`, as well as an adjacent cram index.

The jobs can be seen running to completion and creating outputs for some test data here:
[Driver job](https://batch.hail.populationgenomics.org.au/batches/453331/jobs/1)
[BamToCram jobs](https://batch.hail.populationgenomics.org.au/batches/453334)


cassimons (approved):
Looks good.

Going forward, I am keen we start namespacing pipeline components. Can you please move cpg_workflows/stages/bam_to_cram.py to cpg_workflows/stages/seqr_loader_long_read/bam_to_cram.py

Also, all config options for all pipelines need to be added to  cpg_workflows/defaults.toml (even if commented).

cpg_workflows/stages/bam_to_cram.py
```diff
@@ -0,0 +1,72 @@
...
+        input_bam = b.read_input_group(bam=str(sequencing_group.alignment_input_by_seq_type.get('genome')))
+        job, output_cram = bam_to_cram.bam_to_cram(
+            b=get_batch(),
+            input_bam=input_bam,
+            extra_label='long_read',
+            job_attrs=self.get_job_attrs(sequencing_group),
+            requested_nthreads=1,
+        )
+        b.write_output(output_cram['cram'], str(self.expected_outputs(sequencing_group)['cram']))
+        b.write_output(output_cram['cram.crai'], str(self.expected_outputs(sequencing_group)['cram.crai']))
```
MattWellie, line 66: This section is already returning a resource group of `cram` & `cram.crai`, so you can simplify this to:

```suggestion
        b.write_output(output_cram, str(self.expected_outputs(sequencing_group)['cram']).removesuffix('.cram'))
```

MattWellie (approved):
Comment to simplify, but looks good

EddieLF:
Thanks both! 

@cassimons I have actioned your suggestions and moved the stage into the `cpg_workflows/stages/seqr_loader_long_read/` directory, as well as added the `workflow. bam_to_cram_analysis_type` config field to `defaults.toml`, commented out but with a comment explaining its purpose.

PR #736, opened 2024-05-08
Another failed config_override usage

MattWellie:
https://batch.hail.populationgenomics.org.au/batches/445300/jobs/1

For the AIP bit:
https://batch.hail.populationgenomics.org.au/batches/445299

Issue: The SV labelling stage only runs when a SV VCF is available. A later stage tries to find the output of the SV labelling stage, and fails if the output dictionary is empty. That should probably not cause a failure... but at the moment it does


cassimons (approved):

EddieLF (approved):

MattWellie:
Ingested the changes from #731 - that was a similar config_override failure on a stage that wasn't intended to run

PR #741, opened 2024-05-12
Enable CPG ID pre-commit hook 

vivbak:
The majority of the changes are trivial, existing only in test files and documentation. 

There is one functional change required however that I have not yet addressed. 

[This line ](https://github.com/populationgenomics/production-pipelines/blame/c4012cd419972f89f6e50567f7a593beb9b40352/scripts/find_samples_needing_sv.py#L61) is a functional screen so I can't just switch it. 

@MattWellie, how often do you use this script? Do you have thoughts on either 
A) Pulling out the borked SG IDs into an argument for this script 
B) Creating a cohort that excludes these two SG IDs.


vivbak:
Pending this https://github.com/populationgenomics/production-pipelines/pull/742

illusional (approved):
🎉

jmarshall (approved):
I kind of thought we were going to come up with a convention like `CPGLCL1` etc (LCL = local) or otherwise a separate prefix to make test data with.

And/or I wonder if we could carve off an exception for `/^CPG[1-9]$/` — did we use such small numbers in production?

Anyway, musing over, LGTM.

vivbak:
TIL LCL is an abbreviation for local 😂

In my head that was the conversation in the metamist repo, not prod pipes. The tests there functionally rely on the CPG Id being a valid output -- for example validating the lunh checksum so CPGAAA doesn't work. But I don't believe it's an issue here because we assume CPG IDs are valid in prod pipes. 

Though I don't have any issue keeping in consistent. Merging for now but happy to open up another PR later if we want to keep the same convention here.

jmarshall:
That's what I assume it stands for 🤷 

Yep, fair enough; and no biggie. There's some in _test/test_cohort.py_ which may or may not have their luhns enforced.

PR #742, opened 2024-05-12
Remove find samples needing sv script for linting 

vivbak:
This script fails linting because of CPG1 and CPG2. The script is no longer used so we are just deleting it rather than fixing


MattWellie (approved):
Agreed, this functionality is migrated elsewhere

Issue #737, opened 2024-05-09
Error when not specifying large_cohort.output_versions


michael-harper:
LC pipeline needs to have large_cohort.output_versions for certain stages otherwise dictionary .get() method won't work. See [failed batch](https://batch.hail.populationgenomics.org.au/batches/445312/jobs/1).

If the field is not populated then it should default to the `workflow[output_version]` name

PR #744, opened 2024-05-14
Removing the necessity of having large_cohorts.output_versions in config files

michael-harper:
Fixes issue https://github.com/populationgenomics/production-pipelines/issues/737

Have added the new `config_retrieve()` function with a `default=None` instead of nested `.get()` methods.

Batch that succeeded: https://batch.hail.populationgenomics.org.au/batches/447653/jobs/1 (I cancelled the subsequent batch but the driver kicked). Compared to failed batch linked in the issue https://batch.hail.populationgenomics.org.au/batches/445312/jobs/1 that has the same config


vivbak (approved):
LGTV, everyone else has a fun play on LGTM and i think its unfair. 

(Assuming tests pass)

PR #745, opened 2024-05-15
Use AIP report splitting, config dependent

MattWellie:
Utilises the functionality in https://github.com/populationgenomics/automated-interpretation-pipeline/pull/403

An example config structure would be:

```toml
[workflow.report_splitting]
PROJECT = 2
```

Meaning that reports generated in project `PROJECT` would be generated as normal, then 2 smaller sub-reports would be generated.

This granularity should enable each project to be either published un-split, or to be split into an arbitrary number of partitions, depending on scale of cohort.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.24.2
+current_version = 1.24.3
 commit = True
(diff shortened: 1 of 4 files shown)


EddieLF (approved):
LGTM

PR #746, opened 2024-05-15
Cast somalier_sites reference to path and take the basename

EddieLF:
The somalier extract jobs are failing ([e.g.](https://batch.hail.populationgenomics.org.au/batches/447718?q=state%3Dfailed)) because of changes to cpg-utils `reference_path` function. The somalier extract job was changed in [this PR](https://github.com/populationgenomics/production-pipelines/commit/c2a30ccf5afd124bf90ec836cfffab5a180ba84e#diff-c52574c0e01c6af6a6a87d93c65e9b7a21f773186cd8cfc806483cc14f52a1f6L267
). 

The `reference_path` function used to be in `cpg_utils.hail_batch`, it would return a `Path` object like so:
```python
def reference_path(key: str) -> Path:
    return to_path(retrieve(['references'] + key.strip('/').split('/')))
```

Following the update, `reference_path` (now in `cpg_utils.config`) returns a string
```python
def reference_path(key: str) -> str:
    return config_retrieve(['references', *key.strip('/').split('/')])
```

---

This change restores the `Pathlike.name` aspect of the `somalier_sites` reference path in the context of the extract job. Without this, the batch job tries to append a full `gs://` path to the batch mounted storage, leading to job failure as it tries to read `/io/batch/hash/sites/gs://cpg-common-main/references/somalier/sites.hg38.vcf.gz`.

### DIFF
--- a/cpg_workflows/jobs/somalier.py
+++ b/cpg_workflows/jobs/somalier.py
@@ -2,6 +2,7 @@
 Adding jobs for fingerprinting and pedigree checks. Mostly using Somalier.
 """
 
+import os
 from typing import cast
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


illusional (approved):
I'd probably opt for `os.path.basename` on the result, as IMO it's more common, and very clear what the result is. But change is fine as it is :)

PR #748, opened 2024-05-17
Force sex-unknown pedigree in gCNV

MattWellie:
We've been having a number of issues in gCNV at the Stage `GCNVJointSegmentation`

Up until this stage Ploidy is derived from the model created in `DeterminePloidy`, which runs a ploidy inference based on sequence data. This looks to be accurate.

In this stage sex/ploidy is read from a Pedigree, created from the Metamist entities. This contains the occasional sample swap/mis-assigned sex, such that the Pedigree sex isn't always accurate.

We would like to allow an override when generating the PED file, so that the sex column can be overridden to '0' for all samples. There isn't an expected requirement for using this outside of gCNV, and our preference is for updating Metamist sex with inferred sex where possible, so this is more of a patch than a long term solution.

### DIFF
--- a/cpg_workflows/stages/gcnv.py
+++ b/cpg_workflows/stages/gcnv.py
@@ -232,7 +232,7 @@ class GCNVJointSegmentation(CohortStage):
 
         expected_out = self.expected_outputs(cohort)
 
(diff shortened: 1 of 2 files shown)


MattWellie:
Gotta be something sexier we can do here

PR #751, opened 2024-05-17
Upgrade ped with inferred

MattWellie:
Takes the default pedigree we write from Metamist info
Takes the Tar-ball of DetermineContigPloidy outputs
parses the tar contents to find the inferred X/Y ploidy
Uses the ploidy counts to identify inferred 0/Unknown, 1/Male, 2/Female
Updates attributes in the pedigree and writes a new file


cassimons (approved):
🙌

PR #754, opened 2024-05-17
PED Update tested

MattWellie:
BUG BUG BUG

- didn't call the method, for a start
- Assigned X and Y to 'X', so Y is always None
- No newlines
- 'required' is not a valid argparse argument for positional arguments

During testing of this I found that the re-run VCGS Twist Exomes need to be re-run all the way from the start (currently https://batch.hail.populationgenomics.org.au/batches/448333 does not have allosomal calls)

Tested properly

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.24.5
+current_version = 1.24.6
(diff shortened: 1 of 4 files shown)


cassimons (approved):
Well, if you are going to go and do it all properly 🙄

Issue #710, opened 2024-04-29
Handle Multicohorts 


vivbak:
https://github.com/populationgenomics/production-pipelines/pull/635 Only handles a single cohort, need to add support for multi-cohorts.

PR #764, opened 2024-05-22
:sparkles:MultiCohorts:sparkles:

vivbak:
Closes https://github.com/populationgenomics/production-pipelines/issues/710

Currently, for each pipeline run a ‘cohort’ is defined, comprising a list of one or more ‘datasets’, comprising a list of one or more ‘sgs’

In this PR, we propose that for each pipeline a ‘multicohort’ will be defined, comprising a list of one or more ‘cohorts’, comprising a list of one or more ‘datasets’, comprising a list of one or more ‘sgs’. 🤯 

This means that a user can specify a list of custom cohort ID's (rather than just one), and stages will be generated accordingly. 

The key changes here
- The addition of a new MultiCohortStage, which will act on all of the cohorts at once. 
- The addition of a new MultiCohort Input, to facilitate the above. 
- Switching out the existing cohort global variable, with a multicohot 

It is important to note, that we are still supporting a non-cohort run of production pipelines, which means there needs to be some logic to support the old way of doing things (pre-multi-cohorts) for the time being.

### TODO

**metamist.py**
- [x] Create get_multi_cohort method 

**targets.py**
- [x] Create a new MultiCohort Target. 
   - [x] Add get_cohorts method 
   - [x] Add create_cohort method 
 
**inputs.py**
- [x] Modify get_cohort -> get_multicohort()
- [x] Modify create_cohort() -> create_multi_cohort() 
- [x] Consider how to handle current instances of get_cohort(), for example `intervals_path=inputs.as_path(get_cohort(), PrepareIntervals, 'preprocessed'),` in a SequencingGroupStage where the dependent stage is a CohortStage. *
- [x] Add validation for input_cohorts to ensure it is a list, otherwise a string will be list-ified. 

**workflow.py**

- [x] Create MultiCohortStage 
- [x] Update all instances of get_cohort, as well as relevant implementation changes. 
- [x] Update set_stages, L1148
- [x] Update queue jobs with checks
- [x] Update _prefix 
- [x] Update queue for cohort 
- [x] Update batch name, L883

**test_*.py**
- [x] Update Existing Tests 

**test_cohort.py**
- [x] Add test_multi_cohort() 

**sample_qc.py**
- [x] Switch get_cohort to get_inputs. 

**combbiner.py**
- [x] Switch get_cohort to get_inputs 

_Note for VB, add co-author credit in merge._

*_actually I think this is fine, because it will return the outputs for each cohort, but we need to make sure that the new return structures here are suitable._


test/stages/test_stage_types.py
```diff
@@ -0,0 +1,280 @@
...
+
+    from cpg_utils.hail_batch import get_batch
+    from cpg_workflows.inputs import get_inputs
+
+    run_workflow(mocker, cohort_mocker=mock_multidataset_cohort, stages=[DatasetStage1])
+    print('Job by stage:', get_batch().job_by_stage)
+    assert get_batch().job_by_stage.get('DatasetStage1')
+    # The next two tests are looking at the same behaviour, but from different angles
+    assert get_batch().job_by_stage.get('DatasetStage1').get('job_n') == 2
+    assert get_batch().job_by_stage.get('DatasetStage1').get('job_n') == len(get_inputs().get_datasets())
```
vivbak, line 93: First assert that the len(get_inputs().get_datasets())) == 2, then leave 93 as is.

vivbak (commented):
Co-Reviewed with MH

cpg_workflows/inputs.py
```diff
@@ -42,6 +42,8 @@ def get_inputs() -> Cohort | MultiCohort:
 
     # NOTE: When configuring sgs in the config is deprecated, this will be removed.
     if config_retrieve(['workflow', 'input_cohorts'], []):
```
MattWellie, line 44: `if config_retrieve(['workflow', 'input_cohorts'], []):` can be replaced with `custom_cohort_ids`, this was already set above

MattWellie: Also not really relevant to anything, the logical test above can be changed to
```
if (input_datasets := config_retrieve(['workflow', 'input_datasets'], [])) and (custom_cohort_ids := config_retrieve(['workflow', 'input_cohorts'], [])):
    ...
```

and I think that's cool

MattWellie (commented):
There's 1000 lines in this PR, but I found something outside that scope to comment on. You're welcome 👼

cpg_workflows/targets.py
```diff
@@ -97,18 +97,118 @@ def rich_id_map(self) -> dict[str, str]:
...
+        return {
+            'sequencing_groups': self.get_sequencing_group_ids(),
+            'datasets': [d.name for d in self.get_datasets()],
+            'cohorts': [c.name for c in self.get_cohorts()],
+        }
+
+
 class Cohort(Target):
     """
     Represents a "cohort" target - all sequencing groups from all datasets in the workflow.
```
MattWellie, line 201:
```suggestion
    Represents a "cohort" target - all sequencing groups from a single CustomCohort (potentially spanning multiple datasets) in the workflow.
```
Not fully confident on this, but seems like this is now out of sync. Should this be 'all SG IDs in a single custom cohort, unless only datasets are used as inputs, in which case... it represents all SG IDs, but this use case is being phased out?'

cpg_workflows/targets.py
```diff
@@ -97,18 +97,118 @@ def rich_id_map(self) -> dict[str, str]:
...
     Analysis dataset name is required and will be used as the default name for the
     cohort.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, name: str | None = None, multicohort: MultiCohort | None = None) -> None:
```
MattWellie, line 206: Why does a cohort take a multicohort as an input? Is this to do with getting cohort Stage input from a multicohort stage? I'm assuming (still reading) that we go to metamist, get all cohorts in the analysis (MultiCohort), for each of those create a Cohort, then clustering by Dataset we create the Dataset input types...

MattWellie (commented):
🤯

cpg_workflows/inputs.py
```diff
@@ -8,41 +8,112 @@
...
     """Return the cohort object"""
     global _cohort
     if not _cohort:
-        _cohort = create_cohort()
+        _cohort = deprecated_create_cohort()
     return _cohort
 
 
-def create_cohort() -> Cohort:
+def get_inputs() -> Cohort | MultiCohort:
```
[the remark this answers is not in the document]
vivbak: naming, not my forte.

Does it help that the function won't exist post deprecation, and will be replaced simply by get_multicohort()?

cpg_workflows/inputs.py
vivbak: define cool 😂

cpg_workflows/inputs.py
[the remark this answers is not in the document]
vivbak: truly don't mind renaming, just lacking the divine inspiration required to come up with a better function name 😂

cpg_workflows/targets.py
vivbak: Your correction is right! As is your understanding.

cpg_workflows/targets.py
```diff
@@ -97,18 +97,118 @@ def rich_id_map(self) -> dict[str, str]:
...
         super().__init__()
-        self.name = get_config()['workflow']['dataset']
+        self.name = name or get_config()['workflow']['dataset']
         self.analysis_dataset = Dataset(name=self.name, cohort=self)
         self._datasets_by_name: dict[str, Dataset] = {}
+        self.mutlicohort = multicohort
```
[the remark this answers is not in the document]
vivbak: beautiful catch!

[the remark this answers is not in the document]
vivbak: although I feel like _something_ should have complained about that

cpg_workflows/targets.py
```diff
@@ -97,18 +97,118 @@ def rich_id_map(self) -> dict[str, str]:
...
+    Represents a "multi-cohort" target - multiple cohorts in the workflow.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # NOTE: For a cohort, we simply pull the dataset name from the config.
+        input_cohorts = get_config()['workflow'].get('input_cohorts', [])
+        if input_cohorts:
+            cohorts_string = '_'.join(input_cohorts)
```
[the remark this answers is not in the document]
vivbak: yeah nice!

cpg_workflows/targets.py
vivbak: Every target takes in its 'parent' as input in the constructor so that you can link them. 
You can see this behaviour in SG's taking in a Dataset and Datasets taking in a Cohort. 

> Is this to do with getting cohort Stage input from a multicohort stage? 
More so used the otherway around at the moment. 

Sorry I'm not sure I understand the second part of your message.

cpg_workflows/inputs.py
MattWellie: I love a cheeky walrus 🦭

cpg_workflows/targets.py
MattWellie: Makes sense to me! The other part is probably just incoherent rambling

cpg_workflows/inputs.py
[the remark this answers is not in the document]
vivbak: As per a sugggestion from @jmarshall (thanks!), this has been renammed to `get_multicohort()`, and the current `get_multicohort()` function has been renamed to `actual_get_multicohort()`. This way, when we deprecate, we simply need to delete this `get_inputs()` function (now called `get_multicohort()`) and rename `actual_get_multicohort()`.

vivbak:
We need to review all the existing Cohort stages to see if they are eligible (or more suited) to be a MultiCohort stage.

cpg_workflows/inputs.py
```diff
@@ -8,41 +8,111 @@
...
+    read_pedigree = config.get('read_pedigree', True)
+    for cohort_id in custom_cohort_ids:
+        cohort = multicohort.create_cohort(cohort_id)
+        sgs_by_dataset_for_cohort = datasets_by_cohort[cohort_id]
+        populate_cohort(cohort, sgs_by_dataset_for_cohort, read_pedigree=read_pedigree)
+
+    return multicohort
+
+
+def populate_cohort(cohort: Cohort, sgs_by_dataset_for_cohort, read_pedigree: bool = True):
```
jmarshall, line 70: If this is internal-only like the other `_populate_foo()` functions, maybe mark it as such: `_populate_cohort`?

cpg_workflows/inputs.py
```diff
@@ -8,41 +8,111 @@
...
+                meta=metadata,
+            )
+
+            if reported_sex := entry['sample']['participant'].get('reportedSex'):
+                sequencing_group.pedigree.sex = Sex.parse(reported_sex)
+
+            _populate_alignment_inputs(sequencing_group, entry)
+
+    if not cohort.get_datasets():
+        raise MetamistError('No datasets populated')
```
jmarshall, line 97: (Dumb question, probably…) Are `skip_sgs`/`only_sgs` being removed once the Old Way is removed?

cpg_workflows/metamist.py
```diff
@@ -259,6 +259,13 @@ def __init__(self) -> None:
         self.default_dataset: str = get_config()['workflow']['dataset']
         self.aapi = AnalysisApi()
 
+    def get_sgs_for_cohorts(self, cohort_ids: list[str]) -> dict[str, dict[str, Any]]:
+        """
+        Retrieve the sequencing groups per dataset for a list of cohort IDs.
+        """
+
```
jmarshall, line 266: Possibly no blank line here, but who really cares other than linters? 🙄

cpg_workflows/stages/gcnv.py
```diff
@@ -82,8 +81,10 @@ def queue_jobs(self, seqgroup: SequencingGroup, inputs: StageInput) -> StageOutp
         if seqgroup.cram is None:
             raise ValueError(f'No CRAM file found for {seqgroup}')
 
+        assert seqgroup.dataset.cohort
+
         jobs = gcnv.collect_read_counts(
-            intervals_path=inputs.as_path(get_cohort(), PrepareIntervals, 'preprocessed'),
+            intervals_path=inputs.as_path(seqgroup.dataset.cohort, PrepareIntervals, 'preprocessed'),
```
jmarshall, line 87: It would be good (for posterity) to briefly explain in a commit message why `get_cohort()` was the wrong choice for _gcnv.py_.

cpg_workflows/targets.py
```diff
@@ -97,18 +97,118 @@ def rich_id_map(self) -> dict[str, str]:
...
+    def target_id(self) -> str:
+        """Unique target ID"""
+        return self.name
+
+    def get_cohorts(self, only_active: bool = True) -> list['Cohort']:
+        """
+        Gets list of all cohorts.
+        Include only "active" cohorts (unless only_active is False)
+        """
+        cohorts = [cohort for _name, cohort in self._cohorts_by_name.items()]
```
jmarshall, line 133: Could use `.values()` here if you don't need `_name`, but no biggie.

(Or even just `list(self._cohorts_by_name.values())` or leave it as the view and make it into a list with `return list(cohorts)` only at the end.)

cpg_workflows/workflow.py
```diff
@@ -1239,6 +1254,42 @@ def queue_for_cohort(self, cohort: Cohort) -> dict[str, StageOutput | None]:
 
         return output_by_target
 
+    def queue_for_multicohort(self, multicohort: MultiCohort) -> dict[str, StageOutput | None]:
+        """
+        Plug the stage into the workflow.
+        """
+        output_by_target: dict[str, StageOutput | None] = dict()
+        for cohort in multicohort.get_cohorts():
+            for dataset in cohort.get_datasets():
```
jmarshall, line 1263: Do we need to emit a message for no datasets, like the previous function does?

jmarshall (approved):
Some minor comments, and I ran the tests but didn't read the test code extremely carefully. But LGTM either with or without any particular responses to the comments.

cpg_workflows/inputs.py
vivbak: Done :)

cpg_workflows/inputs.py
vivbak: Yes indeed! Once we deprecate the 'Old Way' a user won't need to/be able to specify sgs to skip in prod pipes :)

cpg_workflows/metamist.py
vivbak: Haha removed :)

cpg_workflows/targets.py
vivbak: Yeah totally fair enough, updated to use .values()! Briefly experimented with a one line comprehension for this entire method, but the readability was gross hahaha

cpg_workflows/workflow.py
vivbak: Added the warning. What is more important (that this comment highlighted to me) is we don't do a warning for no cohorts in the multi-cohort. Added that too!

vivbak: Although these warnings in general are less important now adays, historically and as per 'The Old Way' a user could specify 'skip_datasets' so you could find yourself in a position where all your specified datasets have been skipped -- hence the message. But we're scrapping this functionality, so arguably, we shouldn't encounter them.

jmarshall:
> We need to review all the existing Cohort stages to see if they are eligible (or more suited) to be a MultiCohort stage.

I meant to also say: I guess we can do this at our leisure and on a case-by-case basis. It probably doesn't need to be included in this base infrastructure update PR.

vivbak:
> I meant to also say: I guess we can do this at our leisure and on a case-by-case basis. It probably doesn't need to be included in this base infrastructure update PR.

@jmarshall, the only issue would be if someone supplied multiple cohorts. If most users are using a single cohort at a time, then you're right we can do it case by case.

michael-harper (approved):
Looks like you've covered everything! From a PA perspective we currently don't have any Cohorts defined but the added MultiCohort functionality isn't effecting individual Cohort functionality (which as far as I know has been relatively issue-free!) so we're happy to move forward 👏

MattWellie (approved):
Nice! Happy to approve on the basis that the Cohort->MultiCohort Stage translation can continue on a per-workflow basis instead of lumping everything here as a requirement to merge this

PR #765, opened 2024-05-22
Castrate aneuploidies

MattWellie:
This PR needs a new name. 

The issue: We have data which identifies sex-aneuploidies, gCNV/GATK can only tolerate XX/XY data, and we have no way to remove the data which will cause failures.

The solution? A solution?

- When we update the pedigree file with inferred sex, keep a list of all samples which are not XX/XY
- For each of those SGIDs identified, generate a version of the VCF (& index) with chrX & chrY removed entirely
- When running joint-segmentation, if the VCF was stripped of sex chr data, use the trimmed version, otherwise use the default version

Note: because the workflow needs to be defined in advance in Hail Batch, this design means that a run will need to fail once:

- run for the first time, including generating the updated pedigree and finding any sex-aneuploid samples
- if the samples can't all be processed, the run will die
- upon re-running, the sex-aneuploidy file will be read, new trimmed VCFs will be generated for the relevant samples, and the GCNVJointSegmentation stage will run with the trimmed VCFs as appropriate

### DIFF
--- a/cpg_workflows/jobs/gcnv.py
+++ b/cpg_workflows/jobs/gcnv.py
@@ -43,7 +43,8 @@ def upgrade_ped_file(local_ped: ResourceFile, new_output: str, aneuploidies: str
 
 def prepare_intervals(job_attrs: dict[str, str], output_paths: dict[str, Path]) -> Job:
     j = get_batch().new_job(
-        'Prepare intervals', job_attrs | {'tool': 'gatk PreprocessIntervals/AnnotateIntervals'},
+        'Prepare intervals',
+        job_attrs | {'tool': 'gatk PreprocessIntervals/AnnotateIntervals'},
     )
     j.image(image_path('gatk_gcnv'))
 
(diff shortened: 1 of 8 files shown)


cpg_workflows/jobs/gcnv.py
```diff
@@ -468,7 +456,7 @@ def run_joint_segmentation(
     tmp_prefix: str,
     output_path: Path,
     job_attrs: dict[str, str] | None = None,
-) -> list[Job]:
+) -> list[BashJob]:
```
cassimons, line 459: OOI, What does BashJob give us here over Job?

cassimons (approved):

cpg_workflows/jobs/gcnv.py
MattWellie: Ah this was just some rubbish with the type inspections - we were setting an expected `Job` return, when we were generating a BashJob 🤷 

This doesn't fail in CI, but it fails for me locally - related to the whole 'we aren't doing proper type inspections' issue addressed [here](https://github.com/populationgenomics/production-pipelines/pull/728)

PR #767, opened 2024-05-24
Pin Hail to the version deployed in CPG's installation

jmarshall:
Followup to PRs #749 and #750: Pin this major dependency to avoid ever picking up an incompatible version. Change grpcio-status pinning to versions that are compatible with hail 0.2.130, but keep bounds on it to help pip runtime. We should no longer need anything special for peddy.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.24.9
+current_version = 1.24.10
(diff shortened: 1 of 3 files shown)


milo-hyben (approved):

milo-hyben:
Hail Batch update to 0.2.130 was successful, merging this one so production pipelines are updated as well.

PR #771, opened 2024-05-28
Refactoring main for local dev

milo-hyben:


MattWellie:
I feel weird about modifying the entrypoint of a pipeline which is designed to run in a container, in order that we can run it outside a container.

This does feel like a clean solution, but equally wouldn't building a docker container and using that for local testing be more appropriate, so that we know that "it works on my machine" == "it works in production". Most modern IDEs can run tests while using a python environment inside a container AFAIK (though I can't remember if that's a paid/premium feature, it's been a while since I used it).

I'm also not sure exactly what changes are being tested in the situation this PR addresses, so perhaps that wouldn't be a good solution.

jmarshall:
I don't especially like the proposed solution — as it seems to lead to duplication, for one thing — and I don't really see what it buys you: if you were working on one particular workflow, you're still going to have to edit _workflows_dev.py_ to contain (only?) that workflow. That's basically equivalent to commenting things out in _main.py_, so it doesn't seem like a big win.

But more importantly I still don't understand what the problem this is trying to solve is.

The example given in the meeting was ”it'd be nice not to have to have hail installed locally to do prod-pipes development”. But prod-pipes is very very bound into hail, so I'm not sure what workflows you could be working on locally that wouldn't need to be able to `import hailtop.batch`. (Moreover _main.py_ imports `cpg_workflows` which imports `cpg_utils.hail_batch` which imports hail — so it seems to me that you can't do anything at all in prod-pipes without the hail modules installed…)

PR #777, opened 2024-05-30
Trim analysis callables

MattWellie:
I don't think this is relevant to the current failing workflows, but we have a large number of pointless metadata-update callables. This process of updating an analysis dict with a called method in a later job is... probably not a good mechanic.

I would personally use a combination of analysis entry type and meta.stage to find entries, and AFAIK none of the fields removed here are used in scripts/other stages. The gvcf_qc meta update should have already been removed in https://github.com/populationgenomics/production-pipelines/pull/718

In code these callable methods are annotated with `TODO: Replace this once dynamic analysis types land in metamist.` - we have that now, so we should probably update the analysis types we generate

### DIFF
--- a/cpg_workflows/jobs/stripy.py
+++ b/cpg_workflows/jobs/stripy.py
@@ -29,12 +29,7 @@ def stripy(
     """
     Run STRipy
     """
-    if can_reuse(
-        [
-            out_path,
-        ],
-        overwrite,
-    ):
+    if can_reuse(out_path, overwrite):
         return None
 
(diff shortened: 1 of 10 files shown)


illusional:
Related: https://github.com/populationgenomics/production-pipelines/pull/772

illusional (approved):

MattWellie:
Potentially back on the table, now that @EddieLF has identified the reason this data was never making it into metamist. Still reasonable to delete all this useless code, but not a priority

PR #780, opened 2024-06-04
Add --exclude-intervals h38 telomeres centromeres to genotypegvcfs jobs

EddieLF:
Localises the `hg38telomeresandcentromeres.interval_list` file for every `GenotypeGVCFs` job and passes it to the option `--exclude-intervals`. 

In theory, this should ensure that even though we `--merge-input-intervals` for the `GenomicsDB` jobs, we **do not** traverse the telomeres and centromeres during the `GenotypeGVCFs` jobs. This should give us the best of both worlds. Each GenomicsDB fragment will be created quickly as it merges intervals, and each GenotypeGVCFs fragment will run quickly as it skips over the problematic regions of the genome.

---

**It's unclear to me if this is actually necessary. I did two trial batches:**

Using this branch: https://batch.hail.populationgenomics.org.au/batches/455152

Using the main branch: https://batch.hail.populationgenomics.org.au/batches/455159

I went through each `GenotypeGVCFs` job in both batches, and neither of them ever traverse the telomeres/centromeres. Perhaps this is because the number of genomes is too low, and there is no coverage in those regions when n=3. However it might also indicate that we're fine with things the way they are, and telomeres / centromeres will not be traversed with the current code.

### DIFF
--- a/cpg_workflows/jobs/joint_genotyping.py
+++ b/cpg_workflows/jobs/joint_genotyping.py
@@ -6,14 +6,15 @@ import logging
 import math
 from enum import Enum
 
-import hailtop.batch as hb
 import pandas as pd
-from cpg_utils import Path, to_path
-from cpg_utils.config import get_config, image_path, reference_path
-from cpg_utils.hail_batch import command, fasta_res_group
+
+import hailtop.batch as hb
 from hailtop.batch import Resource
 from hailtop.batch.job import Job
 
+from cpg_utils import Path, to_path
+from cpg_utils.config import get_config, image_path, reference_path
+from cpg_utils.hail_batch import command, fasta_res_group
 from cpg_workflows.filetypes import GvcfPath
 from cpg_workflows.resources import STANDARD, joint_calling_scatter_count
 from cpg_workflows.utils import can_reuse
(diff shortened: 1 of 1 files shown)


EddieLF:
@cassimons @MattWellie tagging you in for visibility - I don't think we need to merge this yet. I think we should give it a go with the current code and monitor if the GenotypeGVCFs jobs are getting stuck on the centromeres, at which point we could give this code change a try.

cassimons:
In your trial jobs were the gvcfs called with the black list? If so then we expect thais to run fine either way. Our problem is that a heap of our existing gvcfs were called without the black list so we will hit the same problem we always used to in centromeres.

EddieLF:
@cassimons good point! Those 3 genomes in the batches from my previous comment had been called with the blacklisted regions.

I reran the test with 5 other genomes, whose gvcf files were created in April 2023, before we were blacklisting any regions.

Using this branch: https://batch.hail.populationgenomics.org.au/batches/455175

Using the main branch: https://batch.hail.populationgenomics.org.au/batches/455178

I didn't thoroughly check every chromosome, but the centromeres don't appear to be traversed.

cassimons (approved):

Issue #647, opened 2024-03-14
Remove/modify Query_command?


MattWellie:
We use query_command a lot in prod-pipes. This method design is unpleasant:

- we pass a module, a function name, and tuple of arguments to query_command
- query_command builds a bash script around this module to start a hail runtime, monitor storage space, etc
- it pulls the module source contents, and runs a `cat` to write those contents to a file
- the final line of the file is a run instruction to execute the specified methods, done by writing `<method name><argument tuple>`

This is a really backwards process that enables some barely-predictable runtime behaviours (e.g. pip installing packages at runtime).

I will advocate for replacing all this with the following:

- take each process which runs using query_command and split out into a script with a specific entrypoint. This script should be fixed in version control.
- each use of query_command should be replaced with running a python script with CLI arguments.
- a job wrapper can still take care of authentication and periodically monitoring space, rather than transcribing a python script through Bash

Each job that would currently run using query_command would now run a specific script path inside a specific image, so we'd more easily troubleshoot failures

jmarshall:
I largely agree with that, except that replacing code that is currently contained in your production-pipelines branch/commit with code from a specific image — so that to iterate and develop it, you would have to build a new image — would be a massive step backwards.

What we need is a simple way to transport scripts from your production-pipelines branch/commit to the worker node, which is something that @cassimons and I have discussed in the past. That was stalled but has started to progress again.

MattWellie:
> except that replacing code that is currently contained in your production-pipelines branch/commit with code from a specific image — so that to iterate and develop it, you would have to build a new image — would be a massive step backwards.

I don't agree - the current problems are evidence that taking the code from one image and the libs in another is already causing problems, and we have the grotty transcription from python-to bash-to python in addition to that. Testing new query_command executed code at the moment requires overriding the target image in config, so it's at worst a step sideways, but not backwards.

The pipeline was designed to be iterated on by iterating on the images - the old CI workflow built a new image with each commit with the intention that each code change = an image change, which is then used to test the change. That's a decision that still haunts us, and we could do with a strategy to fix that. 

Copying a script/file from the current container into GCP batch temp, then copying the temp file into batches to execute is a workaround for creating the script file, but doesn't address dependencies. If we want a test run to prove that an updated library produces equivalent or better results, it's too easy to think we're testing the latest libraries, but accidentally use the canonical cpg_workflows image to execute, which was the problem we had yesterday.

cassimons:
[...]

PR #783, opened 2024-06-05
Extract 'Alignment' and 'Genotype' to be separate independent pipelines

michael-harper:
See the context [here](https://docs.google.com/document/d/1GF2Yxc6p9S5YZtjkd8wVzIoI3aa5ii9K_FCQK-KrO44/edit#heading=h.ktjt9ukaj2pc) in the background of the scoping document


cpg_workflows/large_cohort/combiner.py
```diff
@@ -26,7 +26,8 @@ def _check_gvcfs(sequencing_groups: list[SequencingGroup]) -> list[SequencingGro
                     f'Sequencing group {sequencing_group} is missing GVCF. '
                     f'Use workflow/skip_sgs = [] or '
```
MattWellie, line 27: I think this argument will be disappearing soon following the implementation of custom cohorts, @vivbak ?

cpg_workflows/stages/alignment/align.py
```diff
@@ -18,7 +21,10 @@
 )
 
 
-@stage(analysis_type='cram', analysis_keys=['cram'])
+@stage(
+    analysis_type='cram',
+    analysis_keys=['cram'],
+)
```
MattWellie, line 27:
```suggestion
@stage(analysis_type='cram', analysis_keys=['cram'])
```

cpg_workflows/stages/alignment/align.py
```diff
@@ -3,9 +3,12 @@
 """
 
 import logging
+from enum import Enum
+from logging import config
+from pickle import MARK
```
MattWellie, line 8: lots of unused imports here, not sure how this passed linting checks

cpg_workflows/stages/genotyping/genotype.py
```diff
@@ -10,15 +10,15 @@
...
     StageInput,
     StageOutput,
+    Workflow,
+    WorkflowError,
     stage,
 )
 
-from .. import get_batch
-from .align import Align
+from ... import get_batch
```
MattWellie, line 18: not a fan of this at all. We're installing cpg_workflows as a package in the image we intend to run. Relative imports are garbage.

This is also the wrong import, we're using `cpg_utils.hail_batch.get_batch`
```suggestion
from cpg_utils.hail_batch import get_batch
```

cpg_workflows/stages/genotyping/gvcf_qc.py
```diff
@@ -24,6 +23,8 @@
     stage,
 )
 
+from .genotype import Genotype
```
MattWellie, line 26: BLEH RELATIVE IMPORTS

MattWellie (changes requested):

MattWellie:
I think it's worth having a conversation about how to segregate this properly - i.e. you've sub-divided these stages into `cpg_workflows.stages.XXX`, which feels like a very partial solution. That puts us in a weird position where we're trying to separately version a few files inside cpg_workflows, which is itself versioned at the top level. 

There's a broader issue of how this pipeline is run - we start the driver image by doing a git clone of `production-pipelines` into the VM, then we run main.py. That's fine. But when we import `cpg_workflows.XXX` it imports from the relative path `cpg_workflows.XXX` that we just cloned into, not the installed version of the codebase (because AFAIK local filepaths are traversed first when Python locates modules). That leads to 'fun' issues where the version of the code being run is based on the current commit, up until we run a python job or a query_command - those are executed in a new VM without the git clone, so they read from the installed version of the codebase (potentially a different version of the codebase). I don't think that should really be addressed by this change, but we should resolve this ambiguity.

A blue sky thinking approach to this division:
```
- src
  - alignment_workflow
    - stages
      - align
      - cram_qc
    - version.py
  - genotype_workflow
    - stages
      - genotype
      - gvcf_qc
    - version.py
  - gatk_sv_workflow
    - stages
    - version.py
  - large_cohort_workflow
    - stages
    - version.py
  - rare_disease_workflow
    - stages
    - version.py
  - ...? other pipelines we want to version separately
  - cpg_workflows (the library which represents the abstract pipeline - interactions with metamist, setting up a workflow graph, etc.)
- main.py (imports from all the relevant INSTALLED pipelines and builds workflows from them)
- tests
- pyproject.toml
```

- This would enable us to ensure we're only using the installed version of the code (`from cpg_workflows.stages.XXX import YYY` just doesn't work, there's no local path `.cpg_workflows`, so we know we're using the installed code)
- This would enable use to split out the pipeline framework from the stages that rely on it
- This would enable use to version each coherent pipeline separately

Undoubtably this is more effort, it would involve reorganising the production pipelines codebase and having to build a new version of the cpg_workflows image for each test run in GCP.

However, it would resolve a fundamental ambiguity in our production pipelines workflows - `when we run our pipelines, we don't truly know which version of the code we're running`. That is a very stupid problem to have, when this whole effort is around versioning our pipelines more specifically. We ARE using cloud compute, so we ARE using containerised pipelines. With our current structure, we're using some of the code in the container, and some pulled in from github at runtime.

This is kinda twinned with https://github.com/populationgenomics/production-pipelines/issues/647, a pet peeve of mine - we're aware, and kind of.. OK? with not really knowing exactly what code is going to run in our main pipeline.

MattWellie:
As a side note, this is a huge leap considering that AFAIK the RFC/scoping doc [here](https://docs.google.com/document/d/1GF2Yxc6p9S5YZtjkd8wVzIoI3aa5ii9K_FCQK-KrO44/edit#heading=h.ktjt9ukaj2pc) has not been reviewed by anyone in software, and there are a number of discussion points identifying that @vivbak and other members of data/software should be consulted before going ahead with implementation.

Issue #585, opened 2024-01-05
Metamist timeouts killing pipelines


MattWellie:
e.g. https://batch.hail.populationgenomics.org.au/batches/431584

(I've seen this happen before, so the issue is based on that. For this linked batch I'm not sure if this correctly characterises the issue, as startup was fine, and every registration job has failed)

It looks like when metamist is overloaded (?) we see timeouts when either trying to start a pipeline run, or registering files following completion. Failing when registering output files is tricky, as the pipeline is written in such a way that an extant output file will prevent a stage being repeated, even if the output file isn't registered in the database - this leads to a gap in metamist that either needs to be filled manually, or by forcing a stage to re-run.

Pipeline side:
- Include some (more) repeat/backoff in the registration call?
- Upon failure dump the intended call into a temp file?
- Something tricky so that the pipeline can carry out just the registration step if the file exists

Metamist side:
- Bump metamist server resources?

MattWellie:
Example traceback
```
Traceback (most recent call last):
  File "<string>", line 40, in <module>
  File "<string>", line 32, in <module>
  File "/cpg_workflows/status.py", line 73, in complete_analysis_job
    a_id = aapi.create_analysis(project=project_name, analysis=this_analysis)
  File "/usr/local/lib/python3.10/site-packages/metamist/api/analysis_api.py", line 773, in create_analysis
    return self.create_analysis_endpoint.call_with_http_info(**kwargs)
  File "/usr/local/lib/python3.10/site-packages/metamist/api_client.py", line 848, in call_with_http_info
    return self.api_client.call_api(
  File "/usr/local/lib/python3.10/site-packages/metamist/api_client.py", line 409, in call_api
    return self.__call_api(resource_path, method,
  File "/usr/local/lib/python3.10/site-packages/metamist/api_client.py", line 203, in __call_api
    raise e
  File "/usr/local/lib/python3.10/site-packages/metamist/api_client.py", line 196, in __call_api
    response_data = self.request(
  File "/usr/local/lib/python3.10/site-packages/metamist/api_client.py", line 463, in request
    return self.rest_client.PUT(url,
  File "/usr/local/lib/python3.10/site-packages/metamist/rest.py", line 276, in PUT
    return self.request("PUT", url,
  File "/usr/local/lib/python3.10/site-packages/metamist/rest.py", line 223, in request
    raise ServiceException(http_resp=r)
metamist.exceptions.ServiceException: (504)
Reason: Gateway Timeout
[...]

PR #784, opened 2024-06-05
TASK: Robust Metamist Integrations

milo-hyben:
This PR is implementing items for task **Robust Metamist Integrations** on the road map.

> Add a re-try mechanism to catch failing to connect to metamist when writing analysis entries.
> Gracefully catch and handle failures, when writing to metamist fails.
> Add integration tests to validate these behaviours.

### TODO

- [x] Increase test coverage for metamist.py (currently on main 42.8%) to as close to 100% as possible
WIP, this PR latest coverage at 76%
- [ ] ~~Two functions in metamist are with comments 'not used anywhere' (process_existing_analysis, find_joint_calling_analysis), Should we remove? It would make test coverage at 91% after this PR.~~
- [x] Make API call more robust, esp. differentiate between thrown API Exception
It seems we just need to differentiate between ServiceException and the rest of ApiException.
ServiceException can be retried or other we should stop.
- [x] If getting HTTP 504 (Gateway Timeout), try to reconnect, addressing this issue:
https://github.com/populationgenomics/production-pipelines/issues/585
Error Details: It fails on 
```
"/cpg_workflows/status.py", line 73, in complete_analysis_job
 a_id = aapi.create_analysis(project=project_name, analysis=this_analysis)
```
Exception: 
```
ServiceException(http_resp=r)
metamist.exceptions.ServiceException: (504)
Reason: Gateway Timeout
```
    
- [x] investigate the logic in update_analysis, seems to always set the status, even when API fails:
https://github.com/populationgenomics/production-pipelines/blob/f4801c6804c56f865f2047019ad679f9c59b7de4/cpg_workflows/metamist.py#L318
Solution: We are ging to keep the existing behaviour
- [x] During writing unit tests discover this condition is wrong, missing brackets around or, basically it would never throw MetamistError for invalid index file extension:
https://github.com/populationgenomics/production-pipelines/blob/f4801c6804c56f865f2047019ad679f9c59b7de4/cpg_workflows/metamist.py#L635


vivbak:
> Two functions in metamist are with comments 'not used anywhere' (process_existing_analysis, find_joint_calling_analysis), Should we remove? It would make test coverage at 91% after this PR.

Yes, but could we do it in a separate PR (merged before this one). If something goes wrong it's easier to track down the cause!

cpg_workflows/status.py
```diff
@@ -35,12 +35,10 @@ def complete_analysis_job(
         update_analysis_meta (Callable | None): function to update analysis meta
         tolerate_missing (bool): if True, allow missing output
     """
-    import traceback
 
     from cpg_utils import to_path
-    from metamist.apis import AnalysisApi
```
vivbak, line 41: Nice! Gross that we used to do this here.

cpg_workflows/status.py
```diff
@@ -35,12 +35,10 @@ def complete_analysis_job(
...
-    from metamist.exceptions import ApiException
-    from metamist.models import Analysis, AnalysisStatus
+
+    from .metamist import AnalysisStatus, get_metamist
```
vivbak, line 41: We just need to keep an eye on this it case it creates any pickling issues. Let's try a test run later!

cpg_workflows/metamist.py
```diff
@@ -259,6 +266,49 @@ def __init__(self) -> None:
...
+        try:
+            return self.make_retry_aapi_call(api_func, **kwargv)
+        except (ServiceException, ApiException) as e:
+            # Metamist API failed even after retries
+            # log the error and continue
+            traceback.print_exc()
+            logging.error(
+                f'Error: {e} Call {api_func} failed with payload:\n{str(kwargv)}',
+            )
+        # TODO: discuss should we catch all here as well?
```
ckrishk, line 306: certainly a good idea.

test/test_metamist.py
```diff
@@ -0,0 +1,606 @@
...
+        timeout_counter.increment()
+        raise ServiceException(
+            status=504,
+            reason='Gateway Timeout',
+            http_resp=None,
+        )
+    return True
+
+
+def mock_aapi_update_analysis_timeout_fails(*args, **kwargs):
```
ckrishk, line 127: curious what does `aapi` mean ?

cpg_workflows/metamist.py
```diff
@@ -259,6 +266,49 @@ def __init__(self) -> None:
         self.default_dataset: str = get_config()['workflow']['dataset']
         self.aapi = AnalysisApi()
 
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=3, min=8, max=30),
+        retry=retry_if_exception_type(ServiceException),
```
ckrishk, line 272: is there a way for us to be alerted if too many of these exceptions are raised?

ckrishk (approved):

cpg_workflows/metamist.py
```diff
@@ -314,13 +364,15 @@ def update_analysis(self, analysis: Analysis, status: AnalysisStatus):
...
+            analysis_id=analysis.id,
+            analysis_update_model=models.AnalysisUpdateModel(
+                status=models.AnalysisStatus(status.value),
+            ),
+        )
+        # Keeping this as is for compatibility with the existing code
+        # However this should only be set after the API call is successful
         analysis.status = status
 
     # NOTE: This isn't used anywhere.
```
MattWellie, line 378: 🙃

cpg_workflows/metamist.py
```diff
@@ -635,11 +695,8 @@ def parse_reads(  # pylint: disable=too-many-return-statements
         index_location = None
         if reads_data[0].get('secondaryFiles'):
             index_location = reads_data[0]['secondaryFiles'][0]['location']
-            if (
-                location.endswith('.cram')
-                and not index_location.endswith('.crai')
-                or location.endswith('.bai')
-                and not index_location.endswith('.bai')
+            if (location.endswith('.cram') and not index_location.endswith('.crai')) or (
+                location.endswith('.bai') and not index_location.endswith('.bai')
```
MattWellie, line 699:
```suggestion
                location.endswith('.bam') and not index_location.endswith('.bai')
```

It looks like the original version of this contained a typo, but we never found out because we don't use BAMs

cpg_workflows/status.py
MattWellie, line 41:
```suggestion
    from cpg_workflows.metamist import AnalysisStatus, get_metamist
```

I get that relative imports are fine (now? though previously were recommended against), but as it's part of the package is there a reason not to be explicit?

MattWellie (commented):
Pretty cool! One typo we should fix (not newly introduced, but carried over)

Stylistically there's a lot of splitting across multiple lines that's definitely not required (prod-pipes linting uses a max line length of 120 chars now)

I personally hate relative imports, though I don't think that's functionally important

For that one method that's not used anywhere in the codebase, suggesting that we just delete instead of converting it?

cpg_workflows/metamist.py
vivbak: I suggested we delete the code in a separate PR to not increase the scope here :)

cpg_workflows/metamist.py
milo-hyben: Updated

cpg_workflows/status.py
milo-hyben: Good point, updated.

cpg_workflows/metamist.py
milo-hyben: We can revisit during Notification & Logging task, at the moment it won't be logged.

PR #785, opened 2024-06-07
Add resource override config option for genomics DB import memory

EddieLF:
Adds a config option to set the requested memory for GATK `GenomicsDBImport` jobs.

---

### NOTE
Current code has this comment about the memory request:
```python
    # The Broad: The memory setting here is very important and must be several
    # GiB lower than the total memory allocated to the VM because this tool uses
    # a significant amount of non-heap memory for native libraries.
    xms_gb = 8
    xmx_gb = 25
```
Then, the memory is requested from a `STANDARD` machine type:
```python
    STANDARD.set_resources(
        j,
        nthreads=nthreads,
        mem_gb=xmx_gb + 1,
        storage_gb=20,
    )
```

1. Should this PR also add in the option to request a `HIGHMEM` [machine type](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/resources.py#L193) instead of `STANDARD`? 
2. The requested memory for the VM is defined as `xmx_gb + 1`, so this config option is actually tweaking the java job memory. 
Is this fine? Would a value like `xmx_gb = 63` make sense, meaning we request 64GB memory total for the job VM?

### DIFF
--- a/cpg_workflows/jobs/joint_genotyping.py
+++ b/cpg_workflows/jobs/joint_genotyping.py
@@ -233,18 +233,19 @@ def genomicsdb(
     # GiB lower than the total memory allocated to the VM because this tool uses
     # a significant amount of non-heap memory for native libraries.
     xms_gb = 8
-    xmx_gb = config_retrieve(
+    genomicsdb_import_mem_gb = config_retrieve(
         [
             'resource_overrides',
             'genomicsdb_import_mem_gb',
         ],
-        25,
+        32,  # total memory allocated to the VM
     )
+    xmx_gb = genomicsdb_import_mem_gb - 7  # 25GB heap memory by default
 
     STANDARD.set_resources(
         j,
         nthreads=nthreads,
-        mem_gb=xmx_gb + 4,
+        mem_gb=genomicsdb_import_mem_gb,
         storage_gb=20,
     )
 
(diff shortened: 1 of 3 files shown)


cpg_workflows/jobs/joint_genotyping.py
```diff
@@ -234,6 +234,13 @@ def genomicsdb(
     # a significant amount of non-heap memory for native libraries.
     xms_gb = 8
     xmx_gb = 25
+    xmx_gb = config_retrieve(
+        [
+            'resource_overrides',
+            'genomicsdb_import_mem_gb',
+        ],
+        xmx_gb,
+    )
```
MattWellie, line 243:
```suggestion
    xmx_gb = config_retrieve(['resource_overrides',  'genomicsdb_import_mem_gb'], 25)
```

this is also a replacement for the line above

MattWellie (commented):
That Broad comment looks relevant - we're only using [java memory + 1], not 'several GB' difference between VM provision and Java max

cpg_workflows/jobs/joint_genotyping.py
EddieLF: Good note.
The default is `25 + 1 gb` requested memory, which Hail Batch has been rounding up to 32GB. 
Changed it to `xmx_gb + 4 gb` so it's a bit clearer, I don't imagine the actual requested memory for each VM will change though when this config option is unset.

cassimons (approved):

EddieLF:
Also had to add in the `HIGHMEM` worker option because when requesting 64gb memory with `STANDARD` machine type, the driver will fail to create the batch:
https://batch.hail.populationgenomics.org.au/batches/455413/jobs/1
```
ValueError: Requesting more cores than available on standard machine: 18>16
```

Adding in the HIGHMEM worker option resolved this.
https://batch.hail.populationgenomics.org.au/batches/455415

PR #787, opened 2024-06-09
Fix GenotypeGVCFs HIGHMEM worker if-else option

EddieLF:
Hotfix for the GenotypeGVCFs use HIGHMEM workers config option introduced in https://github.com/populationgenomics/production-pipelines/pull/786

I got confused by the config_retrieve, not much else to say...


MattWellie (approved):
We've all done it

PR #804, opened 2024-06-26
Split Long-Read workflow

MattWellie:
splits the bam2cram an SV annotation workflows for long-read data

#801 added a process to take long-read SV data we already have on file, merge and annotate.

I put this into the Bam2Cram long-read workflow, but... that's not really appropriate.

At some point we would probably harmonise the two in an e2e long-read workflow, but right now there's no connection.

Instead of adding Stages from two separate workflows in one file, I'm splitting into a separate Stage source file


EddieLF (approved):
I think splitting these up makes sense. You're right that there's no connection currently.

We ingest the long read BAMs and VCFs from the source.

From there, we can run `bamtocram` on the BAMs to make CRAMs (duh). Independent of that, we can run this SV annotation workflow on the VCFs. 

So although the BAMs and VCFs are coming from the same long-read samples, these pipeline stages are two totally separate workflows right now doing different things independent of one another.

PR #812, opened 2024-07-03
GATK-SV transition to MultiCohorts

### LINKED ISSUE #811: GATK-SV MultiCohort Transition
Plan!

- Maintain 2 separate workflows:
  1. All the single sample calling through to batch generation - these batches become custom cohorts, so this needs to be a separate precursor workflow.
  2. Everything after that point - instead of running separate groupings as 'Cohorts', we can merge multisample 1 & 2 as a single massive workflow. The Multisample1 Stages remain as CohortStages, the Sandwich workflow and Multisample2 become MultiCohortStages.

Proposed Structure - simplest version is merging the Multisample 1 & 2 files into a single file, leaving the singlesample workflow as a separate script with its own stages, and keeping the common methods in a third file (for now, these are referenced in gCNV & GATK workflows, so maybe we should find a better naming convention/location)

MattWellie:
Closes #811 

Absolutely monstrous line count - this is almost entirely caused by merging the multisample_1 & 2 files (1530 lines deleted) into a single multisample file (1328 lines)

Other changes are the deletion of a few config files which would no longer be required:
- stop_at_filter_batch.toml
- gatk_sv_sandwich.toml
- all_batch_names.toml
- genotypebatch.toml
- gatk_sv_multisample_1/2.toml -> a unified gatk_sv_multisample.toml

Creates a new `write_ped_file` method in MultiCohort(Stage?) (mirroring the same content in CohortStage)


cpg_workflows/stages/gatk_sv/gatk_sv_common.py
```diff
@@ -214,9 +207,9 @@ def add_gatk_sv_jobs(
         out_path = expected_out_dict[key]
         if isinstance(resource, list):
             for source, dest in zip(resource, out_path):
-                cmds.append(f'gsutil cp "$(cat {source})" "{dest}"')
+                cmds.append(f'gcloud storage cp "$(cat {source})" "{dest}"')
         else:
-            cmds.append(f'gsutil cp "$(cat {resource})" "{out_path}"')
+            cmds.append(f'gcloud storage cp "$(cat {resource})" "{out_path}"')
```
MattWellie, line 205: incidental change from gsutil to gcloud, long overdue but not specifically relevant to this change

cpg_workflows/stages/gatk_sv/gatk_sv_multisample.py
```diff
@@ -0,0 +1,1328 @@
+"""
```
MattWellie, line 1: this is the result of smashing together multisample_1 (619 lines) and multisample_2 (911 lines) into a continuous workflow. Looks heinous, isn't really

cassimons (approved):

vivbak (approved):
As always, love your work Matt. So excited to see MultiCohorts in action 🎉

EddieLF:
Hey Matt, this all looks pretty great, thanks for the effort you've put in to enable Multicohort & Cohort inputs for these stages.

Just to make sure I understand, if we were to submit some kind of workflow with multiple cohorts like this:
```
[workflow]
input_cohorts = ["COH1", "COH2", "COH3",]
last_stages = ['AnnotateCohortSv']
```
and invoke the `cpg_workflows/stages/gatk_sv/gatk_sv_multisample.py` workflow with analysis runner
```
analysis-runner --config myconfig.toml ... python main.py gatk_sv_multisample
```

Is this what would happen?
- For each cohort, do:
    - `MakeCohortCombinedPed`
- Then, for all cohorts combined as a single "MultiCohort", do:
    - `MakeMultiCohortCombinedPed`
- Then, for each cohort, do:
    - `GatherBatchEvidence` -> `ClusterBatch` -> `GenerateBatchMetrics` -> `FilterBatch` 
- Then, for the single Multicohort, do:
    - `MergeBatchSites` -> `CombineExclusionLists`
- Then, for each cohort, do: 
    - `GenotypeBatch` 
- Then, for the single Multicohort, do:
    - `MakeCohortVcf ` -> `FormatVcfForGatk ` -> `JoinRawCalls` -> `SVConcordance` -> `GeneratePloidyTable` -> 
    - `FilterGenotypes` -> `UpdateStructuralVariantIDs` -> `AnnotateVcf`
    - `GenotypeBatch`

MattWellie:
> Is this what would happen?

Yeah that's the biscuit. It's a bit of an hourglass shape to it, as there's the `MergeBatchSites` Stage sitting in the middle of what's otherwise an entirely CohortStage workflow (all the Stages previously in multisample_1)

EddieLF (approved):
Fantastic!

PR #816, opened 2024-07-08
Update VCF sample id for long reads

MattWellie:
As discussed - when we loop over the SV VCF to update the fields, this will also run a string replacement, taking out the old ID and putting in the new one

### DIFF
--- a/cpg_workflows/jobs/seqr_loader_long_read.py
+++ b/cpg_workflows/jobs/seqr_loader_long_read.py
@@ -1,12 +1,18 @@
 def modify_sniffles_vcf(file_in: str, file_out: str, ext_id: str, int_id: str):
     """
-    to be run as a PythonJob - scrolls through the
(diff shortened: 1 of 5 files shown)


cpg_workflows/jobs/seqr_loader_long_read.py
```diff
@@ -1,36 +1,41 @@
-def modify_sniffles_vcf(file_in: str, file_out: str):
+def modify_sniffles_vcf(file_in: str, file_out: str, ext_id: str, int_id: str):
     """
     to be run as a PythonJob - scrolls through the
```
EddieLF, line 3: is this the full docstring?

EddieLF (approved):
Looks good to me!

cpg_workflows/jobs/seqr_loader_long_read.py
MattWellie: it absolutely was not

PR #817, opened 2024-07-08
Mt to es outside dataproc

### LINKED ISSUE #800: Replace MtToEs stages Dataproc cluster usage with Hail Batch
There are several good reasons to reduce our reliance on Dataproc clusters, but some important reasons are that they are less transparent than Hail Batch jobs, they require dedicated "start cluster", "submit to cluster", and "stop cluster" jobs to run their workflows, and are generally more cumbersome and less user friendly. 

Dataproc clusters have largely been eliminated from cpg workflows where possible, but one last holdout for them is in the [MtToEs](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/seqr_loader.py#L232) stage, as well as the [MtToEsSv](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py#L840) and [MtToEsCnv](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/gcnv.py#L605) stages. 

In each of these cases, the `dataproc_scripts/mt_to_es.py` script is invoked in the same way, to transform an existing Hail MatrixTable `mt` into a seqr ready elasticsearch index.

```python
script = (
     f'cpg_workflows/dataproc_scripts/mt_to_es.py '
     f'--mt-path {dataset_mt_path} '
     f'--es-index {index_name} '
     f'--done-flag-path {done_flag_path} '
     f'--es-password {es_password_string}'
)
...
j = dataproc.hail_dataproc_job(
      get_batch(),
      script,
      ...,
      depends_on=inputs.get_jobs(dataset)
)
```

We should look at improving this process by eliminating the Dataproc cluster. We should use Hail Batch to connect to the Elasticsearch cluster and write to it directly. The use of the plaintext es-password should also be removed.

MattWellie:
Closes #800

Tested here with a small Exome MT -> ES (2GB): https://batch.hail.populationgenomics.org.au/batches/461296

~Untested~

- We are currently locked into using the seqr-loading-pipelines code as a submodule, and DataProc as an execution environment, when building ElasticSearch indexes for Seqr
- This change takes all the methods and process that we use from seqr-loading-pipelines, skips all the content which is never triggered when we run an MTtoES transition. Once we remove all the code paths we never intend to execute, what's left is a fairly thin wrapper around the elasticsearch library and a few Hail Methods

## Process:

1. copy the target MT into the VM
2. start a Hail local Spark instance
3. generate the ES password from secrets in a config file - this removes the need to pass a secret in plain text
4. Uses the same MT -> flattened HT method as the existing script
5. Creates an ElasticSearchClient, modelled on the Hail version. This contains all the method calls we previously executed, instead of importing those methods from `seqr-loading-pipelines`
6. Removes the ES Index if it already exists (unexpected)
7. Pushes a new index by name to the ES instance, cleans up, and writes a 'DONE' file

This is complete theft:
- where we were importing/inheriting from the HailES Client, I've created an equivalent client with the same functionality (limited to only the methods/calls we actually execute)
- where we referenced constants in the SLP repo, or call methods which require constants, I've copied them in
- where SLP methods called out to Hail methods, we're using those Hail methods directly
- there's no doubt a need to more explicitly credit the client/code I'm ripping off here

some optimisation params: 
https://github.com/broadinstitute/seqr-loading-pipelines/blob/c113106204165e22b7a8c629054e94533615e7d2/hail_scripts/elasticsearch/hail_elasticsearch_client.py#L196-L206
https://www.elastic.co/guide/en/elasticsearch/hadoop/current/configuration.html

### DIFF
--- a/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
+++ b/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
@@ -22,7 +22,9 @@ def main():
     # localise the MT
     mt_name = args.mt_path.split('/')[-1]
     job.command(f'gcloud --no-user-output-enabled storage cp -r {args.mt_name} $BATCH_TMPDIR')
-    job.command(f'mt_to_es --mt_path {mt_name}, --index {args.index} --flag {args.flag} --cpu {ncpu}')
+    job.command(
+        f'mt_to_es --mt_path "${{BATCH_TMPDIR}}/{mt_name}" --index {args.index} --flag {args.flag} --cpu {ncpu}',
+    )
     get_batch().run(wait=False)
 
 
--- a/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
+++ b/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
@@ -21,7 +21,7 @@ def main():
 
     # localise the MT
     mt_name = args.mt_path.split('/')[-1]
-    job.command(f'gcloud --no-user-output-enabled storage cp -r {args.mt_name} $BATCH_TMPDIR')
+    job.command(f'gcloud --no-user-output-enabled storage cp -r {args.mt_path} $BATCH_TMPDIR')
     job.command(
         f'mt_to_es --mt_path "${{BATCH_TMPDIR}}/{mt_name}" --index {args.index} --flag {args.flag} --cpu {ncpu}',
     )
--- a/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
+++ b/cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
@@ -22,9 +22,7 @@ def main():
     # localise the MT
     mt_name = args.mt_path.split('/')[-1]
     job.command(f'gcloud --no-user-output-enabled storage cp -r {args.mt_path} $BATCH_TMPDIR')
-    job.command(
-        f'mt_to_es --mt_path "${{BATCH_TMPDIR}}/{mt_name}" --index {args.index} --flag {args.flag} --cpu {ncpu}',
-    )
+    job.command(f'mt_to_es --mt_path "${{BATCH_TMPDIR}}/{mt_name}" --index {args.index} --flag {args.flag}')
     get_batch().run(wait=False)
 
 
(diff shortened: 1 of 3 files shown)


MattWellie:
Driver job: https://batch.hail.populationgenomics.org.au/batches/461265/jobs/1
Worker Batch: https://batch.hail.populationgenomics.org.au/batches/461266

This run uses a different approach:
- make a big VM (configurable)
- gcloud cp the MatrixTable into the VM 
- process the localised MT into an ES index
- ?????
- PROFIT!

cpg_workflows/dataproc_scripts/mt_to_es_free_of_dataproc.py
```diff
@@ -0,0 +1,304 @@
...
+    flat_table = flat_table.drop(*key)
+    flat_table.describe()
+    return flat_table
+
+
+def _mt_num_shards(mt):
+    """
+    Calculate the number of shards from the number of variants and sequencing groups.
+    """
+    denominator = 1.4 * 10**9
```
cassimons, line 298: Any idea where this comes from? We do get ES warnings about large shards so I wonder if we should tweak?

cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
```diff
@@ -0,0 +1,30 @@
...
+def main():
+    parser = ArgumentParser(description='Argument Parser for the ES generation script')
+    parser.add_argument('--mt_path', help='MT path name', required=True)
+    parser.add_argument('--index', help='ES index name', required=True)
+    parser.add_argument('--flag', help='ES index "DONE" file path')
+    args = parser.parse_args()
+
+    job = get_batch().new_job(f'Generate {args.index} from {args.mt_path}')
+    job.image(config_retrieve(['workflow', 'driver_image']))
+    job.storage(config_retrieve(['workflow', 'storage_requirement'], '10Gi'))
```
cassimons, line 16: Do we need to make this dynamic even if it is a dirty heuristic based on sample size?

cpg_workflows/dataproc_scripts/mt_to_es_free_of_dataproc.py
MattWellie: I'll crawl around and reference the original source for all the bits I stole, might be able to find some explanations in doctrines

cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py
MattWellie: Should be easy enough to squeeze that in if we proceed to use this. We'd be running in a pipeline workflow context, so cohort size & exome/genome will be available info

MattWellie:
Closing as superseded by #829

PR #823, opened 2024-07-12
Cohort analysis dataset name

MattWellie:
- Currently we assign each Cohort object an attribute `analysis_dataset`
- The Dataset is created with ` Dataset(name=self.name, cohort=self)`, where `self.name` is the name of the CustomCohort
- Error: https://batch.hail.populationgenomics.org.au/batches/461943/jobs/1
- Error message: 

```
cpg_utils.config.ConfigError: Storage section for dataset "COH401" not found in config. Please check that you have permissions to the dataset. Expected section: [storage.COH401]
```

Solution
- When each CustomCohort needs to write to a bucket, it needs to access the config section `storage.XXXX[.analysis,default,tmp...]`, where XXXX has been the dataset.
- CustomCohorts are not tied to a single dataset, so this relationship doesn't hold
- analysis_dataset before CustomCohorts was the contents of `workflow.dataset`, a parent dataset with read permission on all data enclosed by the CustomCohort
- this change resumes the behaviour of cohort.anaylsis_dataset being the content of `workflow.dataset`
- this will not affect DatasetStage outputs, just CohortStages

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.8
+current_version = 1.25.9
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.25.8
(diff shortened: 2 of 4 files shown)


vivbak (approved):
Nice catch!

PR #824, opened 2024-07-12
Insert valid reference base in LR SV VCFs

MattWellie:
bcftools concat is running into issues when we merge the long-read SV files prior to annotation:

```
The REF prefixes differ: A vs N (1,1)
Failed to merge alleles at chr1:180578 in /io/batch/847ab6/inputs/A5vQV/<ONE_OF_THE_FILES>
```

- Sniffles `Insertions` contain a REF allele of "N"
- Sniffles `Deletions` contain a REF allele of the full inserted sequence
- Both use SVTYPE and SVLEN fields to determine the insertion content

I've was taking the first base of the REF string to make the VCF annotate-able, but when we merge/concat files together, the same contig & position will have a base if it was a deletion, and a N if it's an insertion, which is invalid.

This change inserts the actual reference base for the Chr & Pos when iterating over the variants.

I'm doing this using Hail - create a reference genome, add the corresponding sequence, then recall specific bases from the reference genome object ([ref](https://hail.is/docs/0.2/functions/genetics.html#hail.expr.functions.get_sequence))

## Alternative

We could also get around this by assigning `N` as the reference base consistently, and `<SVTYPE>` as the symbolic alt, but I'm not sure whether we'll run into issues downstream (Seqr? Hail?) when reading `chr:pos:N` as a valid Locus.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.9
+current_version = 1.25.10
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.25.9
+  VERSION: 1.25.10
 
 jobs:
   docker:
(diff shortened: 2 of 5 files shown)


EddieLF (approved):
Seems good to me

cassimons (approved):

PR #825, opened 2024-07-16
LR SV pythonJob fix

MattWellie:
Hail runtime requires service backend?

See https://batch.hail.populationgenomics.org.au/batches/461977/jobs/1
ValueError: No billing project.  Call 'init_batch' with the billing project or run 'hailctl config set batch/billing_project MY_BILLING_PROJECT'

Despite using localised files, it looks like running hail inside a python_job might need to be initialised with a service backend?

I'm genuinely not sure if this is the solution, I was surprised this was a requirement...

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.10
+current_version = 1.25.11
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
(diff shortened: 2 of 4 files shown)


EddieLF (approved):
I suppose when the batch is initialised in the python job, the relevant params are not passed through automatically?

Calling cpg_utils.init_batch autofills the billing project and GRCh38 reference, so this seems like the right fix

MattWellie:
> Calling cpg_utils.init_batch autofills the billing project

true, but if you're running hail on localised data you don't normally need billing details. Maybe there's something inherent in the pythonJob running in the same way as the parent batch. 

I'll give it a go, but this might be a change to revert if it doesn't solve the issue

PR #827, opened 2024-07-16
Sniffles: No more python job

MattWellie:
Screw python jobs

- create a standalone script to run the VCF modification
- runs with a hail local context specifically, leaving nothing up to chance
- the path to the script it set as an entrypoint in the cpg_workflows package - exact path to the file inside the container isn't required

The output file still goes through recompression and indexing in a subsequent job

This is a departure from the way we do a lot of things, but it removes the need to use python_jobs or query_command, leaving a simple CLI interface

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.12
+current_version = 1.25.13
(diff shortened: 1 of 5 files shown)


EddieLF (approved):
🚫 🐍

Issue #793, opened 2024-06-19
Handle Dataproc Jobs better


MattWellie:
Related to #790 

The Dataproc cluster wrapper method we use creates 3 jobs (start, run, close). Of those only the run job is passed back, so the production-pipelines stage wrapper framework can only automatically set dependencies on one job. All other dependencies need to be gathered and fed forward into the wrapper, which is against the pattern established by the pipeline framework.

Proposal - the wrapper is altered to pass back a list of all jobs, not just the 'worker' job. This removes the special case where dependencies need to be fed forward into the wrapper. This will require some changes to RD, GATK-SV, gCNV, and Large-cohort pipelines, but only a trivial change to each (I believe all stages currently using dataproc only generate a single job, and the Stage implementation equally accepts return of a job or list of jobs).

This issue is primarily with cpg-utils, but requires a counterpart here

MattWellie:
Mitigated by dropping DataProc _almost_ everywhere

PR #829, opened 2024-07-16
Implement ES export outside dataproc

### LINKED ISSUE #666: Streamline usage of Dataproc?
We still need dataproc for the MtToEs Seqr export stage. This is currently a DatasetStage, run once for each dataset (inc. the 15-30 min startup time and relevant IP costs per dataset).

If this was a cohort Stage instead, we could run all the datasets in a single Dataproc instance, mitigating the cost/time expense of starting up 15+ dataproc clusters

### LINKED ISSUE #800: Replace MtToEs stages Dataproc cluster usage with Hail Batch

MattWellie:
Actual implementation of the MT -> ES conversion/post, without DataProc

Mitigates relevance of #793 
Closes #800 
Closes #666 
Template/POC for #647
Proof of concept for https://github.com/populationgenomics/cpg-utils/issues/163

### cpg_workflows/scripts/mt_to_es_without_dataproc.py

   * this replaces `cpg_workflows/dataproc_scripts/mt_to_es.py`. In one script it includes all the code in `mt_to_es.py`, plus all the content `mt_to_es.py` imports from seqr-loading-pipelines. The original script has not been deleted.
   * Instead of being submitted to a DataProc cluster with the `seqr-loading-pipelines` submodule added as pyfiles, this operates as a stand alone script in a standard BashJob. This script requires the MT to be localised within the job (using a `gcloud` recursive copy), then operates on it in a local hail-pyspark process.
   * We do not need to pass the password to this job any more, so it is not visible as a plaintext argument
   * This process (specifically the `cli_main` method in the script, with an argparse CLI) is registered as a console entrypoint for cpg_workflows. This is a solution to the packaging problem - `mt_to_es` inside a container running a version of cpg_workflows at this point or later will link directly to the `cli_main` method, so we don't need to know the exact path to the script (i.e. is not flummoxed by executing in a container without a git clone).
   * Instead of requiring the seqr-loading-pipelines/hail_scripts submodule, this creates a private copy of all the methods we call implicitly and explicitly. We didn't use much of the module's capability in practice, so this is a pretty lean script.
       * This, and `query_modules/seqr_loader.py` are the only uses of seqr-loading-pipelines, should we follow this up by completely detaching from that submodule?

### cpg_workflows/stages/gatk_sv/gatk_sv_multisample.py

* Removed the dataproc MT -> ES conversion, instead using the new script
* Uses a 10Gi storage request (I haven't seen a MT above 2GB so far, 10Gi is the minimum if attaching any storage)
* ES Password is generated but not passed - if we can't generate it (no permission, or config section missing), we gracefully exit. There is a second 'graceful exit' in the MtToEs script itself, but for that script to run we first need to invest the time in copying the MT into the VM. Checking here first acts as a gatekeeper, then grabbing the password again in the script to authenticate prevents the need to pass the password in plaintext as a command line argument in the job log.

### cpg_workflows/stages/gcnv.py

* same as GATK-SV, never seen a MT over 500MB, so using the minimum attached storage

### cpg_workflows/utils.py

* created a new method `tshirt_mt_sizing`
* gives ballpark estimates of the attached disk size required to localise a MT
* This is based on a couple of sparse datapoints. Running a `gcloud storage du -s <MT>` on a wider range of MTs would give us more data, but... I've never seen a MT bigger than 500GB, and 500GB disk for 3.5 days only cost $20 so there's not much juice for the squeeze if we optimise this further.

This method has 3 outcomes:
- take override from config
- if low number of genomes, or medium number of exomes, 50Gi
- if a medium number of genomes, or high number of exomes, 500Gi

```
40 samples	genome	20GB
51 samples	genome	18GB
76 samples	genome	35GB
599 samples	genome	360GB
815 samples	genome	353GB
345 samples	exome	8.3GB
3489 samples	exome	283GB
```

### cpg_workflows/stages/seqr_loader.py

* removal of DataProc
* usage of the tshirt-sizing method to determine storage 
* I've removed the `create_es_index_for_datasets` ES-skipping functionality. Kinda by accident, but based on the previous seqr_loader run we should re-work this a bit so that we skip both AnnotateDataset _and_ MtToEs for the datasets where we don't need a new index. Maybe this should be based on whether the same custom cohort used as input (no new samples, no new MT/index)

### setup.py 

* adds an entrypoint for this new standalone script

### cpg_workflows/stages/talos.py

* uses the shirt sizing method when provisioning the Talos Hail VM

n.b. `cpg_workflows/dataproc_scripts/mt_no_dp_wrapper.py` should be deleted once we're happy with the results - we shouldn't need to keep generating ES indexes outside the pipeline framework


cassimons (approved):
Dataproc is dead. Long live anything not dataproc!

PR #831, opened 2024-07-18
Scrap Hail in sniffles update

MattWellie:
Ok, Hail was a bust - Using Hail to query for individual bases not only blew up the RAM consumption (the VM exploded when it tried to exceed 14GB), it also blew up the time taken (task would fail after 2 hours).

These VCFs only contain ~50k rows...

I've scrapped Hail in this job and used `pyfaidx` ([link](https://pypi.org/project/pyfaidx/)). This is just a pure python implementation of Samtools faidx, using the index file to randomly jump to a base, sequence, contig... It's a super single purpose optimised library, and it is fantastic.

After the last debacle [HERE](https://batch.hail.populationgenomics.org.au/batches/463009) where this stage spun for 2 hours and did nothing but rack up bills, I localised the VCF and tried to check if the Hail process would work on my machine. It would not. Even adding sleeps between every ~5000 variants processed allowed for some memory management in Java, but the tasks still couldn't complete. One attempt got tantalisingly close at 45 mins in, eventually failing only 1000 lines from the end of the file. 

This version with pyfaidx on the same file took 3.5 seconds. F**k Hail reference queries.

I've also dropped in a little validation - for the DELetions we have the Reference bases, so I do a quick check that the pyfaidx reference base matches with the base Sniffles already shoved in the VCF

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.14
+current_version = 1.25.15
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.25.14
+  VERSION: 1.25.15
 
 jobs:
   docker:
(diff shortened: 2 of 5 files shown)


EddieLF (approved):
Looks good and the improvement over Hail is mind blowing

PR #833, opened 2024-07-18
SVAnnotate requires single alt alleles

MattWellie:
It's a day ending in Y, so it's about time we had a LR SV commit

2 issues resolved here

- really dumb: `str.replace(A, B)` returns the new String, rather than doing an in-place change. This means that the VCFs produced so far don't have updated SG IDs in the header
  - the vcfs already created here have been recorded in metamist. They should probably be deleted on file and in metamist, but I've altered the expected output filename instead to force recreation of these intermediate files.
- less dumb? The bcftools merge was failing before due to reference base issues (now thankfully resolved), so I never got far enough to stress test the SV annotation step - the `merge` was done with `-m all` (merge all alts into a single row). That behaviour is fine with VEP, it's not fine with SVAnnotate

```
java.lang.IllegalArgumentException: Expected single ALT allele, found multiple: [<INS>, <DEL>]
```

Easy fix, just change the -m argument. No need to follow the merge with a normalisation step

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.16
+current_version = 1.25.17
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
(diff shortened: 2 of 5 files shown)


EddieLF (approved):
Just say no to new multiallelics 🙅‍♂️

PR #836, opened 2024-07-19
Separate output paths per-cohort

### LINKED ISSUE #835: CohortStages need a personalised write location
Batch ref: https://batch.hail.populationgenomics.org.au/batches/465708

This is an 8-Cohort batch, and all outputs for each cohort are writing to the same path

(`e.g. gs://cpg-PROJECT-main/gatk_sv/d91f8292cf3176f5f759c90db614cac7d5adb6_2043/GatherBatchEvidence/manta.tar.gz`)

Reason being that `self.prefix` as a method for building `DATASET_BUCKET / WORKFLOW / OUTPUT_VERSION` is at the workflow level, and the workflow as a whole only has one output_version, and one dataset.

- Need a new method for generating an output path exclusive to each Cohort

MattWellie:
Closes #835

Creates 2 new methods at the Stage & Workflow level, taking a Cohort as an argument. This constructs an output Path specific to this Cohort. Currently we have:

```
@property
def prefix(self) -> Path:
    return get_workflow().prefix / self.name
```

Which constructs `gs://BUCKET/WF_name/MultiCohortHash/StageName`, but we can't delineate output paths per cohort without building each full path manually. This PR adds:

### Workflow level
```
def cohort_prefix(self, cohort: Cohort, category: str | None = None) -> Path:
    return cohort.analysis_dataset.prefix(category=category) / self.name / cohort.name
```

### Stage level
```
def get_stage_cohort_prefix(self, cohort: Cohort, category: str | None = None) -> Path:
    return get_workflow().cohort_prefix(cohort, category) / self.name
```

The analysis_dataset here is the same, as we populate cohort.analysis_dataset from `get_config()['workflow']['dataset']` but if we later decide to specifically allocate a dataset for a Cohort, this will be automatically adopted.

The output will now look like:

`gs://DATASET_BUCKET/WORKFLOW/COHORT_ID` for the Workflow method, or `gs://DATASET_BUCKET/WORKFLOW/COHORT_ID/STAGE_NAME` for the Stage method

- semi-related change included: altered the key used in `inputs.as_path(target=multicohort, stage=MakeMultiCohortCombinedPed, key=X)` from cohort_ped (wrong) to `multicohort_ped`.


EddieLF (approved):
Looks good, approving as discussed on call

PR #844, opened 2024-07-23
Removing options for biobambam as tool for marking duplicates

michael-harper:
Relates to #871 
We are wanting to simplify the alignment job and remove unsupported and unused configurations. This will help future efforts for version control.

This PR removes the option to use `Biobambam` to mark duplicates in favour of `picard`.


MattWellie (approved):
LGTM, easy little nibble at the bigger problem. Can you add a bump2version patch version update, and just check in with @katiedelange /PA before merging

PR #845, opened 2024-07-24
Createsamplebatches json sgs

EddieLF:
Updates the `CreateSampleBatches` stage so that it now writes the `{sg['id']: sg['meta']}` dict to a file in the tmp bucket, and passes this file path to the partition_batches function. This is because the sg meta dict was too large to pass through the `pyjob.call`. Writing the data to a json in tmp and then reading from the json in the job will allow for better scaling.


MattWellie (approved):
LGTMatt

PR #846, opened 2024-07-24
Add config options for MultiQC jobs

EddieLF:
Adds new config options for toggling the Slack message sent by the multiQC jobs, as well as adding the option for additional config parameters passed to the multiQC job. These can be retrieved from the AR config and will be passed to the job through the existing `extra_config` variable. See https://multiqc.info/docs/getting_started/config/ for the options we could use.

The default for cram_multiqc was `{'table_columns_visible' = {'FastQC': False}}` and the default for gvcf_qc was `{'table_columns_visible' = {'Picard': True}}`. These options have been retained and are passed into the extra_config dict by default, alongside any other options retrieved from the AR config. 

This change will hopefully allow us to create multiQC reports with more samples without disabling exports / forcing the plots into a static or "flat" format. This is disabled for reports with many samples (maybe >100?), but can be enabled with the `plots_force_interactive=True` option. See https://multiqc.info/docs/getting_started/config/#force-interactive-plots

```
[workflow]
[workflow.cram_multiqc]
send_to_slack = true
[workflow.cram_multiqc.extra_config]
plots_force_interactive = true

[workflow.gvcf_multiqc]
send_to_slack = true
[workflow.gvcf_multiqc.extra_config]
plots_force_interactive = true
```


cassimons (approved):
Looks good once you please the linting gods. Thanks for figuring this out.

PR #848, opened 2024-07-25
Genotypegvcfs exome merge

EddieLF:
Adds the `--merge-input-intervals` flag back to the genotypegvcfs jobs for exome joint calling.

Since exomes don't have data in the centromeres, we should feel safe to merge all input intervals when running genotypegvcfs. This flag seems to impact the performance of the genotypegvcfs jobs. 

Without the flag, 1200 scatter count, performance is not very good: https://batch.hail.populationgenomics.org.au/batches/468330?q=name%3D%7Egenotypegvcfs

With the flag, 800 scatter count, performance is fine:
https://batch.hail.populationgenomics.org.au/batches/443610?q=name%3D%7Egenotypegvcfs

Note these two runs have relatively similar number of exomes.

Also noting exome gvcfs were genotyped with the standard exome interval list, which DOES include centromeric regions for some chromosomes. So the gvcfs do have some centromere regions baked into them, which is probably why the performance gets worse when we don't merge the input intervals.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.21
+current_version = 1.25.22
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.25.21
+  VERSION: 1.25.22
 
 jobs:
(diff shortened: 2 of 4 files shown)


MattWellie (approved):
From discussion I'm fairly confident that the exomes shouldn't have anywhere near as much data around telo/centromeres due to the capture, so they shouldn't be prone to the runtime explosions we saw in genomes (and that matches up to the results from previous runs)

PR #851, opened 2024-07-25
Add type, technology, platform to sequencing-groups + simplify logic

illusional:
Noticed the sequencing_type wasn't on the SequencingGroup object, so:

- Add this (and the technology / platform), 
- This is already present on the GraphQL query,
- Simplify `alignment_inputs_by_seq_type` to just `alignment_input` as a sequencing group can by definition only be one type.
- Require keyword arguments when constructing a sequencing_group using the helpers, to remove ambiguity of arg order
- Add these fields to a bunch of tests,
- hopes and prayers


illusional:
FYI @michael-harper, merging this in will cause merge conflicts with your branch

MattWellie (approved):
Looks good, makes sense

vivbak (approved):

Issue #852, opened 2024-07-25
DatasetStages don't work in a MultiCohort world


MattWellie:
https://batch.hail.populationgenomics.org.au/batches/471684/jobs/1

This run contains 14 Cohorts, and ~15 Datasets, but when it got to the AnnotateDataset DatasetStage it was trying to queue ~150 different DatasetStages. This is because the current implementation of DatasetStage when using a MultiCohort is done by iterating over each Cohort, then over each Dataset in each Cohort.

https://github.com/populationgenomics/production-pipelines/blob/a3dfe6172d8c3ff191b17ca49756225564aceadb/cpg_workflows/workflow.py#L1386-L1391

This needs to be restructured so that each DatasetStage runs once per Dataset (collected across all Cohorts). This is because the Dataset concept is not hierarchically 'below' the Cohort, it's a different type of grouping, so when finding Datasets we shouldn't be looking one Cohort at a time, we should be looking across all samples in this analysis, i.e.

```
MultiCohort contains COH1 & COH2

COH1 contains 
- Dataset1: CPGA, CPGB
- Dataset2: CPGC

COH2 contains 
- Dataset1: CPGD, CPGE
- Dataset2: CPGF

When a DatasetStage runs, it should run: 
- once on Dataset1, the collection of SGs [CPGA, CPGB, CPGD, CPGE]
- once on Dataset2, the collection of SGs [CPGC, CPGF]

Instead of the current behaviour:

- once on Dataset1, SGs [CPGA, CPGB]
- again on Dataset1, SGs [CPGD, CPGE]
- once on Dataset2, the collection of SGs [CPGC]
- again Dataset2, SGs [CPGF]
```

MattWellie:
I've poked around in this, and it's actually going to be a lot of work to resolve...

Currently the structure we have is

```
MultiCohort {1}
-- Cohort {1, many}
---- Dataset {1, many}
------ SequencingGroup {1, many}
```

(IMO) This needs to be restructured to 
```
MultiCohort {1}
-- Cohort {1, many}
---- SequencingGroup {1, many}
-- Dataset {1, many}
---- SequencingGroup {1, many}
```

[...]

PR #853, opened 2024-07-25
DatasetStage concept change

MattWellie:
This might solve #852
It might not,

I mostly just want the CI tests to run so I can think about it a bit more


MattWellie:
Closing, #853 is a better attempt

PR #857, opened 2024-07-28
Some suggested sequencing type simplifications

illusional:
Use the new `sequencing_group.sequencing_type` attribute rather than fetching from the config.
These are just soft, take-it-or-leave-it suggestions, I'm not personally tied to them, just thought it would be interesting to see what it might look like.

Stacked on #851, but I'd advise against merging into there, and I'd just rebase when that gets merged if we're interested in these changes.

### DIFF
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -71,9 +71,12 @@ def run(out_vds_path: Path, tmp_prefix: Path, *sequencing_group_ids) -> hl.vds.V
     sequencing_groups = _check_gvcfs(sequencing_groups)
     seq_types = set(s.sequencing_type for s in sequencing_groups)
     if len(seq_types) > 1:
-        sequencing_type = config_retrieve(['workflow', 'sequencing_type'])
-        logging.warning(
-            f'Found multiple sequencing types for large cohort combiner: {seq_types}, using the config-provided {sequencing_type}',
+        by_seq_type = collections.defaultdict(list)
+        for s in sequencing_groups:
+            by_seq_type[s.sequencing_type].append(s.id)
+
+        raise ValueError(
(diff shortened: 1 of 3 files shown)


cpg_workflows/large_cohort/combiner.py
```diff
@@ -70,6 +69,14 @@ def run(out_vds_path: Path, tmp_prefix: Path, *sequencing_group_ids) -> hl.vds.V
         sequencing_groups = [s for s in sequencing_groups if s in sequencing_group_ids]
 
     sequencing_groups = _check_gvcfs(sequencing_groups)
+    seq_types = set(s.sequencing_type for s in sequencing_groups)
+    if len(seq_types) > 1:
+        sequencing_type = config_retrieve(['workflow', 'sequencing_type'])
+        logging.warning(
+            f'Found multiple sequencing types for large cohort combiner: {seq_types}, using the config-provided {sequencing_type}',
+        )
```
MattWellie, line 77: this one is spicy, for now I reckon this should be a log/fail with a ValueError. There might be some situations where we would mix sequencing types but probably not this one

MattWellie (commented):
In general it's logical. For genotype/align when they're split out as per-SG workflows we could run everything, and each SG would have the appropriate jobs queued according to its type.
Iffy feelings about the combiner one though, that feels like cause for failure at the moment

MattWellie (approved):
LGTMatt

cpg_workflows/large_cohort/combiner.py
illusional: Yeah agreed, upgraded to an error

illusional:
Hey @MattWellie, do you want to merge this PR - otherwise I might close.

MattWellie:
> Hey @MattWellie, do you want to merge this PR - otherwise I might close.

I like it, just gonna smash merge on it. I was waiting for @vivbak to review as well, but I think this is straightforward enough

PR #860, opened 2024-07-29
DatasetStages run once for each Dataset in the MultiCohort

### LINKED ISSUE #852: DatasetStages don't work in a MultiCohort world

MattWellie:
Closes #852 

## Description

Since the introduction of MultiCohorts our DatasetStages haven't worked right. Instead of the 'Dataset' Target meaning "All samples in this Dataset", it shifted to "in each Cohort, stratify samples by Dataset, then run each group separately"

In practice this means that if you have 2 Datasets in total, equally split across 2 Cohorts, we run 4 DatasetStages, none of them containing the full Dataset. An additional consequence is that a DatasetStage typically writes to a path `BUCKET/PIPELINE/STAGE/MultiCohortHash/FILENAME` - when we fire off multiple DatasetStages, each with a portion of the overall Dataset, we will attempt to write to the same file path multiple times.

The core issue here is that Cohorts and Datasets need to be at the same level in the hierarchy - a MultiCohort can contain multiple of each, and each can contain multiple SequencingGroups, but there is no real relationship between the two. A `Cohort` is every SG within a CustomCohort in metamist, and a `Dataset` is every SG within a metamist project.

Doc for extended context: https://docs.google.com/document/d/1Zu4QFxuR44M1alRZtA5M9WGROLEUg74gQWorDQjOvTs/edit?usp=sharing

## What This Change Does

1. Currently the MultiCohort contains a dictionary of Cohorts by name, and each Cohort contains a list of Datasets by name. This adds a MultiCohort level Datasets-by-name dictionary.
2. When setting up a MultiCohort we query for each Cohort in turn and populate them with SG objects.
3. Once we've queried for each Cohort in turn we iterate over all the Cohorts and pool the SG objects by Dataset as well.
4. This adds an `add_sequencing_group_object` method that takes the already populated SG object, without re-creating the same object from its raw data
5. `MultiCohort.get_datasets` iterates over Datasets directly, instead of iterating over each Cohort, then each Dataset within that Cohort

This represents a minimal change so that we can restore DatasetStage functionality, without getting too into the weeds about how we should optimise this change. I've scattered a number of TODO statements over the code where we can improve on this change through optimisation. Some methods can be removed, some methods can be simplified, and some Metamist queries can be moved so that we make the minimum number of GraphQL queries.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.25.26
+current_version = 1.26.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.25.26
+  VERSION: 1.26.0
 
 jobs:
   docker:
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -44,8 +44,6 @@ def get_multicohort() -> Cohort | MultiCohort:
         if not isinstance(custom_cohort_ids, list):
             raise ValueError('input_cohorts must be a list')
         return actual_get_multicohort()
-    # TODO (mwelland): this should also return a MultiCohort object, containing one Cohort for the whole run?
-    # TODO (mwelland): this would mean we're always returning a MultiCohort, even if there is only one cohort
     return deprecated_get_cohort()
 
 
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -61,17 +61,20 @@ def create_multicohort() -> MultiCohort:
     for cohort_id in custom_cohort_ids:
         cohort = multicohort.create_cohort(cohort_id)
         sgs_by_dataset_for_cohort = datasets_by_cohort[cohort_id]
+        # TODO (mwelland): future optimisation following closure of #860
+        # TODO (mwelland): this should be done one Dataset at a time, not per Cohort
+        # TODO (mwelland): ensures the get-analyses-in-dataset query is only made once per Dataset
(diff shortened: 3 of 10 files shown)


MattWellie:
@EddieLF  when this is merged you should be good to go with the last of the GATK-SV pipeline

vivbak (approved):

PR #861, opened 2024-07-29
Stripy loci update

EddieLF:
Updates `defaults.toml` `[stripy.target_loci]` list with some new loci compatible with the latest release (v2.5) of STRipy. 

The new loci:
```
'CSTB', 'EIF4A3', 'PRNP', 'RILPL1', 'VWA1', 'ZFHX3'
```

Tested with the latest STRipy release in a batch here:
https://batch.hail.populationgenomics.org.au/batches/474495

Also updates the analysis creation which is sometimes breaking when the STRipy score is cast to an int. If something goes wrong for any of the loci, the score for those loci are "NA". 

Though maybe we want it to break in this situation?


MattWellie (approved):
Not keen to fail during the analysis write as it won't cause STRipy to re-run, but would leave us without a metamist entry. Are these NAs random within a run, or will we potentially get NAs for all loci in a run?

EddieLF:
Seems to fail when there's good reason to. E.g. I tried to run some STRipy jobs with new loci on the old stripy image, before those loci were added to the package. 
https://batch.hail.populationgenomics.org.au/batches/474454

The stripy jobs succeeded overall, but the analyses wouldn't run for the new loci, so they got scores of NA. So it seems like these NAs are limited to any loci which the stripy job failed to successfully process.

cassimons:
I think we should throw an error on a bad loci rather than fail silently. As you say, it should not happen often but when it does we probably need to know about it and address it (unless I am misunderstanding the error mode here).

EddieLF:
@cassimons when some loci fail, the stripy job is still succeeding overall. It's just in the creation of the Metamist analysis that it fails, as we try to cast `score="NA"` to integer.

cassimons:
On reflection, I agree there is not really a nice way to handle this. Lets just run with the solution you have in this PR now. Thanks Ed.

PR #862, opened 2024-07-30
gCNV update to use MultiCohorts

MattWellie:
Inserts cross-cohort merge, moves to MultiCohorts

The majority of the workflow would be operated as CohortStages (based around a CustomCohort, unified on library)

Prior to annotation we merge all the individual Cohort VCFs into one fat VCF

Post Annotation and AnnotateCohort/Dataset we then break out the whole MultiCohort by Dataset to make ES indexes. This solves the issue* we had where we produce multiple non-overlapping gCNV indexes and leave @EddieLF with extra wrangling to add them to Seqr,


cpg_workflows/stages/gcnv.py
```diff
@@ -519,87 +547,82 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
-            'tmp_prefix': str(self.tmp_prefix / dataset.name),
-            'mt': (dataset.prefix() / 'mt' / f'gCNV-{get_workflow().output_version}-{dataset.name}.mt'),
-        }
+        return {'mt': dataset.prefix() / 'mt' / f'gCNV-{get_workflow().output_version}-{dataset.name}.mt'}
 
     def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None:
         """
-        Subsets the whole MT to this cohort only
-        Then brings a range of genotype data into row annotations
+        Subsets the whole MT to this cohort only, then brings genotype data into row annotations
```
EddieLF, line 601:
```suggestion
        Subsets the cohort MT to this dataset only, then brings genotype data into row annotations
```

EddieLF (approved):
Looks great @MattWellie! One minor comment on the AnnotateDatasetCNV queue_jobs docstring, but it might be repetitive of the class docstring anyway.

cassimons (approved):

PR #863, opened 2024-07-30
Seqr loader to MultiCohorts

MattWellie:
n.b. a test is failing here - that is addressed in #865, but requires a slight restructure and should be evaluated separately

- Changes everything after Genotype and Align to use MultiCohorts
- In general removes a lot of repeat calls to expected_outputs. Instead of:

```
output1 = self.expected_outputs(multicohort)['foo']
output2 = self.expected_outputs(multicohort)['bar']
output3 = self.expected_outputs(multicohort)['baz']
```

I've replaced this with one call to get the dict, and pulling values from it instead:
```
outputs = self.expected_outputs(multicohort)
output1 = outputs['foo']
output2 = outputs['bar']
output3 = outputs['baz']
```

- In general the convention in these Stages of the expected_outputs containing an output path as a `Path` (so that the pipeline will check for presence/absence), and passing a tmp path as a String (so the pipeline won't check it) has been removed - there's no need for tmp_path to be in the output dict, it's created as needed using `self.tmp_prefix` (a property of the Stage itself, not exclusive to the expected_outputs method)

---

JointGenotyping
- Simple, one Stage, turns it into MultiCohorts
- Uses MultiCohorts.get_sequencing_groups instead

VQSR
- Simple(ish) 1-stage workflow
- number_of_sgids is only generated/counted once
- moves from Cohort -> MultiCohort
- The tmp_path here is actually not in tmp. I've not changed the convention here but *maybe we should*

VEP
- Simple(ish) 1-stage workflow
- number_of_sgids is only generated/counted once
- moves from Cohort -> MultiCohort
- The tmp_path here is actually not in tmp. I've not changed the convention here but *maybe we should*

seqr_loader
- Removes some update_analysis_meta callables. I'm ambivalent about this, but I don't think these currently work, and I don't think they're needed
- AnnotateCohort -> MultiCohort
- AnnotateDataset remains as Dataset, but picks up results from `dataset.cohort.multicohort` instead of `dataset.cohort`
- Removes `_sg_vcf_meta` which is imported in (and not really required in) HappyValidation

happy_validation
- Remove the `_sg_vcf_meta` update_analysis_meta callable

### DIFF
--- a/cpg_workflows/stages/joint_genotyping.py
+++ b/cpg_workflows/stages/joint_genotyping.py
@@ -62,7 +62,7 @@ class JointGenotyping(MultiCohortStage):
             b=get_batch(),
             out_vcf_path=outputs['vcf'],
             out_siteonly_vcf_path=outputs['siteonly'],
-            tmp_bucket=self.prefix / 'tmp',
+            tmp_bucket=self.tmp_prefix / 'tmp',
             gvcf_by_sgid=gvcf_by_sgid,
             tool=(
                 joint_genotyping.JointGenotyperTool.GnarlyGenotyper
--- a/cpg_workflows/stages/joint_genotyping.py
+++ b/cpg_workflows/stages/joint_genotyping.py
@@ -6,48 +6,39 @@ import logging
 
 from cpg_utils import to_path
 from cpg_utils.config import get_config
+from cpg_utils.hail_batch import get_batch
 from cpg_workflows.filetypes import GvcfPath
 from cpg_workflows.jobs import joint_genotyping
-from cpg_workflows.workflow import (
-    Cohort,
-    CohortStage,
-    StageInput,
-    StageOutput,
-    WorkflowError,
-    stage,
-)
+from cpg_workflows.targets import MultiCohort
+from cpg_workflows.workflow import MultiCohortStage, StageInput, StageOutput, WorkflowError, stage
 
-from .. import get_batch
 from ..resources import joint_calling_scatter_count
 from .genotype import Genotype
 
 
 @stage(required_stages=Genotype)
-class JointGenotyping(CohortStage):
+class JointGenotyping(MultiCohortStage):
     """
     Joint-calling of GVCFs together.
     """
 
-    def expected_outputs(self, cohort: Cohort) -> dict:
+    def expected_outputs(self, multicohort: MultiCohort) -> dict:
         """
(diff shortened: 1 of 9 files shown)


cpg_workflows/stages/seqr_loader.py
```diff
@@ -150,26 +93,19 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
         assert dataset.cohort
         mt_path = inputs.as_path(target=dataset.cohort, stage=AnnotateCohort, key='mt')
```
EddieLF, line 94:
```suggestion
        assert dataset.cohort.multicohort
        mt_path = inputs.as_path(target=dataset.cohort.multicohort, stage=AnnotateCohort, key='mt')
```

EddieLF (approved):
Changes look good to me, cheers @MattWellie.

cpg_workflows/stages/joint_genotyping.py
```diff
@@ -59,19 +50,19 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
         ]
 
         jc_jobs = joint_genotyping.make_joint_genotyping_jobs(
             b=get_batch(),
-            out_vcf_path=vcf_path,
-            out_siteonly_vcf_path=siteonly_vcf_path,
-            tmp_bucket=to_path(self.expected_outputs(cohort)['tmp_prefix']),
+            out_vcf_path=outputs['vcf'],
+            out_siteonly_vcf_path=outputs['siteonly'],
+            tmp_bucket=self.prefix / 'tmp',
```
cassimons, line 65: Should we add a comment here why it is not going to the temp bucket?

cassimons (approved):
Thanks Matt. Looks good and I like the clean up!

cpg_workflows/stages/joint_genotyping.py
MattWellie: Ugh, I've made this into a real TMP path - we should review the stuff that's already being retained (e.g. [here](https://console.cloud.google.com/storage/browser/cpg-seqr-main/seqr_loader/6c662c33c550a9c69d55c909c1388cc7e4930b_3755/JointGenotyping/tmp)) - looks like we're keeping multiple terabytes of additional data per joint call that should probably have been in tmp

PR #868, opened 2024-08-01
Get N even intervals for JointGenotyping from a Cohort MT

EddieLF:
This script takes a cohort mt and a desired number of intervals `N`, and obtains approximately N equally sized regions of the genome, based on the number of rows of the cohort mt. 

First we take the number of rows of the cohort mt and divide that by `N` to get the desired size of our intervals, this value is the `interval_size`.

Then, we subset the cohort mt by chromosome, creating a matrix table for each chromosome. Each chromosome mt is then split into two matrix tables, one before and one after the centromere. From here, we iterate through the rows of the matrix table in steps of size `interval_size`. The start and end position can be collected from these row indexes through the `mt.locus.position` field. These derived intervals can be read by the JointGenotyping jobs in future runs.

### DIFF
--- a/scripts/get_even_intervals_from_mt.py
+++ b/scripts/get_even_intervals_from_mt.py
@@ -60,17 +60,17 @@ def get_telomere_and_centromere_start_end_positions(
 
 
 def get_even_intervals_from_mt(
-    mt: hl.MatrixTable,
+    ht: hl.Table,
     n_intervals: int,
     tc_intervals_filepath: str | None,
     output_intervals_path: str,
 ):
     """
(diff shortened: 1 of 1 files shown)


EddieLF:
Test run:
https://batch.hail.populationgenomics.org.au/batches/474654/jobs/1

The code seems to work, somewhat, but it also seems horribly inefficient...

MattWellie:
@EddieLF  is this closable?

EddieLF:
Thanks @MattWellie - this was superceded by https://github.com/populationgenomics/production-pipelines/pull/883

Issue #869, opened 2024-08-01
When is a DataProc not a DataProc?


MattWellie:
# PRs

#829 (closed)
#870
#878 (stacked on #870)
 
# Problem

Dataproc. It's expensive, hard to maintain, difficult to debug. However, we need it for some functionality which exists in Hail, but has not been migrated to Hail Batch. 

# Solution

We can overcome this by running Hail in a local-mode, inside cloud infrastructure. That's effectively what Dataproc allows us to do. If Dataproc can run it, Hail itself can run it, just not in Query-on-Batch. There are limitations, but also benefits.

# Purpose

This 'issue' is trying to leave a little documentation about why we might want to leave Dataproc, the ways our current scripts need to be altered to escape from dataproc, and a couple of pointers to where we've successfully extricated ourselves in the MT -> ElasticSearch process, which can be used as templates for removing more future jobs from Dataproc. 

This was partially motivated by the chaos yesterday - it seems (I might be wrong?) like Dataproc jobs are failing to run, and the reason for failure is that DP jobs pip-install a wide range of libraries at runtime. Ick.

Please feel free to ignore, untag yourself, close this issue. I just wanted to outline the things I could think of that were useful in getting out of Dataproc, so they can potentially be repeated elsewhere in the codebase.

### Benefits

1. We don't need to maintain separate images and configurations to run Dataproc jobs
2. Debugging should be SO much easier. We're no longer using separate images to run dataproc, then running at-runtime installation of packages into the container.
3. Costs down - Dataproc over-provisions resources relative to most use-cases, and the costs of procuring large storage volumes/high CPU/memory manually are still less than what Dataproc reserves anyway.

### Limitations

1. If we stop running Dataproc tasks we will likely lose the lessons we've learnt, and should we really need Dataproc again in future that might be tough
2. There's no guarantee that all tasks which run now in Dataproc will run in local-mode. So far so good, but we've got hard resource constraints (RAM, CPU) which could end up being impractical to overcome
3. Import and Export feels a bit silly. To run Hail in local-mode you have to localise the MT (or HT, or both). This ends up being cheaper (in testing so far) than running QOB/Dataproc on an equivalent process.
[...]

PR #870, opened 2024-08-02
Relatedness, no Dataproc

MattWellie:
Concept PR, related to #869 

Takes the `Relatedness` Stage of the large_cohort pipeline, and tries to remove it from Dataproc

The Relatedness Stage as it sits in `main` is a little convoluted 
- one entrypoint (run) takes a MT and a HT as input, and makes 2 HTs as output
- The PCRelate stage runs first (uses one MT, makes one HT), and is optionally skipped if the result HT already exists as a checkpoint ([here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/large_cohort/relatedness.py#L50-L58))
- The Sample Flagging stage runs next, taking the PCRelate HT and a SampleQC HT as input, making another HT

This PR splits this into two Stages - `RelatednessPCRelate` and `RelatednessFlag`, each running the separate part of that larger process.

For each of these steps there's a separate script, and a Stage. 
The script sets up a Hail runtime, runs the code, and writes the output locally
The Stage wrapper creates a VM, copys the data into the VM, runs the script, and copys the output back. 

This could all be one Stage, just as it's implemented at the moment, but breaking it up this way seems logical enough as the two methods create distinct output, and we keep both outputs separately in GCP. It also makes the interface easier, instead of having to copy multiple tables in/out of one job.

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -1,19 +1,26 @@
 import logging
 
+from functools import cache
+
 from cpg_utils import Path
 from cpg_utils.config import config_retrieve, get_config, image_path
 from cpg_utils.hail_batch import get_batch, query_command
 from cpg_workflows.targets import Cohort
-from cpg_workflows.utils import slugify
-from cpg_workflows.workflow import (
-    CohortStage,
-    StageInput,
-    StageOutput,
-    get_workflow,
-    stage,
-)
+from cpg_workflows.utils import slugify, tshirt_mt_sizing, ExpectedResultT
+from cpg_workflows.workflow import CohortStage, StageInput, StageOutput, get_workflow, stage
 
 from .genotype import Genotype
+
+
+@cache
+def relatedness_version() -> str:
+    """
+    generate the relatedness version
+    """
+    if relatedness_version := config_retrieve(['large_cohort', 'output_versions', 'relatedness'], default=None):
+        return slugify(relatedness_version)
+    else:
+        return get_workflow().output_version
 
 
 @stage(required_stages=[Genotype])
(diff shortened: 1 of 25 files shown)


cpg_workflows/stages/large_cohort.py
```diff
@@ -125,45 +131,122 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+        dense_mt_path = str(inputs.as_path(cohort, DenseSubset))
+        dense_mt_name = dense_mt_path.split('/')[-1]
+
+        # estimate required storage - we create a new HT, so provision double the storage
+        required_storage = (
+            tshirt_mt_sizing(
+                sequencing_type=config_retrieve(['workflow', 'sequencing_type']),
+                cohort_size=len(cohort.get_sequencing_group_ids()),
+            )
+            * 2
```
michael-harper: Just tried kicking off a test batch which failed because `tshirt_mt_sizing` returns a string, so `50Gi` * 2 becomes `50Gi50Gi` 😅
```suggestion
            t_shirt_size_value = tshirt_mt_sizing(
            sequencing_type=config_retrieve(['workflow', 'sequencing_type']),
            cohort_size=len(cohort.get_sequencing_group_ids()),
        ).split('Gi')[0]
        required_storage_value = int(t_shirt_size_value) * 2
        required_storage = f'{required_storage_value}Gi'
```

michael-harper:
Ran some tests using `tenk10k` data in `bioheart-test`. The results seem to be the same between DP and no-DP
[Relatedness PC Relate batch](https://batch.hail.populationgenomics.org.au/batches/474841) and [Relatedness Flag](https://batch.hail.populationgenomics.org.au/batches/474859) (noting that metamist registration did not work)
Screenshot 2024-08-06 at 12 27 55 PM

MattWellie:
> metamist registration did not work

I assumed a 'mt' analysisType existed, but... maybe it doesn't. That seems to be the foreign key error it's failing on

MattWellie:
@michael-harper just checking before I delete it, is this branch dead?

michael-harper:
@MattWellie yep all good to be closed

PR #871, opened 2024-08-04
Replacing bazam + removing non-dragmap aligners

michael-harper:
This PR combines #864 and #843 as it helps simplify the logic of removing bazam:

Bazam has a bug in it that is [summarised in this post](https://centrepopgen.slack.com/archives/C063WDT5HT4/p1718170029722189) and it has been decided to replace `bazam` with `samtools fastq`.

A [subsequent post](https://centrepopgen.slack.com/archives/C063WDT5HT4/p1720413842196729) covers covers validation of this replacement.

This PR also relates to other efforts to simplify the align job (removal of `Biobambam` https://github.com/populationgenomics/production-pipelines/pull/844), this was initially one big PR but has since been split out into these subsequent ones.

Notes
- I have included a function subset_cram() that was useful for testing purposes. This is for future use-cases
- Bubbled up the checking of crams that they have a cram.reference_assembly
- Added a config parameter realign_from_cram that allows for us to realign from the 'base' cram as the previous implementation (kept in for completeness) did not allow for this to happen.
- Removed indexing of cram/bam prior to alignment as bazam required an index

This PR makes minimal changes to the current cram retrieval logic as significant changes to versioning and config logic will likely adjust how we approach this.

### DIFF
--- a/cpg_workflows/jobs/align.py
+++ b/cpg_workflows/jobs/align.py
@@ -13,7 +13,7 @@ from hailtop.batch.job import Job
 
 from cpg_utils import Path
 from cpg_utils.config import get_config, image_path, reference_path
-from cpg_utils.hail_batch import command, fasta_res_group
+from cpg_utils.hail_batch import command, fasta_res_group, output_path
 from cpg_workflows.filetypes import (
     AlignmentInput,
     BamPath,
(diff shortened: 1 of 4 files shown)


MattWellie:
Removing my review request from here for now - pending working group decision on how best to proceed

PR #872, opened 2024-08-05
Dataset stage fix patch

MattWellie:
I was incorrect - the changes in #860 didn't fully close the DatasetStage issue. I caught the structure of the MultiCohort/Cohort, but missed the fact that the DatsetStages have their own internal queue_for_cohort method, which was iterating over each cohort to find dataset targets.

This change:
- SequencingGroupStages now iterate directly over the `multicohort.get_sequencing_groups()` output, the layers of iteration and logging about Datasets/Cohorts are dropped.
- DatasetStages now iterate directly over the `multicohort.get_datasets()` output during setup
- no longer permits a Cohort to be used when running the pipeline - MultiCohorts only. This change was already made in #860 (i.e. `deprecated_create_cohort` already returns a MultiCohort), and removing a Cohort as an option in get_workflow just solidifies that
- A whole range of tests were still creating Cohorts instead of MultiCohorts, so I've fixed those up
- This means that all the Stage Target `deprecated_queue_for_cohort` methods are inaccessible, and can safely be deleted, but I'm not doing that here

### DIFF
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -472,6 +472,7 @@ class Stage(Generic[TargetT], ABC):
         Queues jobs for each corresponding target, defined by Stage subclass.
 
         Returns a dictionary of `StageOutput` objects indexed by target unique_id.
+        unused, ready for deletion
         """
         return {}
 
@@ -1187,10 +1188,10 @@ class Workflow:
             for i, stg in enumerate(stages):
                 logging.info('*' * 60)
                 logging.info(f'Stage #{i + 1}: {stg}')
-                # NOTE: When manual dataset entry is deprecated, we can remove this
-                if isinstance(inputs, Cohort):
-                    stg.output_by_target = stg.deprecated_queue_for_cohort(inputs)
-                elif isinstance(inputs, MultiCohort):
+                # pipeline setup is now done in MultiCohort only
+                # the legacy version (input_datasets) is still supported
+                # that will create a MultiCohort with a single Cohort
+                if isinstance(inputs, MultiCohort):
                     stg.output_by_target = stg.queue_for_multicohort(inputs)
                 else:
                     raise WorkflowError(f'Unsupported input type: {inputs}')
@@ -1233,6 +1234,7 @@ class SequencingGroupStage(Stage[SequencingGroup], ABC):
     def deprecated_queue_for_cohort(self, cohort: Cohort) -> dict[str, StageOutput | None]:
         """
         Plug the stage into the workflow.
+        unused, ready for deletion
         """
         output_by_target: dict[str, StageOutput | None] = dict()
 
@@ -1270,13 +1272,6 @@ class SequencingGroupStage(Stage[SequencingGroup], ABC):
                 continue
 
             logging.info(f'Dataset {dataset}:')
-            # collect all expected outputs across all samples
-            # find all directories which will be checked
-            # list outputs in advance
-            all_outputs: set[Path] = set()
-            for sequencing_group in dataset.get_sequencing_groups():
-                all_outputs = path_walk(self.expected_outputs(sequencing_group), all_outputs)
-
             # evaluate_stuff en masse
             for sequencing_group in dataset.get_sequencing_groups():
                 action = self._get_action(sequencing_group)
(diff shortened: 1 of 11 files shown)


cpg_workflows/workflow.py
```diff
@@ -1289,54 +1284,18 @@ def queue_for_multicohort(self, multicohort: MultiCohort) -> dict[str, StageOutp
...
-
-        if not (cohorts := multicohort.get_cohorts()):
+        if not (active_sgs := multicohort.get_sequencing_groups()):
+            all_sgs = len(multicohort.get_sequencing_groups(only_active=False))
             logging.warning(
-                f'{len(multicohort.get_cohorts())}/'
-                f'{len(multicohort.get_cohorts(only_active=False))} '
-                f'usable (active=True) cohorts found in the multicohort. Check that '
-                f'`workflow.input_cohorts` is provided, and not all cohorts are skipped ',
+                f'{active_sgs}/{all_sgs} usable (active=True) SGs found in the multicohort. '
```
EddieLF, line 1290:
```suggestion
                f'{len(active_sgs)}/{all_sgs} usable (active=True) SGs found in the multicohort. '
```

EddieLF (approved):
Approving for the cpg_workflows code changes, but can't really comment on the tests disabled in `test_large_cohort.py`

MattWellie:
> can't really comment on the tests disabled in `test_large_cohort.py`

Eddie 'Attention to Detail' Formaini. Nice catch 😓 I just commented those out locally because the large_cohort tests are really long running. Re-added them, and patched up the MultiCohort mock in line with the rest of the tests

cpg_workflows/workflow.py
```diff
@@ -1270,13 +1272,6 @@ def deprecated_queue_for_cohort(self, cohort: Cohort) -> dict[str, StageOutput |
                 continue
 
             logging.info(f'Dataset {dataset}:')
-            # collect all expected outputs across all samples
-            # find all directories which will be checked
-            # list outputs in advance
-            all_outputs: set[Path] = set()
-            for sequencing_group in dataset.get_sequencing_groups():
-                all_outputs = path_walk(self.expected_outputs(sequencing_group), all_outputs)
-
```
MattWellie, line 1279: this usage of 'path_walk' was designed to collect all the expected outputs across the whole cohort, then use some kind of cached iterator to see which of those files were missing, reducing our overall number of GCP existence checks. That whole process was put in the bin when @illusional adapted the `exists()` call to be a much more elegant cached check. This was looping over all the expected_outputs, storing them in an object, and... not doing anything with it.

cpg_workflows/workflow.py
```diff
@@ -1289,54 +1284,18 @@ def queue_for_multicohort(self, multicohort: MultiCohort) -> dict[str, StageOutp
...
-                    logging.warning(
-                        f'{dataset}: '
-                        f'{len(dataset.get_sequencing_groups())}/'
-                        f'{len(dataset.get_sequencing_groups(only_active=False))} '
-                        f'usable (active=True) sequencing groups found. Check logs above for '
-                        f'possible reasons sequencing groups were skipped (e.g. all sequencing groups ignored '
-                        f'via `workflow.skip_sgs` in config, or they all missing stage '
-                        f'inputs and `workflow.skip_sgs_with_missing_input=true` is set)',
-                    )
-                    continue
```
MattWellie, line 1323: This drops the various logging statements - instead of 'get cohort, did the cohort have contents', then 'get dataset, did the dataset have contents', we're really only interested in SGs. During setup we already fail if the Cohort queries have no returned SGs, so I don't think these are ever hit

cpg_workflows/workflow.py
```diff
@@ -1383,11 +1343,11 @@ def queue_for_multicohort(self, multicohort: MultiCohort) -> dict[str, StageOutp
         Plug the stage into the workflow.
         """
         output_by_target: dict[str, StageOutput | None] = dict()
-        for cohort in multicohort.get_cohorts():
-            for dataset_i, dataset in enumerate(cohort.get_datasets()):
-                action = self._get_action(dataset)
-                logging.info(f'{self.name}: #{dataset_i + 1}/{dataset} [{action.name}]')
-                output_by_target[dataset.target_id] = self._queue_jobs_with_checks(dataset, action)
+        # iterate directly over the datasets in this multicohort
```
MattWellie, line 1346: ✨ This ✨  is the main focal change in this PR

MattWellie: Could happily drop all other changes and just include this 4-line edit

MattWellie (commented):
Some comments for discussion

vivbak (approved):

PR #877, opened 2024-08-06
Annotate cohort sv prevent null end_locus row field

EddieLF:
Recently importing SV es-indexes in to seqr is breaking the variant searches, most likely because the `xstop` field is null for some hits. This PR changes the logic used to derive the `end_locus` field so that it won't be null and matches the logic that used to be in this job, before it was changed in [this commit](https://github.com/populationgenomics/production-pipelines/commit/fbbde24ab22e53a5eaf57919d4f0052f07a566f7).

Change python `if-else` to `hl.if_else` and `mt.info.END2` existence check to `hl.is_defined` - should prevent nulls flowing through into the annotated cohort, dataset, and eventual es-index.


MattWellie:
This is the syntax that doesn't quite work. if_defined method doesn't tolerate a key not being in the schema, which creates issues with LR SV compatibility. 

#876 should tolerate this - if END2 is in the schema, do this, otherwise do something else

PR #878, opened 2024-08-06
Ancestry without DP

MattWellie:
Stacked branch, splits Ancestry into 3 separate stages. I kinda ran out of stamina on this one so it's partially implemented...

1. Add Background - runs in QOB, no need for Dataproc here. This is just a join across many different tables. This stage is optional, instead of ancestry_pca deciding whether or not to run a method inside Dataproc based on config 
2. Run PCA - local Hail, removes samples that need to be removed, and generates 3 new tables which are copied to GCP
3. Infer Ancestry - local Hail, either uses the SampleQC table, or output of the AddBackground stage if it runs (depending on config)...

None of this is polished, the `Ancestry` stage still exists as originally implemented, and AncestryPlots still uses the standard Dataproc-generated outputs as input

For discussion - a comment in cpg_workflows/large_cohort/ancestry_pca.py
- It looks like we're adding some annotations only to a VDS object, and not if we generate a dense MT from a MT. Is that intentional?

### DIFF
--- a/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
+++ b/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
@@ -164,7 +164,7 @@ def main(qc_in: str, dense_in: str, qc_out: str, dense_out: str, background_data
     # start a Hail Query (on-batch) Runtime
     init_batch()
 
-    dense_mt = hl.read_matrix_table(dense_in).select_entries('GT', 'GQ', 'DP', 'AD')
+    dense_mt = hl.read_matrix_table(dense_in).select_cols().select_rows().select_entries('GT', 'GQ', 'DP', 'AD')
     sample_qc_ht = hl.read_table(qc_in)
 
     dense_mt, sample_qc_ht = add_background(dense_mt, sample_qc_ht, background_datasets, tmp_path)
--- a/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
+++ b/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
@@ -41,11 +41,11 @@ def prepare_background_mt(data_path: str, dataset: str, qc_variants_ht: hl.Table
         raise ValueError('Background dataset path must be either .mt or .vds')
 
 
-def prepare_metadata_table(table_list: list[str], sample_qc_ht: hl.Table) -> hl.Table:
+def prepare_metadata_table(background_sample_qc_paths: list[str], sample_qc_ht: hl.Table) -> hl.Table:
     """
 
     Args:
-        table_list ():
+        background_sample_qc_paths ():
         sample_qc_ht (hl.Table):
 
     Returns:
@@ -54,7 +54,7 @@ def prepare_metadata_table(table_list: list[str], sample_qc_ht: hl.Table) -> hl.
     # annotate background mt with metadata info derived from SampleQC stage
     allow_missing_columns = config_retrieve(['large_cohort', 'pca_background', 'allow_missing_columns'], False)
     metadata_tables = []
-    for path in table_list:
+    for path in background_sample_qc_paths:
         metadata_tables.append(hl.read_table(path))
 
     metadata_table = hl.Table.union(*metadata_tables, unify=allow_missing_columns)
--- a/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
+++ b/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
@@ -105,7 +105,8 @@ def add_background(dense_mt: hl.MatrixTable, sample_qc_ht: hl.Table, tmp_path: s
             )
             logging.info(f'Finished filtering background, kept samples that are {populations_to_filter}')
         if background_relateds_to_drop := config_retrieve(
-            ['large_cohort', 'pca_background', dataset, 'background_relateds_to_drop'], False
+            ['large_cohort', 'pca_background', dataset, 'background_relateds_to_drop'],
+            False,
         ):
             logging.info(
                 f'Removing related samples from background dataset {dataset}.\n'
--- a/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
+++ b/cpg_workflows/large_cohort/scripts/ancestry_add_background.py
@@ -12,7 +12,7 @@ from cpg_utils.config import config_retrieve
 from cpg_utils.hail_batch import init_batch
 
 
-def prepare_background_mt(data_path: str, qc_variants_ht: hl.Table, tmp_path) -> hl.MatrixTable:
+def prepare_background_mt(data_path: str, dataset: str, qc_variants_ht: hl.Table, tmp_path) -> hl.MatrixTable:
     """
 
     Args:
@@ -23,7 +23,7 @@ def prepare_background_mt(data_path: str, qc_variants_ht: hl.Table, tmp_path) ->
     Returns:
         A densified MT, checkpointed
     """
-    background_mt_checkpoint_path = join(tmp_path, 'densified_background.mt')
+    background_mt_checkpoint_path = join(tmp_path, f'{dataset}_densified_background.mt')
 
     if data_path.endswith('.mt'):
         background_mt = hl.read_matrix_table(data_path)
@@ -31,7 +31,6 @@ def prepare_background_mt(data_path: str, qc_variants_ht: hl.Table, tmp_path) ->
         background_mt = background_mt.semi_join_rows(qc_variants_ht)
         return background_mt.densify().checkpoint(background_mt_checkpoint_path, overwrite=True)
     elif data_path.endswith('.vds'):
-        background_mt_checkpoint_path = join(tmp_path, 'densified_background.mt')
         background_vds = hl.vds.read_vds(data_path)
         background_vds = hl.vds.split_multi(background_vds, filter_changed_loci=True)
         background_vds = hl.vds.filter_variants(background_vds, qc_variants_ht)
@@ -93,7 +92,7 @@ def add_background(dense_mt: hl.MatrixTable, sample_qc_ht: hl.Table, tmp_path: s
         logging.info(f'Adding background dataset {dataset_dict["dataset_path"]}')
 
         # prepare the background MT, from either a Mt or VDS
-        background_mt = prepare_background_mt(dataset_dict['dataset_path'], qc_variants_ht, tmp_path=tmp_path)
+        background_mt = prepare_background_mt(dataset_dict['dataset_path'], dataset, qc_variants_ht, tmp_path=tmp_path)
         metadata_table = prepare_metadata_table(dataset_dict['metadata_table'], sample_qc_ht)
 
         background_mt = background_mt.annotate_cols(**metadata_table[background_mt.col_key])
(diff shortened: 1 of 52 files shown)


cpg_workflows/large_cohort/scripts/ancestry_add_background.py
```diff
@@ -23,15 +23,14 @@ def prepare_background_mt(data_path: str, qc_variants_ht: hl.Table, tmp_path) ->
     Returns:
         A densified MT, checkpointed
     """
-    background_mt_checkpoint_path = join(tmp_path, 'densified_background.mt')
+    background_mt_checkpoint_path = join(tmp_path, f'{dataset}_densified_background.mt')
```
MattWellie, line 26: I'm actually kinda keen to cut this prep stage out into a separate Stage - it looks like we read these MT/VDSs in, do some stuff on them, and write them into temp... then next time the pipeline runs we do it all over again. Why not make a stage where we prepare each background MT as a densified MT, save it to long term storage, then we don't need to do this each time.

Are these the 1+TB tables?

michael-harper: Yep, the background vds for hgdp-1kg is around 1.6Tib (`gs://cpg-hgdp-1kg-test/vds-main.vds`). We ran into issues with this a couple of months ago ([Slack here](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1712720352383989)).

I added the saving to `tmp` not that long ago because it saved time during debugging. I guess that these runs on main are somewhat infrequent and so storing (what could be a fairly large file) may be reason to avoid more permanent storage. That's just me being naive to how much it costs to store things and how big the files for may be. Suffice to say, I think it would be good to have a more permanent storage for them, if anything to inspect for accuracy and quality control

michael-harper: To be fair, I don't know how many different types of background datasets we want to add/will have access to. This may very well be the only one that suits our needs. Which would mean it's okay to/we probably should store permanent intermediates of this dataset

MattWellie: Cool cool. I wasn't sure how many background tables exist or how large they are. If it's one fat table that we reprocess with each (infrequent) run that's something we can plug into the ol' calculator and make a decision about. I haven't seen this stage run on main data, so not sure how long this is expected to take, or how much it cost in dataproc

michael-harper: So I looked into possibly splitting this out as it's separate stage. Couple of things:
- The dataproc removal process is already adding stages and I'm not sure how this sits within the PA team, so at this point if we can prevent adding another stage that would be preferential
- We currently have a `DenseSubset` stage which essentially does what we want it to but for the dataset in question (i.e. `bioheart`, `tob-wgs`, `tenk10k` and not the background datasets). If we wanted to do it for the background dataset then it would require a completely separate pipeline run to run just that stage which kind of breaks the autonomy.
- I think we can implement a check with within the `prepare_background()` function (with possibly a another config parameter 🙄) for the existence of the background datasets densified matrix table. This would mean that it would all happen within the current pipeline run

michael-harper:
```suggestion
    bg_path = config_retrieve(['storage', bg_dataset, 'default'])
    background_mt_checkpoint_path = join(bg_path, 'dense_background.mt')
    if can_reuse(background_mt_checkpoint_path):
        return hl.read_matrix_table(background_mt_checkpoint_path)
```

cpg_workflows/large_cohort/scripts/ancestry_add_background.py
```diff
@@ -122,7 +122,9 @@ def add_background(dense_mt: hl.MatrixTable, sample_qc_ht: hl.Table, tmp_path: s
         # save metadata info before merging dense and background datasets
         sample_qc_ht = sample_qc_ht.union(background_mt.cols(), unify=allow_missing_columns)
 
-        background_mt = background_mt.select_entries('GT', 'GQ', 'DP', 'AD').naive_coalesce(5000)
+        background_mt = (
+            background_mt.select_cols().select_rows().select_entries('GT', 'GQ', 'DP', 'AD').naive_coalesce(5000)
+        )
```
MattWellie, line 127: As far as I can work out, what this is doing is:
- select_rows with no argument drops any row fields which aren't part of a key - here that would... drop everything except the locus/alleles?
- select_cols with no argument drops any col fields which aren't part of a key, so only keeps sample ID
- selects a portion of the entry fields
- union_cols reduces dense_mt to the sample IDs in the background MT

Maybe? That would mean that the select_rows() call is probably useless, and if that is what's going on here this method should probably be documented better 😅

cpg_workflows/large_cohort/scripts/ancestry_add_background.py
MattWellie: > The dataproc removal process is already adding stages and I'm not sure how this sits within the PA team, so at this point if we can prevent adding another stage that would be preferential

From my POV you're thinking about stage splitting all wrong - the current process is super complicated, takes tons of inputs, writes tons of outputs, and not in a way that's easily disentangled. All the individual parts of the process run independently, but all one after another in a single Stage, so changing any part of the process means changing the whole Stage. 

The most important part here is that the script as written includes a number of `resume-from-checkpoint` statements, so a new run of the 'Relatedness' or 'Ancestry' Stage doesn't mean you're running the whole process, it means the method can resume from a previous failed run. That failure might have been from a previous run or a previous commit you don't want to continue from. I heavily advocate against any statements which allow `this` run to continue from `that` run under any circumstances.

Splitting this one trash stage into several stages means that each step of the process is discrete - only take the relevant inputs for `this` part of the process, and do the minimum amount of work to achieve that. Once it's implemented we have multiple simple steps instead of one monolithic step, but the pipeline runs automatically, so that doesn't cause any more effort to operate, and it's much easier to debug.

MattWellie: > If we wanted to do it for the background dataset then it would require a completely separate pipeline run to run just that stage which kind of breaks the autonomy.

I don't agree here. The pipeline stages only run if their outputs don't exist. If you create a stage called `MakeDenseHGDP` (for example) which writes a densified version of the HGDP MT into permanent storage (maybe CPG common?), you can add that to the start of the pipeline and then use the outputs in other stages. Subsequent runs of the pipeline would check that it exists, and not run that stage, but would be able to re-use the output. That wouldn't have to be a separate pipeline/workflow in order to work.

I'm not convinced that it needs to be a Stage of its own, but if it were to be broken out like that I would create it as a Stage within the large_cohort pipeline, and only expect it to ever run once (unless the logic of how to generate the BG datasets changes)

cpg_workflows/stages/large_cohort.py
```diff
@@ -255,16 +272,242 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
...
+            for bg_dataset in background_datasets
+        }
+
+    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
+        # no datasets, no running
+        if not config_retrieve(['large_cohort', 'pca_background', 'datasets'], False):
+            return self.make_outputs(target=cohort)
+
+        # inputs
+        background_datasets: list[str] = config_retrieve(['large_cohort', 'pca_background', 'datasets'], False)
```
MattWellie, line 297:
```suggestion
        if not (background_datasets := config_retrieve(['large_cohort', 'pca_background', 'datasets'], False)):
            return self.make_outputs(target=cohort)
```

cpg_workflows/stages/large_cohort.py
```diff
@@ -255,16 +272,242 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
...
+            Path(config_retrieve(['storage', dataset, 'default']) for dataset in background_datasets),
+        ]
+
+        # construct input paths
+        # TODO: Allow for input background datasets to be '.mt' files
+        background_vds_version = dense_background_vds_version()
+        input_paths = {
+            bg_dataset: background_storage_paths[i] / 'vds' / f'{background_vds_version}.vds'
+            for i, bg_dataset in enumerate(background_datasets)
+        }
```
MattWellie, line 308:
```suggestion
        # construct input paths
        # TODO: Allow for input background datasets to be '.mt' files
        background_vds_version = dense_background_vds_version()
        input_paths = {
            dataset: to_path(config_retrieve(['storage', dataset, 'default'])) / 'vds' / f'{background_vds_version}.vds'
            for dataset in background_datasets
        }
```

cpg_workflows/stages/large_cohort.py
```diff
@@ -255,16 +272,242 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
...
+        # localise the Dense MT
+        for bacgkround_vds in input_paths.values():
+            job.command(f'gcloud --no-user-output-enabled storage cp -r {bacgkround_vds} $BATCH_TMPDIR')
+
+        # expected outputs
+        outputs = self.expected_outputs(cohort)
+
+        # localise QC variants table
+        qc_variants_ht = config_retrieve(['references', 'ancestry', 'sites_table'])
+        job.command(f'gcloud --no-user-output-enabled storage cp -r {qc_variants_ht} $BATCH_TMPDIR')
```
MattWellie, line 322: AFAIK you should be fine without localising the data for this stage - this is just reformatting/annotating the contents of a MT/VDS, and Query-on-batch should be able to handle that OK. 

The stages where we need Dataproc are where there's a feature that's not been ported to QOB yes, like PCA. To pull those out of Dataproc we need to run locally, but for this work we should be fine running in QOB

MattWellie (commented):
Couple of comments. If we go with this Background preparation stage approach this looks like a pretty good start

cpg_workflows/stages/large_cohort.py
```diff
@@ -279,9 +279,13 @@ class DenseBackground(CohortStage):
...
 
     def expected_outputs(self, cohort: Cohort) -> dict[str, Path]:
+        access_level = config_retrieve(['workflow', 'access_level'])
         background_datasets: list[str] = config_retrieve(['large_cohort', 'pca_background', 'datasets'], False)
         background_storage_paths: dict[str, Path] = {
-            dataset: Path(config_retrieve(['storage', dataset, 'default'])) for dataset in background_datasets
+            dataset: to_path(
+                config_retrieve(['storage', ('-').join(dataset.split(f'-{access_level}')[:-1]), 'default']),
+            )
+            for dataset in background_datasets
```
MattWellie, line 288: I'm not sure I understand this... Well, I do, but I think there's a better solution. 

In the `large_cohort.pca_background.datasets` part of the config you've provided `['thousand-genomes-test']`, and later on there's `large_cohort.pca_background.thousand-genomes-test`. 

You can just coordinate this in config - if you list `['thousand-genomes']` as the dataset, and `large_cohort.pca_background.thousand-genomes` as the config section there's no need for this splitting. You'd have a separate config file for main and test modes, each pointing to the relevant input tables.

MattWellie:
Closing as...... we're going in a different direction here, and it's not my job 😇

PR #879, opened 2024-08-07
T-Shirt MT sizing returns an int

MattWellie:
Instead of a String, MT sizing method returns an int, which is more easily modified (i.e. doubling)

This has implications for the relatedness/ancestry feature branches which already use the method with a String result


michael-harper (approved):
LGTM

PR #880, opened 2024-08-07
Check for overlapping cohorts

### LINKED ISSUE #866: GATK-SV requires mutually exclusive cohorts
Need to add a step to the GATK-SV multisample pipeline which checks that there are no sample overlaps in the various Cohorts in the pipeline.

Pretty late on (in the MakeCohortVcf stages) the pipeline will fail if it tries to merge VCFs which have an overlapping sample set. 

MergeBatchSites is the first MultiCohortStage, so check there, raise an informative message, and quit.

MattWellie:
Closes #866

I borked up when creating the GATK-SV cohorts for Bioheart/TOB, and the process ran until nearly the end before failing. Due to a copy-paste error I created 2 custom cohorts with the same group of SG IDs. This failed when merging a VCF, as bcftools merge (or a GATK equivalent?) requires non-overlapping sample sets to create merged output VCFs.

When assembling a MultiCohort from separate cohorts, there's no check that the SG IDs in each don't overlap. The pipeline API shouldn't need that, as it might be useful in some workflows. It is fatal for GATK though, and the stage which fails is ~2 days after initially triggering the workflow, so an up-front check is needed.

This change inserts a new method at the start of the GATK-SV pipeline, checking that all the component cohorts have no overlaps between. If there are overlaps, the process will fail. If there are not, we continue to writing a MultiCohortPedigree - this file is written once per MultiCohort ID, so this check is only run once per MultiCohort

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.26.4
+current_version = 1.26.5
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.26.4
+  VERSION: 1.26.5
 
 jobs:
   docker:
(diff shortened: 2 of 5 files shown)


vivbak:
Approving so you're unblocked, but will write a post to see if we need to handle this for any other cases at a higher level :) For example when the multicohort object is generated before we get to the stage definitions.

vivbak (approved):

PR #881, opened 2024-08-08
Add CN calculation to sniffles modifier

MattWellie:
To get data in the right format for Seqr we need a 'CN' attribute in the VCFs, which Sniffles has not provided. This change updates the Sniffles VCF modifier script: 

- take the sex (int) of the individual
- for each variant use the sex and contig name to determine a baseline copy number
- count number of alt alleles at this variant, and adjust the CN up or down according to the variant type

Based on Seqr's display, CN only matters for DEL and DUP. All other types just take a placeholder value.

Slight change to the Stage

- no need to pass both Fasta and index separately
- pass the sex value for the SequencingGroup from the pedigree as an integer
- I changed some variable names to reduce the line count, but... that didn't work

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.26.5
+current_version = 1.26.6
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


cassimons (approved):
Looks good thanks!

PR #882, opened 2024-08-08
Adaptation to seqr-loader with repartitioning

### LINKED ISSUE #839: Hail MT Partitions
Operations in Talos (and soon in MtToEs) localise MatrixTables, then operate on the result using a local Hail PySpark runtime. This is much cheaper than QOB, with comparable or better runtimes, but I've found a few issues around high partition counts. When trying to operate on localised data in a VM with finite resources, some Hail Query jobs fail deterministically with Java Heap Space errors.

These issues aren't linked to overall data size, but do seem to be connected to partition counts. I was processing 2 sets of genome data using a local spark Hail instance, both similar sizes on disk (~400GB), but one consistently failed, one always worked. The partition count of the successful MT was ~7500. The Failing one was ~120000. This count results in an estimate of only ~250 variant rows per partition. When I repartitioned a previously failing dataset MT down to 80 (10 * available CPUs) partitions, all my issues went away (and runtime went from taking 50+ mins to fail, to sub-20 mins to complete).

Our exome MTs clock in at 3-6k partitions and never ran into similar heap space errors, though the number of variants per partition is similar given the smaller datasets.

I think I've tracked down the source of the problem - https://github.com/populationgenomics/production-pipelines/pull/514

This PR was the adoption of Hail Query-on-Batch for parts of the Seqr-loader pipeline which were previously trapped on DataProc. During the development of that PR (based on the commit messages) I was decreasing the `block_size` (MB/partition) when importing the VCF into a MatrixTable, with the final commit landing on 10MB. I don't remember whether the block size was being reduced because of runtime, or because the job failed, or something else... The default value for block-size is 128MB ([ref](https://hail.is/docs/0.2/methods/impex.html#hail.methods.import_vcf)), so we're shooting way lower than that. My guess is that the default block size works well for local operations, but QOB makes use of more distributed compute, so the balance shifts in favour of sending more, smaller data packets. This may not hold true, iterating on these block sizes was not well documented at the time, and QOB has matured a lot since that time.

We now have a really low block size during AnnotateCohort (multiple datasets) which creates hyper partitioned data. We then run AnnotateDataset on this data which breaks out individual Dataset subsets from the larger MT, but without repartitioning. If we have 3000 samples in the full callset, but only 30 samples in a subset, the number of partitions for the per-Dataset MT is inappropriately large.

This raises a few issues for investigation:

1. Why is our block size set so low, and can it be revised up
2. Should we repartition during the Dataset subsetting process - currently the only consumers of AnnotateDataset results are Talos and MtToEs, both of which are moving to a finite-resource VM model instead of DataProc/QOB

When this was under development I was running it in test using the Validation dataset - this was run as a single-Dataset cohort (i.e. there would be no subsetting from AnnotateCohort to AnnotateDataset), and the overall number of variants would have been low. This just wasn't a great stress test for appropriately provisioning the ~TB MTs we now split between 10s of different Datasets

MattWellie:
Closes #839 

This tries to address the super high number of partitions in our MT data. The issue is twofold:

1. when we read the (Multi)Cohort VCF into a MatrixTable, we use a block size of 10MB, down from a default of 128MB, which is used in calculating partitions. I genuinely don't remember the logic of that.
2. when we shard the single-dataset MTs off from the MultiCohort MT we don't repartition. If a 40-sample dataset is sharded off from the 4000-sample MT, the rough data content of each partition is ~100kb, which might be just a handful variants. In reality many of those ~120k partitions could be just a single variant, or none.

There is computational overhead in processing each partition, so we should try raising the data content per-partition, and maintain rough parity in data density between our MultiCohort and Dataset MTs. It'll either break, or speed up processing by reducing QOB overhead 🤷‍♂️ 

When I moved Talos from running in QOB to running in a local-spark session, as soon as the compute resources became finite, the super high partition count was fatal - particular in smaller MTs with high partitioning where the data was quite sparse. This suggests that although QOB can process a bazillion partitions through infinite scaling, it's probably not an optimal way to store/process the data.

n.b. we don't _need_ this for our production workflows to work, and I've mitigated this in Talos through repartitioning the localised data. It's possible that it will either bring our cost/runtimes down, or break everything 🙃. I don't know if we have enough data in test to really stress test this change.

- Adds type hints to the seqr_loader methods. As a result this removes the many instances of `str(string)`.


cpg_workflows/query_modules/seqr_loader.py
```diff
@@ -169,58 +171,71 @@ def annotate_cohort(
...
+    # work out the proportion of the original sample count that we'll keep
+    downsample_ratio = len(mt_sample_ids) // len(unique_sids)
+    # calculate the proportional number of partitions to generate
+    downsample_partitions = int(mt.n_partitions() * downsample_ratio)
+
+    logging.info(f'Found {len(mt_sample_ids)} samples in mt, subsetting to {len(unique_sids)} samples.')
+    logging.info(f'Re-reading the MT with {downsample_partitions} partitions, reduced from {mt.n_partitions()}')
+    # re-read instead of a repartition/naive_coalesce
+    # naive coalesce could produce really unbalanced partitions, and repartitioning might be more expensive
+    # "it's basically free to read with different partitioning than you wrote with"
```
MattWellie, line 221: I don't believe this - in Talos both running on a high-partition dataset and reading that partitioned dataset with a different number of partitions cause runtime failures. Reading a dataset, writing it back out with a different number of partitions, then reading that in again worked OK..

I expect QOB can handle it, but local-pyspark'y Hail cannot

cassimons (approved):

ckrishk (approved):

MattWellie:
Test run [here](https://batch.hail.populationgenomics.org.au/batches/475076) - this takes the full MT which was copied into test, and creates a subset containing 30 of those samples. It does a proportional downsampling of the dataset based on the number of selected samples relative to the full cohort size.

`Re-reading the MT with 892 partitions, reduced from 115973` 

This hasn't tested the small secondary change: reading in the VCF using a different block size (default of 128MB instead of 10MB), please advise if you'd like that dropped from this PR. The VCF is sitting in main, so we can copy it over and try it, but it is a hefty 1.1TB, so we'd need to delete it pretty sharpish after lest we incur chunky storage costs

PR #883, opened 2024-08-09
Generate new intervals from MT

MattWellie:
Generate new intervals from a naive MatrixTable interval inference, polished using a telo/centromeres region file.

This is a pretty simplistic approach:
- use hail to calculate `n` intervals in a MT
- if any of those intervals span two chromosomes, split 'em
- If this is the first interval on the contig, force the start position to 0. If it's the last, force the end position on == contig length.
- then clean each interval:
   - if it overlaps with a telomere, shrink the start/stop so that it doesn't
   - if it overlaps with a centromere, split it into two new non-overlapping intervals

I've only run this on a teeny dataset to test, but it completed in a few seconds and the results look usable. I've not checked if the resulting intervals are roughly balanced in terms of data content.

The Centro/Telomere splitting will also create some baby intervals, a secondary cleanup stage where intervals under a certain size threshold are merged with an adjacent interval could be a decent addition.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.26.9
+current_version = 1.26.10
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.26.9
+  VERSION: 1.26.10
 
 jobs:
   docker:
--- a/cpg_workflows/scripts/generate_intervals_wrapper.py
+++ b/cpg_workflows/scripts/generate_intervals_wrapper.py
@@ -3,21 +3,32 @@ Generate new intervals from a MatrixTable based on data distribution
 A wrapper for the generate_new_intervals.py script
 """
 
+from argparse import ArgumentParser
+
 from cpg_utils.config import config_retrieve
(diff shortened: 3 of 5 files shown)


EddieLF:
> it completed in a few seconds and the results look usable

Fantastic! Is this all happening in the `_calculate_new_partitions` function? 
My trouble with this was with the aggregation/collection of the Table's values into a list or something that could actually be written to a disk, so it's awesome to hear this happens quick using this method. [This line](https://github.com/hail-is/hail/blob/7d2577954afe81491f77a8cc4992f3a15d872063/hail/python/hail/matrixtable.py#L4569) looks like it might be doing the collection? Not quite sure I get what's actually happening but if it works it works. Keen to test this on a larger mt.

cassimons:
If I am following this right, the start of the first interval and the end of the last interval in each chrom are determined by the first/last variants in the MT (before polishing with telomeres). If so, this can lead to a corner case where the start/end of contigs are inappropriately omitted from the intervals.

Can we hardcode the first interval start and the last interval end of each contig is 0, contig.length() before polishing is done?

Ignore me if you already have this sorted and I messed it.

MattWellie:
> If so, this can lead to a corner case where the start/end of contigs are inappropriately omitted from the intervals.

Yup, that's correct. The intervals are worked out using a key-sampling approach, so the first/last of each contig was based on the actual variants. I've altered this to bump the first interval on each contig to start at 1, and the final one to the contig length.

cassimons (approved):
Did some QC on the output and it looks to be exactly what we want.

PR #887, opened 2024-08-16
Refactor scripts/fix_sq_headers.py to always get analysis records from metamist

jmarshall:
This PR updates the _fix_sq_headers.py_ script to handle updating records in metamist so that the newly reheadered CRAM file is attached as an Analysis record to the existing sample / active sequencing group.

Context: https://centrepopgen.slack.com/archives/C020PDNH3N0/p1724726840838769

As `path.stem` no longer suffices to guess the sequencing group identifier (if indeed it ever did!), we always query the relevant information from metamist, and filter by the paths requested if any source paths are given.

Copy the old analysis record metadata when creating new analysis records, updating size and source fields. Also mark old analysis records as inactive (depends on populationgenomics/metamist#899). (There's a mismatch between the status strings, which are returned in uppercase, and models.AnalysisStatus, which insists on lowercase, so we hack this explicitly via `lower()`.)

Allow specifying either GCS files or directories to limit processing.

### DIFF
--- a/scripts/fix_sq_headers.py
+++ b/scripts/fix_sq_headers.py
@@ -1,5 +1,18 @@
 #!/usr/bin/env python3
 
+"""
+Copy CRAM files while replacing existing @SQ headers with new ones as specified,
+add new analysis records to metamist pointing to the new CRAM files, and
+inactivate the existing analysis records for these sequencing groups.
+
+Typical usage:
+
+analysis-runner ... scripts/fix_sq_headers --ref gs://.../reference.dict
+    --expected 'Homo_sapiens_assembly38[unmasked]'
+    --intended 'Homo_sapiens_assembly38_masked'
+    [gs://... directory or file paths to limit processing]...
+"""
+
 import hashlib
 import os
 import subprocess as sb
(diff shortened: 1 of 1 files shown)


jmarshall:
Batches 479351 (submission) and 479352 (execution) are a representative run of this in test. They demonstrate CRAM files being ignored because there is no Analysis record (as test is a bit borked!) and because it has already been reheadered, and then two other CRAMs being processed and the database updated accordingly.

scripts/fix_sq_headers.py
```diff
@@ -138,23 +157,33 @@ def do_reheader(dry_run, refset, newref, newrefset, incram, outcram, outcrai):
...
+    aid = aapi.create_analysis(dataset, newanalysis)
     print(f'Created Analysis(id={aid}, output={newpath}) in {dataset}')
 
+    for analysis in activeseqgroup['analyses']:
+        activeid = analysis['id']
+        analysisupdate = models.AnalysisUpdateModel(
+            active=False,
+            status=models.AnalysisStatus(analysis['status'].lower()),
+        )
+        if not aapi.update_analysis(activeid, analysisupdate):
```
dancoates, line 183: would it be worth putting a small wait time in here to avoid hitting the metamist server a bunch of times in a loop without any let up? Or is the amount of the analyses on the seq group expected to be small?

jmarshall: To be honest, the number of analyses on the seq group is expected to be 1 but I put the loop in just to make sure!

dancoates: No worries, that was my guess but just wanted to double check, all good with me then

dancoates (approved):

jmarshall:
Thanks for the review. Apropos overwhelming the metamist server, I did think about limiting the number of jobs (for different samples) that could run simultaneously. But that's a bit painful, and I think I'd rather leave it to Hail max_instances and/or just batching them a bit during submission.

PR #888, opened 2024-08-16
Dragmap parity investigation

michael-harper:
We are wanting to investigate the parity between different versions of dragmap used between NAGIM and our own pipeline. 

This involves taking a nagim vds (or equivalent matrix table) and a vds produced by our pipeline, converting them to a bi-allelic matrix table and running `hl.concordance()` on them. This function produces three outputs discussed in Hail docs here.

In building this script I produced a nagim vds from nagim-produced gvcf files and a `new` vds produced from gvcf files produced from our pipeline. 
See output of this script [here](https://batch.hail.populationgenomics.org.au/batches/475240/jobs/1) that produced output in this directory: gs://cpg-tob-wgs-test/dragmap_parity/v2/

**Note**: the nagim file in main that will be used is not a `vds` but a `MatrixTable`, produced quite a while ago, and it being on main I am unable to test whether this script works with that file.


silkm (approved):
Gone over it, looks awesome. Assuming that `annotate_cols()` call works on the test data, looks perfect

michael-harper:
Moving this PR over to [tob-wgs repo](https://github.com/populationgenomics/tob-wgs/pull/178)

PR #899, opened 2024-09-11
Fix cohort.get_sequencing_groups in cpg_workflows targets.py

EddieLF:


MattWellie:
I think this is also close-able WRT #959, which guts all crossover between Cohorts and Datasets

Issue #901, opened 2024-09-12
MultiCohorts Overpopulating Datasets 


vivbak:
https://github.com/populationgenomics/production-pipelines/pull/860 Introduced a bug where a dataset is populated with all the sgs regardless of cohort restrictions. 

This means that the first cohort, when given a list of cohorts, will be populated by all sgs in the datasets that are members of it. 

This created an issue here https://centrepopgen.slack.com/archives/C018KFBCR1C/p1726017285437029

PR #903, opened 2024-09-13
Fix MultiCohort Bug when populating Datasets. 

vivbak:
Closes https://github.com/populationgenomics/production-pipelines/issues/901 

The initial PR to fix the dataset issue as described here, https://docs.google.com/document/d/1Zu4QFxuR44M1alRZtA5M9WGROLEUg74gQWorDQjOvTs/edit?usp=sharing 
introduced a bug when created datasets on a multicohort, as reported here https://centrepopgen.slack.com/archives/C018KFBCR1C/p1726017285437029 

This PR uses deepcopy, because the pointers were creating dramas!

### DIFF
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -2,7 +2,6 @@
 Metamist wrapper to get input sequencing groups.
 """
 
-import copy
 import logging
 
 from cpg_utils.config import config_retrieve, update_dict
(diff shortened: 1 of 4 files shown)


cpg_workflows/inputs.py
```diff
@@ -66,18 +67,18 @@ def create_multicohort() -> MultiCohort:
...
-    #             mc_dataset.add_sequencing_group_object(sg)
+    # now populate the datasets uniquely in the multicohort, each containing all sequencing groups in this dataset
+    # the MultiCohort has a dictionary of {dataset_name: Dataset} objects
+    # each of those Dataset objects links to all SequencingGroups in that Dataset, across all Cohorts
+    for cohort in multicohort.get_cohorts():
+        for dataset in cohort.get_datasets():
+            # is this dataset already in the multicohort, if so use it, otherwise add this one
+            # Since datasets are created on the Cohort, and are added to the multicohort, we need to
+            # perform a deepcopy to avoid manipulating the cohort dataset at this point
+            mc_dataset = multicohort.add_dataset(copy.deepcopy(dataset))
```
milo-hyben, line 78: After our session, I am thinking this deep copy should be here, it would speed up the function as it only created copy when needed.

https://github.com/populationgenomics/production-pipelines/blob/20da8c3b8939b434eeeed404c80bdb7d8d79fdf7/cpg_workflows/targets.py#L191

vivbak:
Thanks @jmarshall, we've moved the call to deepcopy and added a better test!

violetbrina (approved):

PR #912, opened 2024-09-23
Annotate cohort and create es-index for LRS SNPs Indels

EddieLF:
Pipeline workflow for Long Read Individual VCFs -> ES-Index

**New stage** - `seqr_loader_long_read/long_read_snps_indels_annotation`

**New jobs** - `seqr_loader_snps_indels`

Mostly adapts the SV code from `seqr_loader_long_read/long_read_sv_annotation`

---

## Code changes

The intention for all these code changes is to ensure that other pipelines which share the code are not impacted. 

#### `main.py`
- Add a new pipeline, `'long_read_snps_indels_annotation'`, with the last stage being `MtToEsLrSNPsIndels`

#### `cpg_workflows/query_modules/seqr_loader.py` 
- Add a new boolean argument to the `annotate_cohort` query module: `long_read`. 
If true, then drop the `'AF'` field from the MatrixTable and recompute it with `hl.variant_qc(mt)`. Default is `False`, to prevent the standard short-read pipeline from being impacted by this change. 

#### `cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py` 
- Add a new boolean argument to the script: `sv`. 
If True, then do the formatting required to process the SV VCFs before writing the lines.
If False (for SNPs_Indels VCFs), then extra processing is not needed and the lines can just be written as-is.

#### `cpg_workflows/utils.py`
- Move `get_intervals_from_bed` from `cpg_workflows.joint_genotyping` to `cpg_workflows.utils`

#### `cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py`

| Stage | Jobs / Notes |
|--------|--------|
| `ReFormatPacBioSNPsIndels` | Uses the existing `modify_sniffles` job with argument `sv = False` |
| `MergeLongReadSNPsIndels` | Merges the input SNPs_Indels VCFs with `bcftools merge` |
| `SiteOnlyVCFs` | Splits the VCF into parts based on the input intervals, creates site-only VCFs as well, all for VEP. |
| `VepLRS` | Uses the existing `add_vep_jobs` from `cpg_workflows.jobs.vep` |
| `AnnotateCohortLRSNPsIndels` | Uses the existing `query_modules.seqr_loader.annotate_cohort` function to create the annotated cohort mt, with minor modifications to ensure the `'AC'`, `'AF'`, and `'AN'` fields are populated correctly for seqr. | 
| `AnnotateDatasetLRSNPsIndels` | Uses the existing annotate dataset job from `cpg_workflows.jobs.seqr_loader`.   |
| `MtToEsLrSNPsIndels` | Uses the existing mt to ES without dataproc script to export the mt to Elastic. |

---

General process for this pipeline:

1. Find the `"SNPs_Indels.phased.vcf.gz"` files associated with the LRS sequencing groups by querying Metamist for `"pacbio_vcf"` analyses.
2. Reformat each of the PacBio VCFs using `modify_sniffles`, block-gzipping and indexing the output VCF with bgzip and tabix.
3. Merge all of the sniffles VCFs into a cohort VCF using `bcftools merge`.
4. Split the merged VCF into `scatter_count` VCFs using GATK `SelectVariants` and a list of `scatter_count` intervals.
5. Create `scatter_count` sites only VCFs using GATK `MakeSitesOnlyVcf` on the VCF parts from 4.
6. Run a VEP job over each sites only VCF created in 5. and gather each of the VEP outputs into a Hail Table ht.
7. Create a cohort MatrixTable from the merged VCF created in 3. and the full VEP ht. created in 6.
8. Create a dataset MatrixTable from the cohort mt created in 7. by subsetting to the dataset's sequencing groups.
9. Write an elasticsearch index to the cluster based on the dataset mt. created in 8.

Note - Steps 4. and 5. are necessary for parallelising the VEP jobs. Even though we merged all the input VCFs in step 3, we still have to split the merged VCF by intervals so that the VEP jobs can do their thing. 

**Not sure if this approach is optimal.** 

~To split the VCF and create sitesonly VCFs over `scatter_count` intervals, we need to pass the merged VCF around to `scatter_count` jobs. This means after step 3., we could have a large VCF that is downloaded into `scatter_count` VMs to be split by the GATK `SelectVariants` jobs. Might not be an issue if the merged VCF isn't too big. Could potentially cause scaling issues in the future though.~

As suggested by Matt
> We can read the merged VCF in once, and queue scatter_count commands within that VM, to generate scatter_count subset outputs (+ corresponding indices). That way we provision one VM, it requires roughly the same resources as a single scatter job, but runs all the separate subset commands, generating all outputs. We'd be trading runtime for cost, but would probably be far more efficient overall


MattWellie:
> To split the VCF and create sitesonly VCFs over scatter_count intervals, we need to pass the merged VCF around to scatter_count jobs. This means after step 3., we could have a large VCF that is downloaded into scatter_count VMs to be split by the GATK SelectVariants jobs. Might not be an issue if the merged VCF isn't too big. Could potentially cause scaling issues in the future though.

I think we should solve both this and the main-pipeline operation in a different way... We can read the merged VCF in once, and queue `scatter_count` commands within that VM, to generate `scatter_count` outputs (+ corresponding indices). That way we provision one VM, it requires roughly the same resources as a single scatter job, but runs all the separate subset commands, generating all outputs. We'd be trading runtime for cost, but would probably be far more efficient overall

cassimons (commented):
LGTM from a quick skim and the results in staging work well as far as I can see. Approved pending @MattWellie reading the actual code 😅  and giving his thumbs up.

cpg_workflows/jobs/seqr_loader_snps_indels.py
```diff
@@ -0,0 +1,225 @@
...
+            if split_vcf_j:
+                siteonly_j.depends_on(split_vcf_j)
+
+    jobs = [j for j in jobs if j is not None]
+    for j in jobs:
+        j.name = f'Long Read SNPs Indels Annotation: {j.name}'
+    return jobs
+
+
+def _add_split_vcf_job(
```
MattWellie, line 138: Shouldn't really be a `_` prefix here, this isn't a private class method

cpg_workflows/jobs/seqr_loader_snps_indels.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    j = b.new_job('SplitVcf', (job_attrs or {}) | {'tool': 'gatk SelectVariants'})
+    j.image(image_path('gatk'))
+    res = STANDARD.set_resources(j, ncpu=2)
+
+    for idx, interval in enumerate(intervals):
+        output_vcf_path = output_vcf_paths[idx]
+        j.declare_resource_group(
+            **{
+                str(idx + 1): {
```
MattWellie, line 157: why are the outputs/resource groups `idx + 1`? It looks like this would work just as well with plain ol' `idx`

cpg_workflows/jobs/seqr_loader_snps_indels.py
```diff
@@ -0,0 +1,225 @@
...
+        j.command(command(cmd, monitor_space=False, setup_gcp=True, define_retry_function=True))
+        b.write_output(j[str(idx + 1)], str(output_vcf_path).replace('.vcf.gz', ''))
+
+    # Wait for all parts to be written before returning
+    j.command('wait && echo "All parts written"')
+
+    return j
+
+
+def _add_make_sitesonly_job(
```
MattWellie, line 179: same here, _ prefix doesn't feel necessary

cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
```diff
@@ -0,0 +1,428 @@
...
+    MultiCohortStage,
+    SequencingGroupStage,
+    StageInput,
+    StageOutput,
+    get_workflow,
+    stage,
+)
+from metamist.graphql import gql, query
+
+from ...resources import joint_calling_scatter_count
```
MattWellie, line 32:
```suggestion
from cpg_workflows.resources import joint_calling_scatter_count
```

feels like this should be pulled from the installed module instead of the relative path, which would place it above wit the rest of the cpg_workflows import statements

cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
```diff
@@ -0,0 +1,428 @@
...
+    def expected_outputs(self, multicohort: MultiCohort) -> dict[str, Path]:
+        return {
+            'vcf': self.prefix / 'merged_reformatted_snps_indels.vcf.bgz',
+            'index': self.prefix / 'merged_reformatted_snps_indels.vcf.bgz.tbi',
+        }
+
+    def queue_jobs(self, multicohort: MultiCohort, inputs: StageInput) -> StageOutput | None:
+        """
+        reuse the gVCF method which does the same? FastCombineVcfs. Nah, not yet - that contains other modifications
+        """
```
MattWellie, line 162:
```suggestion
```

cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
```diff
@@ -0,0 +1,428 @@
...
+            tmp_bucket=self.tmp_prefix / 'tmp',
+            # out_siteonly_vcf_path=outputs['siteonly'],
+            out_siteonly_vcf_part_paths=out_siteonly_vcf_part_paths,
+            intervals_path=intervals_path,
+            exclude_intervals_path=exclude_intervals_path,
+            job_attrs=self.get_job_attrs(),
+        )
+        jobs.extend(vcf_jobs)
+        for job in jobs:
+            assert job
```
MattWellie, line 248: I'm not sure what this assert is doing - the method above already grooms out `None` jobs from the list of jobs, so it returns an empty list or a list of Jobs. Probably best to drop it, or add a statement to make it informative if it fails

cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
```diff
@@ -0,0 +1,428 @@
...
+        Subsets the whole MT to this cohort only
+        Then brings a range of genotype data into row annotations
+
+        Args:
+            dataset (Dataset): SGIDs specific to this dataset/project
+            inputs ():
+        """
+        assert dataset.cohort
+        assert dataset.cohort.multicohort
+        mt_path = inputs.as_path(target=dataset.cohort.multicohort, stage=AnnotateCohortLRSNPsIndels, key='mt')
```
MattWellie, line 351:
```suggestion
        mt_path = inputs.as_path(target=get_multicohort(), stage=AnnotateCohortLRSNPsIndels, key='mt')
```

possibly coupled with an import of `cpg_workflows.workflow.get_multicohort`. This was my jank before I realised there was a cleaner way to get to the MC object

MattWellie (approved):
A couple of stylistic things, but looks good. Nice job reusing the other LR stuff, there's more short-read variant calling we could re-use with interval/sites-only generation, but probably not worth reusing it prior to restructuring

EddieLF:
Thanks for the thorough review @MattWellie! I've implemented your suggestions and added a bumpversion commit. Will merge this into main, pending the checks.

PR #916, opened 2024-09-29
Config option to run stages for specific datasets in cohort

EddieLF:
Adds in two new workflow config options (re-adding the MtToEs option since it was removed at some point?) for the seqr_loader pipeline.

`workflow.write_mt_for_datasets` - populate this config entry with the list of datasets for which you want the `AnnotateDataset` stage to run for. Useful because each time a new joint callset is created, only certain datasets may have new data for seqr, so we don't always want to run this stage for every dataset in the callset.

`workflow.create_es_index_for_datasets` - same as above but for the `MtToEs` stage

---

Also updates the `cpg_workflows.metamist` module with a new method: `get_metamist_proj`. This method has been added to resolve an issue faced while testing. As the metamist entries for datasets and sequencing groups within a cohort were being fetched, the `"-test"` suffix was being added to the `test` dataset names, e.g. https://batch.hail.populationgenomics.org.au/batches/489269/jobs/1.

This change will ensure that the test datasets don't get a duplicate suffix, as well as reduce some repetition in the code.

### DIFF
--- a/cpg_workflows/stages/seqr_loader.py
+++ b/cpg_workflows/stages/seqr_loader.py
@@ -91,7 +91,7 @@ class AnnotateDataset(DatasetStage):
         Annotate MT with genotype and consequence data for Seqr
         """
         # only create dataset MTs for datasets specified in the config
-        eligible_datasets = config_retrieve(['workflow', 'write_mt_for_datasets'])
+        eligible_datasets = config_retrieve(['workflow', 'write_mt_for_datasets'], default=[])
         if dataset.name not in eligible_datasets:
             return None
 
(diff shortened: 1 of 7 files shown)


cpg_workflows/stages/seqr_loader.py
```diff
@@ -90,6 +90,11 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
         """
         Annotate MT with genotype and consequence data for Seqr
         """
+        # only create dataset MTs for datasets specified in the config
+        eligible_datasets = config_retrieve(['workflow', 'write_mt_for_datasets'], default=[])
+        if dataset.name not in eligible_datasets:
```
MattWellie, line 95: I can foresee a situation where someone (me) expects this to run and spends at least one whole day failing to realise this is a skip condition. Could you please add a logging statement in to say `Dataset X is being skipped as it doesn't appear in write_mt_for_datasets`

MattWellie (approved):
LGTMatt

PR #919, opened 2024-10-02
Upgrade Talos Job resources

MattWellie:
I was being incredibly stingy with resources in these two jobs. That was fine when I was doing a minimal bit of parsing, and now I'm pulling more objects and cross-comparing this is insufficient. 

This PR adds some more juice to some API-polling jobs. They're shortlived, minimal cost implications

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.27.5
+current_version = 1.27.6
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


MattWellie:
I mean... I've run this same query locally and I've no idea how these resources are blowing out like this. This should be a ~18MB data payload, and it's blowing up a job that previously consumed ~125MB of RAM to crashing a 250MB VM

silkm (approved):

PR #921, opened 2024-10-02
Filterwham outputs fix

EddieLF:
Minor fix to the FilterWham stage added in #920 

See failure mode:
https://batch.hail.populationgenomics.org.au/batches/492753/jobs/1

Probs didn't need a bumpversion but added it out of habit


MattWellie (approved):
Dang

PR #924, opened 2024-10-04
Talos timecap

MattWellie:
Stop those expensive runs burning a hole in our budgets


EddieLF (approved):
These could be added to configs/defaults/talos.toml for consistency

PR #925, opened 2024-10-08
Remove sgids from job attrs

### LINKED ISSUE #914: Hard-Cap in Job Attribute value sizes
For large cohorts, we've hit a size limit in the Hail Job attributes

https://centrepopgen.slack.com/archives/C030X7WGFCL/p1727226994302319

We're hitting this because we're trying to submit jobs with a `sequencing_groups` field containing all CPG IDs in the analysis (e.g. here https://github.com/populationgenomics/production-pipelines/blob/ed59e02b541d700416e323de9bbdb0f7f21be090/cpg_workflows/targets.py#L209)

Maybe we should stop collecting all SG IDs into job attributes, other than at the SequencingGroupStage level?

MattWellie:
Closes #914

See https://centrepopgen.slack.com/archives/C030X7WGFCL/p1727226994302319

Hail jobs have a hard cap on data payload size. We are now exceeding this with SequencingGroups, as the `get_job_attrs` methods for MultiCohorts, Cohorts, Datasets, and SequencingGroups all push the full list of SG IDs into the job attributes.

I'm not aware of a situation where this data is used, and it is now a blocker in the pipeline processing clinical data.

This is not used to register files/analysis outputs, that list of SG IDs is generated by the same underlying method (`target.get_sequencing_group_ids()`), but is used directly instead of posting as job attributes.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.27.10
+current_version = 1.27.11
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.27.10
+  VERSION: 1.27.11
 
 jobs:
   docker:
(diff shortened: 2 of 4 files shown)


EddieLF (approved):
LGTM as a temp fix for very large cohorts

vivbak (approved):
Since you're leaving the equivalent attribute in the SG Stage, and the datasets/cohort IDs can be used to backtrace the sg ids for debugging, I think this is a good work-around.

PR #926, opened 2024-10-08
Strip extra .vcf.bgz extension from FilterWham stage outputs

EddieLF:
Removes the extra `.vcf.bgz` from the `FilterWham` output filtered vcf.

Outputs were created with extra extensions like this:
```
gs://cpg-dataset-main/gatk_sv/MultiCohortHash/FilterWham/filtered.vcf.gz.vcf.bgz
gs://cpg-dataset-main/gatk_sv/MultiCohortHash/FilterWham/filtered.vcf.gz.vcf.bgz.tbi
```

And as such were not picked up by the next stages
https://batch.hail.populationgenomics.org.au/batches/493256/jobs/92


MattWellie (approved):
I really messed this one up eh?

EddieLF:
@MattWellie right to jail!

PR #929, opened 2024-10-10
Crush some noisy logging

MattWellie:
This is an unsolicited PR to try and make the pipeline startup logging more user-friendly.

This un-does an unintentional change introduced in [this earlier PR](https://github.com/populationgenomics/production-pipelines/pull/619#issuecomment-1955732269), and it was noted at the time of merging that this completely clouded the logging. Analysts don't want this, and it should be reverted.

Example batch: https://batch.hail.populationgenomics.org.au/batches/493507/jobs/1

There's a lot of chaff in the logs here, and it makes these pages really slow to load/scroll through.

Logging messages starting with "Assuming outputs exist:" for skipped stages were printed 50k times. Fifty Thousand Times. We did this intentionally - the skipping was defined in config. I've updated this to print this only if we intend to print debugging messaged, using the `--verbose` argument to `main.py`.

The logging messages also follow the template `fmt = '%(asctime)s %(levelname)s (%(name)s %(lineno)s): %(message)s'`, e.g. "2024-10-10 00:54:29 INFO (root 700):".

`name/root` here is the name of the logger, which is useless, as we never set a logger name. This would be a meaningful if instead of the logger name, we insert `pathname` into the logging template, so the log messages would show the File and Line number, instead of a line number in an unknown file.

`Using VEP 110` is also printed once for every sample - pretty silly as we now have this as a permanent default


cpg_workflows/workflow.py
```diff
@@ -697,13 +698,14 @@ def _is_reusable(self, expected_out: ExpectedResultT) -> tuple[bool, Path | None
                 Path | None: first missing path, if any
         """
         if self.assume_outputs_exist:
-            logging.info(f'Assuming outputs exist. Expected output is {expected_out}')
+            # change this to debug, as it's a bit noisy
```
jmarshall, line 701: This is a comment for a commit message, not a source code comment — It doesn't make sense without the context of the change.

main.py
```diff
@@ -87,11 +87,7 @@
     help='Dry run: do not actually communicate with Metamist or Hail Batch, '
     'instead only print a final config and stages to be run',
 )
-@click.option(
-    '--verbose',
-    'verbose',
-    is_flag=True,
-)
+@click.option('--verbose', 'verbose', is_flag=True)
```
jmarshall, line 90: Better to add `help='…',` than to reformat this one so that it's different from all the others.

jmarshall: Perhaps I was unclear, so I've added a commit showing what I meant.

vivbak (approved):

PR #933, opened 2024-10-15
Use more efficient path checks

MattWellie:
# Problem

Our pipeline regularly takes over an hour just to start up...

# Solution?

This issue has (at least) 2 components

1. Time taken doing existence checks in GCP
2. Time taken scheduling jobs

2 might just be an intractable Hail Batch problem, so shelve that for now...

1 - the solution for this already exists in the `cpg_workflows.utils` module. What we currently do is a cached per-path existence check. That saves us no time, as we only check each path once anyway.

The `check_exists_path` instead takes the path it's given, finds the parent directory, and does a cached check on the whole directory at once. That means that `<bucket>/CPGA.vcf` and `<bucket>/CPGB.vcf` are only a single cached glob call in GCP.

We already wrote this method, we just don't use it during pipeline startup...

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.27.14
+current_version = 1.27.15
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.27.14
+  VERSION: 1.27.15
 
 jobs:
   docker:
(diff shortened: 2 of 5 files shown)


EddieLF (commented):
Love this change to speed up driver jobs.

Only thing we will need to take care with using it is for hail paths, the `exists` function has some handling for hail tables / matrix tables / hails VDSs, but this `check_exists_path` does not have that handling as far as I can tell. 

The changes to use it in the exomiser jobs should be fine, but will it be ok to change it in the `workflow.py` module too? I can see a driver job getting stuck on the iterdir() call if it's investigating the existence of a matrix table. But I haven't studied this too hard so I'm not sure.

MattWellie:
Firstly, I'm a silly fool - we _do_ use the better cached method... kinda. We currently have a chain of method calls:
1. is_reusable
2. exists (cached per path)
3. exists_not_cached
4. check_exists_path (cached per directory)

I don't know if the double layering of cached calls should help or hinder this process... It definitely doesn't have the speedup I'd expect. I still think the `exists` method we have here should be binned, but I don't really understand why it's currently so slow.

PR #937, opened 2024-10-17
Add config option for custom stripy report path prefix

EddieLF:
As we trial adding new loci to the STRipy reports using the `stripy.custom_loci_path` bed file config option, it would be useful to have the option of saving the reports do a different path aside from the standard `main-web/dataset/stripy/CPGID.html`. This would prevent overwriting existing reports while allowing us to analyse the value of additional loci, ad hoc.

By changing this config option, the report would be saved to `main-web/dataset/<stripy.output_prefix>/CPGID.html`.

Also fixes a bug with the Metamist status reporter used to write analyses - the `_update_meta_dict` function was being passed through to the analysis creation, but the resulting updated meta dict was not being saved to the `meta` variable. This fixes that, so the `update_analysis_meta` function now works and can be passed through to stage decorators.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.27.17
+current_version = 1.27.18
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


MattWellie (approved):
nice catch!

vivbak (approved):
🙌

PR #939, opened 2024-10-17
Rewrite the combiner stage in large cohort to be able to combine both GVCF and VDS files

Alexander-Stuckey:
Rewrite the combiner stage to allow for combining both GVCF and existing VDS files.

New:

- New combiner stage
- New runner for the large cohort pipeline (rather than `main.py`
- Apparently also reformat a ton of single quotes to double quotes.

Additional things to do:

- Write docstring for combiner stage
- Better logging (currently none for the combiner)
- Fix tests probably

Discussion points:

- Do we want to allow for custom interval lists?
- Do we want to allow for custom GVCF and VDS batch sizes in the combiner (just using default at the moment)
- Any other suggestions?

### DIFF
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -76,7 +76,7 @@ def run_combiner(
     cohort: str,
     tmp_prfx: str,
     existing_vds_ids: int | None = None,
-) -> VariantDataset:
+) -> None:
     input_vds_samples: list[str] = []
     use_genome_default_intervals: bool = False
     use_exome_default_intervals: bool = False
@@ -126,4 +126,3 @@ def run_combiner(
     )
 
     combiner.run()
-    return read_vds(str(output_vds_path))
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -1,128 +1,53 @@
-"""Runs the combiner.
+"""Runs the hail combiner
 
 Inputs:
-    File listing sequenging group IDs corresponding to the gvcf files you want to combine
-    File listing VDSes in gs:// format, one per line
-
-Raises:
-    ValueError: _description_
-    ValueError: _description_
-    ValueError: _description_
-    ValueError: _description_
-    ValueError: _description_
-    ValueError: _description_
-
-Returns:
-    _type_: _description_
+    - output_vds_path: str - The destination that the VDS will be saved at
+    - sequencing_type: str - Used to specify what intervals to use (exome or genome)
+    - tmp_prefix: str - Where to store temporary combiner plans
+    - genome_build: str - What reference genome to use for the combiner
+    - gvcf_paths: list[str] | None - The optional list of gvcf paths in string format to combine
+    - vds_paths: list[str] | None - The optional list of VDS paths in string format to combine
 """
 
-from itertools import chain
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
-from graphql import DocumentNode
+from hail.vds import new_combiner
 
-from hail.vds import VariantDataset, new_combiner, read_vds
-
-from cpg_utils import Path
 from cpg_utils.hail_batch import init_batch
-from metamist.graphql import gql, query
+from cpg_workflows.utils import can_reuse, to_path
 
 if TYPE_CHECKING:  # TCH002 https://docs.astral.sh/ruff/rules/typing-only-third-party-import/
     from hail.vds.combiner.variant_dataset_combiner import VariantDatasetCombiner
 
-# Get a cohort. Samples already in a VDS will be filtered out from the gvcf combiner list
-# and used as a VDS input
-GET_COHORT_QUERY: DocumentNode = gql(
-    """
-query getCohort($cohort: String!) {
-  cohorts(name: {eq: $cohort}) {
-    sequencingGroups {
-      id
-      analyses(type: {eq: "gvcf"}, status: {eq: COMPLETED}, active: {eq: true}) {
-        output
-      }
-    }
-  }
-}
-""",
-)
 
-GET_VDS_ANALYSIS_QUERY: DocumentNode = gql(
-    """
-query getVDSByAnalysisIds($vds_ids: [Int!]!) {
-  analyses(id: {in_: $vds_ids}) {
-    output
-  }
-}
-""",
-)
-
-
-def _get_samples_from_vds(input_vds: str) -> list[str]:
-    return read_vds(input_vds).variant_data.s.collect()
-
-
-def _parse_metamist_cohort_output(cohort_query_output: dict[str, Any]) -> list[dict[str, Any]]:
-    return [sg for sg in cohort_query_output["cohorts"][0]["sequencingGroups"]]
-
-
-def _parse_metamist_analysis_output(vds_analysis_query_output: dict[str, Any]) -> list[str]:
-    return [analysis["output"] for analysis in vds_analysis_query_output["analyses"]]
-
-
-def run_combiner(
-    output_vds_path: Path,
+def run(
+    output_vds_path: str,
     sequencing_type: str,
-    cohort: str,
-    tmp_prfx: str,
-    existing_vds_ids: int | None = None,
+    tmp_prefix: str,
+    genome_build: str,
+    gvcf_paths: list[str] | None = None,
+    vds_paths: list[str] | None = None,
 ) -> None:
-    input_vds_samples: list[str] = []
     use_genome_default_intervals: bool = False
     use_exome_default_intervals: bool = False
 
-    existing_vds: list[str] | None = None
-    project_gvcf_paths: list[str]
+    if not can_reuse(to_path(output_vds_path)):
+        init_batch()
 
-    cohort_query_results: dict[str, Any] = query(
-        GET_COHORT_QUERY,
-        variables={
-            "cohort": cohort,
-        },
-    )
+        if sequencing_type == "genome":
+            use_genome_default_intervals = True
+        elif sequencing_type == "exome":
+            use_exome_default_intervals = True
 
-    init_batch()
-    # Only support a single input cohort
-    formatted_query_results: list[dict[str, Any]] = _parse_metamist_cohort_output(cohort_query_results)
-
-    if existing_vds_ids:
-        vds_query_results: dict[str, Any] = query(
-            GET_VDS_ANALYSIS_QUERY,
-            variables={
-                "vds_ids": existing_vds_ids,
-            },
+        combiner: VariantDatasetCombiner = new_combiner(
+            output_path=output_vds_path,
+            gvcf_paths=gvcf_paths,
+            vds_paths=vds_paths,
+            reference_genome=genome_build,
+            temp_path=tmp_prefix,
+            use_exome_default_intervals=use_exome_default_intervals,
+            use_genome_default_intervals=use_genome_default_intervals,
+            force=True,
         )
-        existing_vds = _parse_metamist_analysis_output(vds_query_results)
-        input_vds_samples = list(chain(input_vds_samples, *[_get_samples_from_vds(vds) for vds in existing_vds]))
-        # Remove any samples from the query that are already present in a VDS
-        formatted_query_results = [sg for sg in formatted_query_results if sg["id"] not in input_vds_samples]
 
-    project_gvcf_paths = [entry["analyses"][0]["output"] for entry in formatted_query_results]
-
-    if sequencing_type == "genome":
-        use_genome_default_intervals = True
-    elif sequencing_type == "exome":
-        use_exome_default_intervals = True
-
-    combiner: VariantDatasetCombiner = new_combiner(
-        output_path=str(output_vds_path),
-        gvcf_paths=project_gvcf_paths,
-        vds_paths=existing_vds,
-        reference_genome="GRCh38",
-        temp_path=tmp_prfx,
-        use_exome_default_intervals=use_exome_default_intervals,
-        use_genome_default_intervals=use_genome_default_intervals,
-        force=True,
-    )
-
-    combiner.run()
+        combiner.run()
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -28,16 +28,8 @@ def run(
     gvcf_paths: list[str] | None = None,
     vds_paths: list[str] | None = None,
 ) -> None:
-    use_genome_default_intervals: bool = False
-    use_exome_default_intervals: bool = False
-
     if not can_reuse(to_path(output_vds_path)):
         init_batch()
-
-        if sequencing_type == "genome":
-            use_genome_default_intervals = True
-        elif sequencing_type == "exome":
-            use_exome_default_intervals = True
 
         combiner: VariantDatasetCombiner = new_combiner(
             output_path=output_vds_path,
@@ -45,8 +37,8 @@ def run(
             vds_paths=vds_paths,
             reference_genome=genome_build,
             temp_path=tmp_prefix,
-            use_exome_default_intervals=use_exome_default_intervals,
-            use_genome_default_intervals=use_genome_default_intervals,
+            use_exome_default_intervals=sequencing_type == 'exome',
+            use_genome_default_intervals=sequencing_type == 'genome',
             force=True,
         )
 
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -1,132 +1,129 @@
-import collections
-import logging
+"""Runs the combiner.
 
-import hail as hl
+Inputs:
+    File listing sequenging group IDs corresponding to the gvcf files you want to combine
+    File listing VDSes in gs:// format, one per line
+
+Raises:
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+
+Returns:
+    _type_: _description_
+"""
+
+from itertools import chain
+from typing import TYPE_CHECKING, Any
+
+from graphql import DocumentNode
+
+from hail.vds import VariantDataset, new_combiner, read_vds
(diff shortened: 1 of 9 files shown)


cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
...
-    elif sequencing_type == 'exome':
-        params.setdefault('use_exome_default_intervals', True)
-    elif sequencing_type == 'genome':
-        params.setdefault('use_genome_default_intervals', True)
-    else:
-        raise ValueError(
-            'Either combiner/intervals must be set, or workflow/sequencing_type must be one of: "exome", "genome"',
-        )
+def _get_samples_from_vds(input_vds: str) -> list[str]:
+    return read_vds(input_vds).variant_data.s.collect()
```
MattWellie, line 62: In an ideal world, the analysis entry in metamist which points to this VDS will contain an `analysis.sequencingGroups[.id]` attribute, collecting all the SG IDs which were part of generating the previous object.

However, it looks like the analysis results thus far haven't been registered at all ([here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/large_cohort.py#L19)). You've addressed that, so it should be solved going forwards, in which case you _shouldn't_ need this check to be done on the file itself.

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
...
+
+    if existing_vds_ids:
+        vds_query_results: dict[str, Any] = query(
+            GET_VDS_ANALYSIS_QUERY,
+            variables={
+                "vds_ids": existing_vds_ids,
+            },
+        )
+        existing_vds = _parse_metamist_analysis_output(vds_query_results)
+        input_vds_samples = list(chain(input_vds_samples, *[_get_samples_from_vds(vds) for vds in existing_vds]))
```
MattWellie, line 106: I guess in theory you can merge as many VDSs as you want, but would this always be an increment based on the latest VDS, or is there a scenario where you'd want a union across many?

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
...
+        temp_path=tmp_prfx,
+        use_exome_default_intervals=use_exome_default_intervals,
+        use_genome_default_intervals=use_genome_default_intervals,
         force=True,
-        **params,
     )
+
     combiner.run()
-    return hl.vds.read_vds(str(out_vds_path))
+    return read_vds(str(output_vds_path))
```
MattWellie, line 129:
```suggestion
```

I could be wrong, but returning here doesn't do anything. Same in the current implementation. This is run as a python job in a container that's torn down once the VDS is written, so this return goes nowhere

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
...
+    sequencingGroups {
+      id
+      analyses(type: {eq: "gvcf"}, status: {eq: COMPLETED}, active: {eq: true}) {
+        output
+      }
+    }
+  }
+}
+""",
+)
```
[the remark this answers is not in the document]
MattWellie: Ah wait... I see. Because this is in a separate runtime from the pipeline workflow, you have to re-query for the Cohort content.

IMO it would be cleaner and easier to do this prep work in the Stage:
- Query for latest VDSs
- Get the SG IDs in those VDSs from the analysis meta (or direct from the object as a fallback)
- Use the already populated Cohort (MultiCohort?) to get the universe of CPG IDs in this analysis
- Back filter all the CPG IDs we have in the VDS out of the to-do list
- Use the already populated SG.gvcf attributes to get the string path to each gVCF that needs adding
- Pass each new gVCF path to include as an argument (or persist to a temp file that's opened inside the combiner job for scalability)
- Pass the VDS path(s) to combine if any, instead of having to re-query from metamist in the Job

This would make the most of the objects already populated when the Stage runs instead of reinventing the wheel by re-querying for the same data.

That would also mean that this Combiner stage has a much tighter scope - "take 0 or more VDS paths, and 0 or more gVCFs, and make a new VDS". That would be much more versatile for code re-use

cpg_workflows/stages/large_cohort.py
```diff
@@ -13,72 +13,63 @@
...
-                "None of 'highmem_workers', 'highmem_drivers', or 'driver_cores' were specified in the config. If you're getting OOM errors, ensure these are included in the config.",
-            )
+        output_vds_path: Path = self.expected_outputs(cohort)
+        tmp_prfx = slugify(
+            f"{self.tmp_prefix}/{workflow_config['cohort']}-{workflow_config['sequencing_type']}-{combiner_config['vds_version']}",
+        )
 
-        j.image(image_path('cpg_workflows'))
-        j.command(
-            query_command(
```
MattWellie, line 52: YEAH GET IN THE BIN

MattWellie (changes requested):
Looks like it will all work, but it involves re-querying metamist for a ton of content that has already been obtained at the Stage level. Probably a lot of simplification if the steps are moved into different places

cpg_workflows/large_cohort/combiner.py
MattWellie: actually that's not entirely true - Hail can try to return values from PythonJobs, which is done through pickling. A whole cohort VDS should absolutely not be pickled for a return ⚠️

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
-import collections
-import logging
+"""Runs the combiner.
 
-import hail as hl
+Inputs:
+    File listing sequenging group IDs corresponding to the gvcf files you want to combine
```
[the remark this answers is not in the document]
Alexander-Stuckey: Oh this comment is totally dead (need to properly write and update the docstring). For combining GVCFs, you have to define a custom cohort in Metamist, and pass that ID in the config. It'll then grab the GVCFs from there.

cpg_workflows/large_cohort/combiner.py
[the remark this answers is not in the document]
Alexander-Stuckey: I like this suggestion, I'll do that. Especially since it does clean up the combiner and limit its scope as you say.

cpg_workflows/large_cohort/combiner.py
Alexander-Stuckey: This is mostly just for the base case where you might have generated a bunch of VDSs already, and want to merge them all. Going forward from that point it'd be almost always the case that you'd be adding some extra samples to an existing VDS.

cpg_workflows/large_cohort/combiner.py
Alexander-Stuckey: Fair, was just copying from the previous code. I can remove this `return`

Alexander-Stuckey: Return removed

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,129 @@
...
+    File listing VDSes in gs:// format, one per line
+
+Raises:
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
+    ValueError: _description_
```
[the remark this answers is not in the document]
Alexander-Stuckey: Fixed docstring

cpg_workflows/large_cohort/combiner.py
```diff
@@ -1,132 +1,53 @@
...
+            use_exome_default_intervals = True
+
+        combiner: VariantDatasetCombiner = new_combiner(
+            output_path=output_vds_path,
+            gvcf_paths=gvcf_paths,
+            vds_paths=vds_paths,
+            reference_genome=genome_build,
+            temp_path=tmp_prefix,
+            use_exome_default_intervals=use_exome_default_intervals,
+            use_genome_default_intervals=use_genome_default_intervals,
```
MattWellie, line 49:
```suggestion
    if not can_reuse(to_path(output_vds_path)):
        init_batch()

        combiner: VariantDatasetCombiner = new_combiner(
            output_path=output_vds_path,
            gvcf_paths=gvcf_paths,
            vds_paths=vds_paths,
            reference_genome=genome_build,
            temp_path=tmp_prefix,
            use_exome_default_intervals=sequencing_type == 'exome',
            use_genome_default_intervals=sequencing_type == 'genome',
```

cpg_workflows/metamist.py
```diff
@@ -127,10 +127,10 @@
 )
 
 
-_metamist: Optional['Metamist'] = None
+_metamist: Optional["Metamist"] = None
```
MattWellie, line 130: this file has been reformatted to have double instead of single quotes - I don't think there's any changes, but hard to tell at a glance. can you reformat it with `ruff format --config "format.quote-style = 'single'"` to be consistent with the rest of the repo?

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
 
-from .genotype import Genotype
+if TYPE_CHECKING:
+    from graphql import DocumentNode
 
+    from hailtop.batch.job import PythonJob
 
-@stage(required_stages=[Genotype])
+
+@stage(analysis_type="combiner", analysis_keys=["combiner"])
```
MattWellie, line 25:
```suggestion
@stage(analysis_type="combiner")
```

Relevant code [here](https://github.com/populationgenomics/production-pipelines/blob/d5bf2b8ae33b5274acf16915ebde1075302a2316/cpg_workflows/workflow.py#L562) - if your `expected_outputs` method returns a dict, `analysis_keys` tells it which entry in the dict to register as an output. If the return is a Path, `analysis_keys` is ignored. I don't think it would break, but this arg conflicts with the lack of dictionary return here.

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
-        return cohort.analysis_dataset.prefix() / 'vds' / f'{vds_version}.vds'
+        workflow_config = config_retrieve("workflow")
+        combiner_config = config_retrieve("combiner")
+        output_vds_name: str = slugify(
+            f"{workflow_config['cohort']}-{workflow_config['sequencing_type']}-{combiner_config['vds_version']}",
+        )
+        return cohort.analysis_dataset.prefix() / "vds" / f"{output_vds_name}.vds"
+
+    def get_vds_ids_output(self, vds_id: int) -> list[str]:
+        get_vds_analysis_query: DocumentNode = gql(
```
MattWellie, line 36: If you create this at the root of the script (as with [this](https://github.com/populationgenomics/production-pipelines/blob/d5bf2b8ae33b5274acf16915ebde1075302a2316/cpg_workflows/metamist.py#L34)) you can do GraphQL validation in unit tests. Not that we do that, but you can.

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
+                    sequencingGroups {
+                        id
+                    }
+                }
+            }
+        """,
+        )
+        query_results: dict[str, Any] = query(get_vds_analysis_query, variables={"vds_id": vds_id})
+        vds_path: list[str] = [query_results["analyses"][0]["output"]]
+        vds_ids: list[str] = [sg["id"] for sg in query_results["analyses"][0]["sequencingGroups"]]
```
MattWellie, line 50:
```suggestion
        vds_sgids: list[str] = [sg["id"] for sg in query_results["analyses"][0]["sequencingGroups"]]
```

Can you rename to `vds_sgids` to differentiate from `vds_id` above which is a very different ID?

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
+        return vds_path + vds_ids
```
MattWellie, line 51: This feels petty, but a return type of `tuple[str | Path, list[str]]` would separate these two and present a clearer interface
```suggestion
        return vds_path, vds_ids
```

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
-            init_batch_args['driver_memory'] = 'highmem'
-        if 'driver_cores' in config:
-            init_batch_args['driver_cores'] = config['driver_cores']
-        if not init_batch_args:
-            logging.warning(
-                "None of 'highmem_workers', 'highmem_drivers', or 'driver_cores' were specified in the config. If you're getting OOM errors, ensure these are included in the config.",
-            )
+        output_vds_path: Path = self.expected_outputs(cohort)
+        tmp_prefix = slugify(
+            f"{self.tmp_prefix}/{workflow_config['cohort']}-{workflow_config['sequencing_type']}-{combiner_config['vds_version']}",
```
MattWellie, line 63: The config file allows for multiple cohorts to be used (`workflow.input_cohorts: list[str]`), and this is a `CohortStage`, so by definition it runs separately for each cohort. Once we get the ID/name shenanigans sorted you'd be better off using `cohort.name/id`, deriving it from the `Cohort` Target which is an input to this `queue_jobs` method, instead of going back to the config and picking up one config String.

That would hypothetically mean you can run this for several cohorts in parallel...

If you're not going to use it that way, this should be made into a MultiCohort stage (runs only once per analysis)

cpg_workflows/stages/large_cohort.py
MattWellie: something similar is referred to as `vds_sg_ids` below, so would have to avoid a clash there...

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
+                query_res: list[str] = self.get_vds_ids_output(vds_id)
+                vds_paths.append(query_res[0])
+                vds_sg_ids = vds_sg_ids + query_res[1:]
+
+        # Get SG IDs from the cohort object itself, rather than call Metamist.
+        # Get VDS IDs first and filter out from this list
+        sg_ids: list[SequencingGroup] = cohort.get_sequencing_groups(only_active=True)
+        new_sg_gvcfs = [str(sg.gvcf) for sg in sg_ids if vds_sg_ids and sg.id not in vds_sg_ids]
+        if len(new_sg_gvcfs) == 0:
+            new_sg_gvcfs = None
```
MattWellie, line 83: maybe a check here for `if gVCFs == 0 and VDSs == 1, just quit`

cpg_workflows/stages/large_cohort.py
```diff
@@ -12,73 +14,110 @@
...
         )
+
         return self.make_outputs(cohort, self.expected_outputs(cohort), [j])
 
 
 @stage(required_stages=[Combiner])
 class SampleQC(CohortStage):
     def expected_outputs(self, cohort: Cohort) -> Path:
-        if sample_qc_version := config_retrieve(['large_cohort', 'output_versions', 'sample_qc'], default=None):
+        if sample_qc_version := config_retrieve(["large_cohort", "output_versions", "sample_qc"], default=None):
```
MattWellie, line 106: some more quote fiddlin'

large_cohort_runner.py
```diff
@@ -0,0 +1,100 @@
...
+from cpg_workflows.stages.large_cohort import AncestryPlots, Frequencies, LoadVqsr
+from cpg_workflows.workflow import StageDecorator, run_workflow
+
+WORKFLOWS: dict[str, list[StageDecorator]] = {
+    "large_cohort": [
+        LoadVqsr,
+        Frequencies,
+        AncestryPlots,
+    ],
+}
```
MattWellie, line 24: I'd be tempted to go nuclear here... the Workflows dict and a workflow CLI parameter are just for multi-pipeline flexibility. This whole script can be rewritten as:

```
    # even this section isn't necessary if you give the analysis runner argument `--config configs/defaults/large_cohort.toml `
    workflows = 'large_cohort'
    wfl_conf_path = to_path(__file__).parent / f"configs/defaults/{workflow}.toml"
    config_paths = os.environ["CPG_CONFIG_PATH"].split(",") + list(config_paths)
    set_config_paths(config_paths[:1] + [str(wfl_conf_path)] + config_paths[1:])

    # just run these stages
    run_workflow(stages=[LoadVqsr, Frequencies, AncestryPlots], dry_run=dry_run)
```

MattWellie (changes requested):

cpg_workflows/metamist.py
Alexander-Stuckey: Sure

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey: You know, I had that originally, but changed it to a list where the first element was the VDS, just to be lazy. I'll make it nicer.

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey: Yes on the `cohort.name` (this is how it came up in the first place).

On Cohort vs MultiCohort, I'm not sure what direction PA wants. I'll ask @katiedelange and see.

large_cohort_runner.py
Alexander-Stuckey: Will get to this when large cohort is split out as it's own repo

MattWellie (approved):
LGTMatt

jmarshall:
I see this adds a new entry point, _large_cohort_runner.py_, alongside _main.py_.

It largely duplicates _main.py_. What is the reason for adding it? This duplication would seem to add a maintenance burden.

vivbak:
Sorry @jmarshall! I Ok'd this. As part of the refactor each pipeline will get it's own entry-point anyhow, and I believe it simplified things from a development perspective.

jmarshall:
I see, it's just getting ahead of the refactoring, in what might well be its eventual final form. Fair enough, thanks.

@Alexander-Stuckey et al: It would be good to mention this sort of thing in the commit messages and/or PR comments so that this sort of context is captured for reviewers and posterity.

Alexander-Stuckey:
Fair point @jmarshall , I only passingly mentioned it in the initial PR message and didn't include the rationale. But yes, it is indeed just getting a tiny headstart on the refactoring into a separate repo eventually.

PR #942, opened 2024-10-18
Talos uses phenopackets

MattWellie:
Counterpart to https://github.com/populationgenomics/talos/pull/446

Some arguments changed. some stages dropped, one new one added. It's a brave new world, get on board

- removes the Talos/GeneratePED call - we instead make a basic Pedigree in prod-pipes
- add a MakePhenopackets call to generate cohort data in the new format
- adapt to the CLI argument harmonisation (all modules take explicit flags, not ordered params, `input` and `output` preferred)


cpg_workflows/stages/talos.py
```diff
@@ -16,10 +16,9 @@
 [cohorts.DATASET_NAME.genome]  # repeated for exome if appropriate
 - DATASET_NAME is taken from config[workflow][dataset]
 - historic_results, str, path to historic results directory for this project. If
-  present, new genes and new variant categories will be identified based on the
-  state from the previous run. If absent, panel searching will default to using
-  gene_prior above, and the results from this current run will not be stored in
-  a suitable format to inform the next run.
+  present, new variant categories will be identified based on the state from the
+  previous run. If absent, the results from this current run will not be stored
```
cassimons, line 20: ???

```suggestion
  previous run. If absent, the results from this current run will be stored
```

cassimons (approved):

cpg_workflows/stages/talos.py
MattWellie: The author of this PR has made some... interesting changes

MattWellie: This specific change is weird, but in the context it makes sense

RE: the `historic results` location in the configuration

"""
historic_results, str, path to historic results directory for this project. If
  present, new variant categories will be identified based on the state from the
  previous run. If absent, the results from this current run will not be stored
  in a suitable format to inform the next run.
"""

Just letting a user know that without specifying a location to save the history file(s), the format required to carry into the next run isn't present

cassimons: Ah go ya. I was reading it so the "if absent" was referring to the file not the argument, ie: "if the history file does not exist, one will be created that can be used in future runs" Perhaps just rephrase it to be explicit.

PR #946, opened 2024-10-23
Speed up the slowdown

MattWellie:
The update to the combiner only tolerates exome or genome intervals when combining. The previous test case was limited to only a couple of small chromosomes, to keep timings in check. 

This reinstates the ability to pass custom intervals to the combiner, and passes them inside the test suite


Alexander-Stuckey (approved):
Looks good to me

PR #948, opened 2024-10-24
Uncomment the sitonly expected output for SiteOnlyVCFs

EddieLF:
Small fix needed to this pipeline - this was commented out late into development, as I was considering dropping support for the `scatter_count = 1` option. 

This output isn't actually created or used except for when `scatter_count = 1`, which is not recommended. However it needs to be in the expected outputs for the VEP job to work. 

See the error mode:
https://batch.hail.populationgenomics.org.au/batches/518674/jobs/1

Test run on this branch with this fix, using `scatter_count = 3`
https://batch.hail.populationgenomics.org.au/batches/518683/jobs/1

### DIFF
--- a/cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
+++ b/cpg_workflows/stages/seqr_loader_long_read/long_read_snps_indels_annotation.py
@@ -209,7 +209,7 @@ class SiteOnlyVCFs(MultiCohortStage):
         Generate site-only VCFs from the merged VCF.
         """
         return {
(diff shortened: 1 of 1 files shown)


MattWellie (approved):

MattWellie:
It feels like there's a neater solution here...

MattWellie:
Oh yeah I'm familiar with this trash bag. It's nonsense that the VEP module requires so many different inputs... It either takes the separate VCFs, or one VCF, then contains a full code block for splitting into segments. It's just not relevant to running VEP and feels like terrible bloat.

I've tried grooming this out a few times, but it always ends up being a really invasive change

PR #949, opened 2024-10-24
Remove multiple VCF interfaces in VEP

MattWellie:
The VEP module takes a mandatory argument of a whole-genome/exome VCF, and optionally several subdivided VCFs. The latter is what we use in the pipeline, as these VCFs have already been subdivided by prior stages according to `scatter_count`.

The VEP module contains a lot of bloat - it allows for a single VCF but a high scatter count to be provided, in which case it generates a set of intervals and subdivides the input VCF.... This shouldn't be the remit of the VEP task.

This PR removes this option and takes a single argument - a list of VCFs. This list can be of length 1, equivalent to a full VCF and scatter == 1.


EddieLF (approved):
Great idea, love this solution

PR #950, opened 2024-10-24
Add robustness to gql SG query

MattWellie:
The backoff wrapper around query in metamist [here](https://github.com/populationgenomics/metamist/blob/d433e6213632e4d970d8818306288174d1724202/metamist/graphql/__init__.py#L145-L151), which I @MattWellie implemented, is trash. It looks (to me) like it's parameterised appropriately, but the pauses between re-attempts are a joke, e.g.:

```
2024-10-24 01:23:33 INFO (backoff 105): Backing off query(...) for 0.3s (gql.transport.exceptions.TransportServerError: 504 Server Error: Gateway Timeout for url: https://sample-metadata-api-mnrpw3mdza-ts.a.run.app/graphql)
2024-10-24 01:23:33 INFO (root 342): Getting sequencing groups for dataset perth-neuro
2024-10-24 01:23:34 INFO (root 342): Getting sequencing groups for dataset ravenscroft-arch
2024-10-24 01:28:34 INFO (backoff 105): Backing off query(...) for 0.5s (gql.transport.exceptions.TransportServerError: 504 Server Error: Gateway Timeout for url: https://sample-metadata-api-mnrpw3mdza-ts.a.run.app/graphql)
2024-10-24 01:33:35 ERROR (backoff 120): Giving up query(...) after 2 tries (gql.transport.exceptions.TransportServerError: 504 Server Error: Gateway Timeout for url: https://sample-metadata-api-mnrpw3mdza-ts.a.run.app/graphql)
```

Three failures and one successful call, all in 2 seconds. The pipeline startup still does a ton of Metamist interaction, and when setting up for more than ~5 projects, the startup routinely fails. [Example job](https://batch.hail.populationgenomics.org.au/batches/518690/jobs/1).

Elsewhere in metamist.py @milo-hyben has added tenacity decorators to retry failing calls, usually with a wrapper method to avoid directly wrapping the gql query. I don't know whether this or that is a better fix, but please can we have some more robustness during pipeline setup.


milo-hyben (approved):
Hey Matt, very good point regarding inconsistency.
Usually REST API do not implement backoff, it is up to a client consuming those APIs.
I guess that would be reason why none of our metamist REST API have it.
I had created my wrapper around RESTFull metamist API in production pipelines.

Regarding Metamist GraphQL Query, that is kind of different API, it is not RESTfull and I suspect idea was to add backoff there at some stage in the past.

I do not see any problem with your implementation of backoff in production-pipelines, but we should really get it fixed in Metamist GraphQL Query part.
I had created a JIRA task for fixing metamist (SET-220).

PR #951, opened 2024-10-24
Cohort ID vs Cohort Name

vivbak:
Cohort IDs have been incorrectly referred to as the Cohort name and this is causing confusion. 

See https://centrepopgen.slack.com/archives/C030X7WGFCL/p1729124262948239 

This aims to rename the variables accordingly. 

A decision was made for now to stick with accepting the IDs in the .toml files.


vivbak:
Hello @MattWellie,

Thanks for your patience on this change! An update for you: 

There are 25 instances of cohort.name in the pipelines that need to be investigated (i.e. should we now change them to cohort.id), but it's hard for me to make the calls myself. If you have some time to take a look that would be great. 

Only one test is failing, and its the seqr-loader, I think its likely one of these instances that needs to be fixed that's the problem. It can't find an output because it appears keyed on a cohort name instead of a cohort ID. I'm trying to figure out which one, but if you end up making some of these calls it may resolve anyway. 

FYI @Alexander-Stuckey

MattWellie:
> Only one test is failing, and its the seqr-loader

I actually think this is caused by seqr-loader not being migrated from Cohorts to MultiCohorts, there shouldn't be any Cohort steps in that pipeline. I'll have a look.

cpg_workflows/metamist.py
```diff
@@ -126,6 +126,17 @@
     """,
 )
 
+GET_COHORT_NAMES_QUERY = gql(
```
MattWellie, line 129: can we drop this whole query and just return `id` when we query for the SGs in a cohort above. Maybe slightly rephrasing that `SGByCohortQuery` query as `GetCohortContents` to be more specific

cpg_workflows/inputs.py
```diff
@@ -56,10 +56,11 @@ def create_multicohort() -> MultiCohort:
     multicohort = MultiCohort()
 
     datasets_by_cohort = get_metamist().get_sgs_for_cohorts(custom_cohort_ids)
+    cohort_names_by_id = get_metamist().get_cohort_name_by_ids(custom_cohort_ids)
```
MattWellie, line 59: as previous comment, we can roll this into the same query as the SGs

MattWellie:
The failing test here would be corrected by [this PR](https://github.com/populationgenomics/production-pipelines/pull/955)

cpg_workflows/metamist.py
vivbak: yeah I thought the separate query may be clear-er, but it does add an additional call we don't really need. good catch!

MattWellie (approved):
How exciting!

PR #952, opened 2024-10-24
Add the --sv flag to the modify_sniffles command for SV VCFs

EddieLF:
SV VCFs were not being reformatted correctly because of the new --sv argument to modify_sniffles, which was added in #912 to accomodate both SNPs_Indels VCFs alongside the SV VCFs.


MattWellie (approved):
LGTMatt, will just need a force re-run from that stage

PR #955, opened 2024-10-25
Joint genotyping qc to multicohorts

MattWellie:
Simple uplift of ClinvArbitration workflow to use Cohorts

Slightly more involved uplift of JointVcfQC and JointVcfMultiQC (note, these are not terminal stages in any workflow, so we... don't run them in production)
- Cohort -> MultiCohort
- Corrections to imports
- Only call self.expected_outputs() once
- Use built-in self.prefix and self.web_prefix (these Stages seem to predate the builtin Stage methods)

All tests and linting pass locally for me


EddieLF (approved):
Looks good to me

PR #956, opened 2024-10-25
Reduce repeated method calls

MattWellie:
Picked up during https://github.com/populationgenomics/production-pipelines/pull/955

We have a bad habit of unnecessarily repeated method calls. Two main offenders are treated here:

The first is Stage B picking up multiple outputs from Stage A. Instead of grabbing both at once:
```
StageA_Inputs = inputs.as_dict(target, StageA)
in_1 = StageA_Inputs['one']
in_2 = StageA_Inputs['two']
```

We regularly get each input separately:
```
in_1 = inputs.as_dict(target, StageA, 'one')
in_2 = inputs.as_dict(target, StageA, 'two')
```

Seems pretty innocent, but duplicates the method calls, most of which cascade down through continued regeneration of the alignment_inputs_hash. We're probably doing more hashing than a Bitcoin miner, and no $BTC to show for it.

Second example is the repeated generation of the expected_outputs, e.g. 
```
json_path = self.expected_outputs(dataset)['json']
html_path = self.expected_outputs(dataset)['html']
checks_path = self.expected_outputs(dataset)['checks']
...
return self.make_outputs(dataset, data=self.expected_outputs(dataset), jobs=jobs)
```

Which could be simplified down to 
```
outputs = self.expected_outputs(dataset)
json_path = outputs['json']
html_path = outputs['html']
checks_path = outputs['checks']
...
return self.make_outputs(dataset, data=outputs, jobs=jobs)
```

Resulting in only a single call to `self.expected_outputs` instead of 4, and all the resulting non-calls of the hashing algorithm.

I don't _know_ how much time this will save, but we have pipelines that take 40 mins to start up, and this probably isn't helping.

### DIFF
--- a/cpg_workflows/stages/align.py
+++ b/cpg_workflows/stages/align.py
@@ -56,6 +56,8 @@ class Align(SequencingGroupStage):
             / f'{sequencing_group.id}.markduplicates-metrics'
         )
 
+        outputs = self.expected_outputs(sequencing_group)
+
         try:
             jobs = align.align(
                 b=get_batch(),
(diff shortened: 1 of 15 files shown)


MattWellie:
Closing this PR, it's substantially mitigated by #958, and muddies the water with additional spurious PRs at this time

PR #958, opened 2024-10-27
Create alignment hash on first get, reuse thereafter

MattWellie:
# Problem

* the pipeline can take ages to start up

## Why?

* In prod-pipes/cpg-flow we use 'output version', a hashed String, to keep our analysis results stored in an appropriate way. The `output_version` is generated from the hash of all the sequencing groups that relate to a Target (e.g. all the **CRAMs** relating to the **SGs** in this **Dataset**); if the Dataset contents change in any way, the `output_version` automatically changes, changing the output paths so that different pipeline runs don't attempt to overwrite each others' files.
* Every individual call to `self.prefix`, `self.tmp_prefix`, or similar Target properties does the following:
    1. Scan the Target for all SGs
    2. Get the Alignment Input for each SG
    3. Sort the Alignment Input Strings
    4. Generate a Hash from the sorted list
* This is wild, considering that we don't just call this once per Stage, we sometimes call it once per individual output file from a stage ([example](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/gatk_sv/gatk_sv_multisample.py#L749-L753)), and worse (as covered on #956) we frequently call the method that generates those outputs paths multiple times within a single Stage.
* The maths here is mind boggling - we call this hash method multiple times per (`# target instances` x `# self.*prefix calls` x `# stages in workflow`).

## Solution

This solution is simple - each Target is initialised with an un-populated `self.alignment_inputs_hash` variable. Instead of calling the hashing method directly each time, which cannot be cached, we create a getter and setter.

We call the getter when we need a hash - if the variable hasn't been instantiated yet, we first set it, then get it. This is incredibly simple, and doesn't create any stinky cache problems or circular imports.

## Why this is good

The workflow setup is in two discrete steps

1. Populating the Targets (Cohort/Dataset/MultiCohort)
2. Running the Stages

Part 1 is always completed before Part 2 begins, and the Targets cannot be modified during execution of Stages. In Part 1 the SGs are loaded into each Target, in Part 2 the `alignment_inputs_hash` are used to generate output paths.

This guarantees that every call to `alignment_inputs_hash` within a Target/pipeline run is guaranteed to be identical, which is why it is used in building output paths. As it is guaranteed to be identical, we should stop wasting time and cache the result the first time it's generated.


MattWellie:
This is being closed as included in the larger-scope change #959

vivbak (approved):

PR #959, opened 2024-10-28
Separate cohorts from datasets, NO hash changes

MattWellie:
Really extensive changes - this probably requires a whole RFC worth of documentation. Notes in brief:

- The MultiCohort changes were initially created with a hierarchy of MCs consist of Cohorts, consist of Datasets, consist of SGs. This was a fault with the initial specification of the feature, happy to take some flak for that.
- In reality we don't want a division of Datasets within Cohorts, but it's the way we currently set up our pipelines. Bugs like #903 resulted from the previous attempt to re-work this, because we still have the underlying structure (Cohort Targets have Datasets, which is wrong, and we have messy work-arounds)

The structure we currently have when setting up a pipeline using custom cohorts (`inputs.create_multicohort()`):

* query Metamist for each cohort
* stratify the SGs in that cohort by Dataset
* iterate over each Cohort's SGs in dictionary form
  * iterate over the datasets in that Cohort
    * Create a Dataset within the Cohort
    * Create all SGs in that Dataset
* For each Dataset in this Cohort, go to Metamist and get the analysis entries, and pedigree data

Then once we've obtained all that info:

* for each Cohort in the MultiCohort
  * for each Dataset in the Cohort
    * create or fetch a Dataset in the MultiCohort (this is where #903 was needed)
    * also add all the SGs directly to the MultiCohort's Dataset

Most of this structure is just in the way. There is still ongoing mess during the workflows - `Cohort.get_sequencing_groups()` still loops over all Datasets in the Cohort, then all SGs in each Dataset. 

---

This PR alters the structure in a lot of ways:

- Query Metamist for each Cohort. Do not stratify the output by Dataset
- Create a Cohort object in the MultiCohort
- For each SequencingGroup:
  - Identify the Dataset the SG is in, and create or fetch that Dataset at the MultiCohort level
  - Add the SG to the Dataset
  - Also add the SG to the Cohort

Simple, once this loop is complete we've populated all the Cohorts, Datasets, and MultiCohorts in a way which is only as nested as it needs to be (MCs have Cohorts and Datasets. Datasets don't have Cohorts, Cohorts don't have Datasets). Once populated, we do the following:

- Once per Dataset, go to Metamist for the Analysis entries and Pedigree data.

This means we only make the minimum number of Metamist queries. This simplification extends to Workflows:

* `get_sequencing_groups()` methods at the Cohort and Dataset level both just return values from their dictionary of SGs by ID. Ideally the MultiCohort should also have a dictionary of SGs by ID, but there's more thought required on whether to/how to support a single SG being in multiple Cohorts
* each Class instance is unique, with no copy/deep copying required.

The fun part was how much code can be deleted:

- `get_sgs_by_project_from_cohort` and the method that calls it, `get_sgs_for_cohorts` are simplified to one simple metamist call
- In the Cohort Class, `get_datasets`, `add_dataset`, `create_dataset`, and `get_dataset_by_name` are all deleted. So is the attribute `_datasets_by_name`
- The Dataset Constructor no longer requires a Cohort, that would be inappropriate

n.b. the diff in `cpg_workflows.targets.py` is brutal, because the Class definitions are now in hierarchical order: Sex, SG, PedigreeInfo, Dataset & Cohort, MultiCohort. This rearrangement isn't well handled by git. This has been handled so far by using fake type hints, e.g. [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/targets.py#L608). It's better just to put the Classes in the right order.

# Testing

Other than adding to/amending the tests we currently have in place, I've done a dry-run of a `gCNV` workflow set up. This workflow took 1 hour 37 minutes to fail during set up when run in production, and has bothered me ever since. 

- Using the code in main, the hail.dry_run of the workflow took 6 mins 49 seconds
- Using this code took 3 mins 35 seconds

That's substantial speed up of the workflow setup process, and a net reduction in complexity. All the logging/queued jobs were identical between all executions of the pipeline, as would be expected if this change doesn't alter the paths/cohorts/datasets.


cpg_workflows/workflow.py
```diff
@@ -467,15 +461,6 @@ def expected_outputs(self, target: TargetT) -> ExpectedResultT:
         Can be a str, a Path object, or a dictionary of str/Path objects.
         """
 
-    def deprecated_queue_for_cohort(self, cohort: Cohort) -> dict[str, StageOutput | None]:
```
MattWellie, line 470: There isn't a single instance of `deprecated_queue_for_cohort` being called in this whole codebase, and we're pretty far into the MultiCohort changes, so instead of renovating these methods we should delete them.

cpg_workflows/workflow.py
```diff
@@ -682,7 +667,6 @@ def _get_action(self, target: TargetT) -> Action:
                 logging.info(f'{self.name}: {target} [REUSE] (expected outputs exist: {expected_out})')
                 return Action.REUSE
 
-        logging.info(f'{self.name}: {target} [QUEUE]')
```
MattWellie, line 685: This is repeated on line 1324 in `queue_for_multicohort` - we log the name, multicohort, and action.name for every stage in the analysis. The result is that this line is printed twice for every Stage instance

cpg_workflows/workflow.py
```diff
@@ -697,18 +681,18 @@ def _is_reusable(self, expected_out: ExpectedResultT) -> tuple[bool, Path | None
                 Path | None: first missing path, if any
         """
         if self.assume_outputs_exist:
-            logging.info(f'Assuming outputs exist. Expected output is {expected_out}')
+            logging.debug(f'Assuming outputs exist. Expected output is {expected_out}')
```
MattWellie, line 684: I've set a couple of these messages to be printed only when we use `--verbose` to up the logging level to debug. Reasoning here is that the code as it stands both prints all the paths it intends to check, then the paths which don't exist. This really clouds the logging when resuming runs which have partially completed, when what we really want to know is the paths that were expected but don't exist, which is printed below (line 707).

cpg_workflows/workflow.py
```diff
@@ -1231,54 +1215,6 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
         """
         pass
 
-    def deprecated_queue_for_cohort(self, cohort: Cohort) -> dict[str, StageOutput | None]:
```
MattWellie, line 1234: Dead. RIP. Delete. Bye.

test/stages/__init__.py
```diff
@@ -22,7 +22,7 @@
 )
 
 
-def add_sg(ds, id, external_id: str):
+def add_sg(ds, id, external_id: str) -> SequencingGroup:
```
MattWellie, line 25: All changes in the test setup is just reflecting that we no longer create cohorts, create datasets in those cohorts, then create SGs.

test/stages/__init__.py
```diff
@@ -74,29 +71,22 @@ def mock_multicohort() -> MultiCohort:
...
-    ds2 = cohort_a.create_dataset('projectc')
-    dm2 = mc.add_dataset(ds2)
-    sg3 = add_sg(ds2, 'CPGCCCC', external_id='SAMPLE3')
-    sg4 = add_sg(ds2, 'CPGDDDD', external_id='SAMPLE4')
-    dm2.add_sequencing_group_object(sg3)
-    dm2.add_sequencing_group_object(sg4)
+    ds = mc.create_dataset('projecta')
+
+    # Add sequencing groups to the cohort AND dataset
+    cohort_a.add_sequencing_group_object(add_sg(ds, 'CPGXXXX', external_id='SAMPLE1'))
```
MattWellie, line 77: This is an inner call to `add_sgs` to create the SG in the Dataset, which returns the SG. That SG is fed directly into `Cohort. add_sequencing_group_object`. Saves an extra line.

test/test_multicohort.py
```diff
@@ -184,26 +183,18 @@ def mock_get_sgs_by_cohort_project_b(*args, **kwargs) -> list[dict]:
     ]
 
 
-def mock_get_cohorts(*args, **kwargs) -> dict:
+def mock_get_cohorts(cohort_id: str, *args, **kwargs) -> list[dict]:
```
MattWellie, line 186: As mentioned previously, the SGs from metamist process has been simplified from 3 calls to 1. Now this mock is parametrised to return a specific list of dictionaries depending on which cohort ID was given as an argument

test/test_multicohort.py
```diff
@@ -241,7 +232,8 @@ def test_multicohort(mocker: MockFixture, tmp_path):
 
     mocker.patch('cpg_workflows.metamist.Metamist.get_ped_entries', mock_get_pedigree)
     mocker.patch('cpg_workflows.metamist.Metamist.get_analyses_by_sgid', mock_get_analysis_by_sgs)
-    mocker.patch('cpg_workflows.metamist.Metamist.get_sgs_for_cohorts', mock_get_cohorts)
+    # don't patch the method location, patch where it's imported/called
+    mocker.patch('cpg_workflows.inputs.get_cohort_sgs', mock_get_cohorts)
```
MattWellie, line 236: Because the method is now static, we don't mock it in the context of the Metamist instance, we mock it where it's imported

MattWellie (commented):
Dumping a load of explanatory comments all over this change

cpg_workflows/workflow.py
vivbak: Funnily, we made this call in a separate PR around the same time.

cpg_workflows/workflow.py
vivbak: Funnily, we made this call in a separate PR around the same time.

cpg_workflows/inputs.py
```diff
@@ -251,88 +236,83 @@ def _populate_alignment_inputs(
...
+            crai_path = analysis.output.with_suffix('.cram.crai')
+            if not exists(crai_path):
+                crai_path = None
+            sequencing_group.cram = CramPath(analysis.output, crai_path)
+
+        elif exists(sequencing_group.make_cram_path()):
+            logging.warning(
+                f'We found a cram file in the expected location {sequencing_group.make_cram_path()},'
+                'but it is not logged in metamist. Skipping. You may want to update the metadata and try again. ',
+            )
```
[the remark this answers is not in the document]
Alexander-Stuckey: If I'm not wrong here, if you had multiple files that didn't exist, this code will fail on the first one, and would take you multiple runs to identify all of them. Would it be sensible to change it to make a log of all files that don't exist, and then error at the end if it finds any, and print all the error messages then?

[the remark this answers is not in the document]
MattWellie: Quick disclaimer, this code block is moved but unchanged, so I probably wouldn't alter it in this PR

So yes but also maybe no - The code only hard fails if there's a metamist entry pointing to a non-existent file. We could refactor this to collect all the failures, that would be a fairly simple change. 

The Warning messages are for an unregistered file that DOES exist, and the fallback is leaving these attributes as None, which... is accurate, and it's down to the Stage whether that's important (tolerate or throw an error)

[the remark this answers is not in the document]
Alexander-Stuckey: Yeah the warnings I am fine with, was just wondering about the failures. But for sure we could leave this bit of refactoring to after the migration.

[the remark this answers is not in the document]
MattWellie: This block actually needs a refactor anyway, given that since the first-class files change ([here](https://github.com/populationgenomics/metamist/pull/856)) the analysis entries in Metamist do still have `X.output`, but we also have a more full-featured `X.outputs.path` and a bunch of derivative attributes like file size and integrity hashes. We should probably upgrade to use that.

cpg_workflows/targets.py
```diff
@@ -121,14 +141,30 @@ def target_id(self) -> str:
...
+
     def get_cohorts(self, only_active: bool = True) -> list['Cohort']:
         """
         Gets list of all cohorts.
         Include only "active" cohorts (unless only_active is False)
         """
         cohorts = list(self._cohorts_by_name.values())
         if only_active:
-            cohorts = [c for c in cohorts if c.active and c.get_datasets()]
+            cohorts = [c for c in cohorts if c.active]
```
Alexander-Stuckey, line 166: What is the difference between an `active` and `inactive` cohort? Related, should we even allow work on `inactive` cohorts?

MattWellie: So... no. The Target ABC itself includes an Active attribute - a SG can be active or inactive, but it doesn't make sense for a Dataset, and it's not ever changed from the default (True) in Cohorts. There is some talk of deprecating Cohorts, and where/if a pipeline should fail if you query for inactive cohorts

Alexander-Stuckey: Just wondering then if the check should be moved into the gql, instead of getting everything and then filtering here. Fetch and process less data that way.

MattWellie: Yup! I think that's where we'll end up going, but Cohort's don't currently have an 'active' attribute we can use in queries. I'm tempted to remove this `only_active` parameter in this method completely as it's meaningless, but if Cohorts do start having an active/inactive status, it probably shouldn't be up to `cpg-flow` to unilaterally decide that inactive cohorts should crash the pipeline

cpg_workflows/workflow.py
```diff
@@ -1317,27 +1253,6 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
...
-                f'usable (active=True) datasets found in the cohort. Check that '
-                f'`workflow.input_datasets` or `workflow.input_cohorts` is provided, and not all datasets are skipped '
-                f'via workflow.skip_datasets`',
-            )
-            return output_by_target
-        for dataset_i, dataset in enumerate(datasets):
-            action = self._get_action(dataset)
-            logging.info(f'{self.name}: #{dataset_i + 1}/{dataset} [{action.name}]')
-            output_by_target[dataset.target_id] = self._queue_jobs_with_checks(dataset, action)
-        return output_by_target
```
MattWellie, line 1339: As discussed with @vivbak - this method (and all other `deprecated_queue_for_cohort` instances) are unused in `main` and scheduled for removal already. 

Their removal here is not just housekeeping - they would have to be re-written in light of the other other changes in this PR (Cohorts don't have Datasets), so we'd need to renovate these unused methods just to prevent the code from causing failures during workflow setup. On that basis they're being bundled into the broader change.

violetbrina (commented):
@violetbrina / @nevoodoo
We've reviewed the PR. The code changes themselves we have no issues with and are happy to take it on board with the cpg-flow migration.

We have not done any functional testing, but trust that @MattWellie is happy with the functional equivalence with regard to hail batch runs on this PR and the main of production pipelines.

We have effectively jointly approved this code and will move the relevant changes over but will leave this PR unapproved on GitHub to prevent an accidental merge.

Not all files in this PR are part of the move over to cpg-flow. It's worth noting in particular changes to jobs or workflow specific stages will need to adopt these changes to work with cpg flow moving forward.

MattWellie:
Closing as outdated now that https://github.com/populationgenomics/cpg-flow/pull/8 is open

PR #960, opened 2024-10-28
Fix broken gCNV Annotation pipeline

MattWellie:
## Problem!

- We can't complete gCNV runs anymore, they're broken

## Reason!

See https://batch.hail.populationgenomics.org.au/batches/518770/jobs/358 as an example - nothing but a miserable Hail dump exception. 

@EddieLF tracked the issue to the worker batch, e.g. https://batch.hail.populationgenomics.org.au/batches/518781/jobs/4

```
Caused by: com.fasterxml.jackson.core.exc.StreamConstraintsException: 
String length (20051112) exceeds the maximum length (20000000)
```

I did some poking around and tracked the problem down to [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/seqr_loader_cnv.py#L125). In the gCNV and GATK-SV workflows we load up a GTF file from Gencode, parse it as a dictionary, then use that dict to update the gene symbols to ENSGs. I guess this dictionary grows over time, but it now contains upwards of 40k entries, each being a key: value mapping of long Strings. This exceeds the 20MB cap in Spark when creating and evaluating expressions. 

This change covers a few things:

1. We currently get the Gencode GTF file each time AnnotateCohort runs in the gCNV or GATK-SV pipelines, copy it as a local file, then throw it away. This adds a script we can run to get a specific version of this file and copy it into the common bucket. It's added as a config entry for the gCNV and GATK-SV pipelines, which should save a few mins each run.
2. When parsing this file there's an optional argument for `chunk_size`. Instead of returning a single dictionary, this gives more granular control to break that monolith up into smaller dicts
3. Using this collection of smaller dicts in the gCNV pipeline, we annotate with each fragment in turn, then write a checkpoint. This forces Hail's lazy evaluation, so the size of the expression being passed around is tiny. The MTs in general about only a few MB in size, so there's no issues with writing multiple checkpoints. I don't think GATK-SV has failed yet, so I've not made the same edits there.
4. The AnnotateCohort/Dataset script is currently run using query_command, and I really hate that, so I've written a command line ArgParse entrypoint, and this will now work as a standalone script.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,6 @@ setup(
             'mt_to_es = cpg_workflows.scripts.mt_to_es_without_dataproc:main',
             # Generate new intervals from a MatrixTable
             'new_intervals_from_mt = cpg_workflows.scripts.generate_new_intervals:cli_main',
-            # Generate new intervals from a MatrixTable
             'seqr_loader_cnv = cpg_workflows.scripts.seqr_loader_cnv:cli_main',
         ],
     },
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,8 @@ setup(
             'mt_to_es = cpg_workflows.scripts.mt_to_es_without_dataproc:main',
             # Generate new intervals from a MatrixTable
             'new_intervals_from_mt = cpg_workflows.scripts.generate_new_intervals:cli_main',
+            # Generate new intervals from a MatrixTable
+            'seqr_loader_cnv = cpg_workflows.scripts.seqr_loader_cnv:cli_main',
         ],
     },
 )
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -68,6 +68,8 @@
             'mt_to_es = cpg_workflows.scripts.mt_to_es_without_dataproc:main',
             # Generate new intervals from a MatrixTable
             'new_intervals_from_mt = cpg_workflows.scripts.generate_new_intervals:cli_main',
+            # Generate new intervals from a MatrixTable
```
EddieLF, line 71:
```suggestion
```

EddieLF (approved):
Approving! The solution to the size issue sounds really good, and the code changes look fine. I'm not deeply familiar with this pipeline, but if it's currently broken then it's only up from here right? Thanks Matt

PR #963, opened 2024-10-30
Combiner setup fixes

MattWellie:
Got to the bottom of this locally:
```
new_sg_gvcfs = [str(sg.gvcf) for sg in sg_ids if sg_ids_in_vds and sg.id not in sg_ids_in_vds]
```

The truthiness of this statement was off:
- keep SGs if `bool(sg_ids_in_vds)` is True, AND `sg.id not in sg_ids_in_vds`
- as we don't have a prior VDS, `sg_ids_in_vds` is None/empty/False, so the whole expression is False

The additional change here is to instantiate all the `list[str] | None` as an empty list - if they're None they're converted to empty lists inside new_combiner anyway, so it removes the need to first check that it's not None, then check its contents.

There was an extra bug further down:
```workflow_config['workflow'].get('genome_build', 'GRCh38'),```

was trying to get to `config[workflow][workflow]`, so I've dropped in the config helper method to do this same thing: [here](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config.py#L543)

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -3,7 +3,7 @@ from itertools import chain
 from typing import TYPE_CHECKING, Any, Tuple
 
 from cpg_utils import Path
-from cpg_utils.config import config_retrieve, get_config, image_path
+from cpg_utils.config import config_retrieve, genome_build, get_config, image_path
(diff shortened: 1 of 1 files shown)


Alexander-Stuckey (approved):
Thanks for getting to the bottom of this

PR #964, opened 2024-10-30
set batch to highmem workers slight job creation reorder

Alexander-Stuckey:
Change the combiner to use highmem workers by default. Also only create the python job just before running, instead of the start, and then running some logic.

Set a constant to please SonarLint.


MattWellie (approved):
all hail SonarLint

PR #965, opened 2024-11-01
Add ability to spawn multiple loggers

MattWellie:
Very related to https://cpg-populationanalysis.atlassian.net/browse/SET-242

We use the root logger everywhere. That's bad 
- the settings we apply (log level, format string) can get overridden by other libraries using the root logger
- we can't add separate logging granularity to the workflow setup and the pipeline Stage logic (i.e. I want to run a pipeline with debug settings, but I don't want to log every metamist call at debug level)
- we don't currently have the granularity to override the format strings. Our current log prints contain the full datetime (`YYYY-MM-DD HH:MM:SS`), which means that we cannot do a direct comparison of any two logging outputs - the seconds are guaranteed to differ between executions

This PR proposes fleshing out the `get_logger` method we already have to do a few more fun things:

... MATT TO COMPLETE THIS PR

Ideally every instance of `logging.LEVEL(message)` in prod-pipes would be changed to use a named logger, whether that's pipeline-wide, per-module, whatever.

### DIFF
--- a/cpg_workflows/utils.py
+++ b/cpg_workflows/utils.py
@@ -21,38 +21,46 @@ from hailtop.batch import ResourceFile
 from cpg_utils import Path, to_path
 from cpg_utils.config import config_retrieve, get_config
 
-LOGGER: logging.Logger | None = None
+DEFAULT_LOG_FORMAT = '%(asctime)s - %(name)s - %(pathname)s: %(lineno)d - %(levelname)s - %(message)s'
+LOGGERS: dict[str, logging.Logger] = {}
 
 
-def get_logger(logger_name: str | None = None, log_level: int = logging.INFO) -> logging.Logger:
+def get_logger(
+    logger_name: str = 'cpg_workflows',
+    log_level: int = logging.INFO,
+    fmt_string: str = DEFAULT_LOG_FORMAT,
+) -> logging.Logger:
     """
     creates a logger instance (so as not to use the root logger)
     Args:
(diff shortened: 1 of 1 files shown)


cpg_workflows/utils.py
```diff
@@ -15,44 +15,61 @@
...
+import coloredlogs
+
 import hail as hl
 from hailtop.batch import ResourceFile
 
 from cpg_utils import Path, to_path
 from cpg_utils.config import config_retrieve, get_config
 
-LOGGER: logging.Logger | None = None
+DEFAULT_LOG_FORMAT = '%(asctime)s - %(name)s - %(pathname)s: %(lineno)d - %(levelname)s - %(message)s'
```
Alexander-Stuckey, line 26: Do you still want `asctime` in the default logger, since that was one of the issues in the PR description?

Alexander-Stuckey (approved):

cpg_workflows/utils.py
MattWellie: I think yes - in general usage it's really valuable to determine how long different parts of the process are taking. It's a kind of soft-profiling we get for free using timestamps.

The adaptation in this change is that if I'm running something to be equivalent plain text for comparison purposes I would add a config entry `workflow.logger.NAME.format` and set that to be just "%(message)s". That would override the default on a per-logger basis.

I've just pushed a more general change so that we can override this default for all loggers with a single config entry.

PR #967, opened 2024-11-03
Implement cohort name -> ID

MattWellie:
cohort.name -> cohort.id for all workflows

For Fraser and Outrider, the argument `cohort_name` has also been updated to reflect the change

### DIFF
--- a/cpg_workflows/stages/gatk_sv/gatk_sv_common.py
+++ b/cpg_workflows/stages/gatk_sv/gatk_sv_common.py
@@ -288,8 +288,8 @@ def make_combined_ped(cohort: Cohort | MultiCohort, prefix: Path) -> Path:
 
 
 def queue_annotate_sv_jobs(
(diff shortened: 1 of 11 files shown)


cpg_workflows/stages/gatk_sv/gatk_sv_common.py
```diff
@@ -288,8 +288,8 @@ def make_combined_ped(cohort: Cohort | MultiCohort, prefix: Path) -> Path:
 
 
 def queue_annotate_sv_jobs(
-    cohort: Cohort | MultiCohort,
-    cohort_prefix: Path,
+    multicohort: MultiCohort,
+    prefix: Path,
```
MattWellie, line 292: This is now always a multicohort, the cohort argument type/name is a legacy issue. Makes sense to fix it here

vivbak (approved):

PR #968, opened 2024-11-04
Bump hail

Alexander-Stuckey:
Update hail to the latest version.
@vivbak @jmarshall  did I do this right?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.29.6
+current_version = 1.30.0
 commit = True
 tag = False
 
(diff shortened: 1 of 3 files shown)


jmarshall:
Yes, this looks fine — thanks. Usually we make this a patch-level bump rather than minor, as there are no actual prod-pipes changes. However this does affect compatibility with other packages, so there's an argument that minor-level as you have done is a better choice.

Usually we do the bump2version as an afterthought as the last commit, not the first one. So when you squash-merge this, please be sure to edit the commit message so “Update pinning to Hail 0.2.133” (with the number rather than “latest”) is the subject line and the bump2version message is the second paragraph.

Last time (PR #820) we needed to hack some of the other requirements too, but the tests succeeded on this PR so that should be unnecessary. I think it may be instructive to merge this now and then see if not regenerating the analysis runner driver image from 132 to 133 causes any trouble. If it does, we can regenerate that image pronto.

jmarshall (approved):

PR #971, opened 2024-11-06
Dont merge single VCFs

### LINKED ISSUE #970: BCFTools merge fails when asked to only merge one file
https://github.com/populationgenomics/production-pipelines/blob/a1960b567ba129b45d1350569bf2873144d60262/cpg_workflows/jobs/gcnv.py#L549

See https://batch.hail.populationgenomics.org.au/batches/529148/jobs/325

We can either do the merge anyway with `--force-single`, or adapt the command to jump straight to normalising which we might still want... There may be no impact on normalising with no merge, so maybe just skip if # VCFs == 1

MattWellie:
Closes #970 

When I translated the gCNV pipeline from Cohorts to MultiCohorts there were an UNBELIEVABLE number of issues, and unforeseen bugs.

Now the pipeline runs for each Cohort in parallel, then merges all the per-cohort joint-calls into a single massive callset prior to merging. This bug occurs when the MultiCohort contains a single Cohort - we try to `bcftools merge` a single VCF, which is not valid.

### DIFF
--- a/cpg_workflows/jobs/gcnv.py
+++ b/cpg_workflows/jobs/gcnv.py
@@ -540,7 +540,9 @@ def trim_sex_chromosomes(sgid: str, sg_vcf: str, no_xy_vcf: str, job_attrs: dict
     localised_vcf = get_batch().read_input_group(**{'vcf.gz': sg_vcf, 'vcf.gz.tbi': f'{sg_vcf}.tbi'})['vcf.gz']
     autosomes = ' '.join([f'chr{i}' for i in range(1, 23)])
     job.command('set -euo pipefail')
-    job.command(f'bcftools view {localised_vcf} {autosomes} | bgzip -c > {job.output["vcf.bgz"]}')  # type: ignore
+
+    # TODO when we adopt bcftools 1.20+ as standard we can drop the separate tabix step
+    job.command(f'bcftools view -Oz -o {job.output["vcf.bgz"]} {localised_vcf} {autosomes}')  # type: ignore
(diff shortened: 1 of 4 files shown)


Alexander-Stuckey (approved):
Looks good to me

cpg_workflows/jobs/gcnv.py
```diff
@@ -592,7 +594,7 @@ def merge_calls(sg_vcfs: list[str], docker_image: str, job_attrs: dict[str, str]
         merge_job.command(f'bcftools merge {" ".join(batch_vcfs)} -Oz -o {merge_job.tmp_vcf} --threads 4 -m all -0')
 
     # now normlise the result, splitting multiallelics
-    merge_job.command(f'bcftools norm -m -any {merge_job.tmp_vcf} | bgzip -c > {merge_job.tmp_vcf_split}')
+    merge_job.command(f'bcftools norm -m -any -Oz -o {merge_job.tmp_vcf_split} {merge_job.tmp_vcf}')
```
Alexander-Stuckey, line 597: Do you actually want `-m -any` : if SNPs and indels should be merged into a single record, specify any
or `-m -both` : if both SNPs and indels should be merged separately into two records

MattWellie: because this task is exclusively used for CNV data, I went with `any` as a catch all for splitting DUP/DEL/WT calls, wasn't really sure how these call types sit when the documentation explicitly notes SNVs and Indels :/

PR #972, opened 2024-11-06
Quick hack to enable overriding the QoB JAR commit hash

jmarshall:
Enable the combiner code (and other code that may wish to) to override the QoB JAR used. In this case, if a `workflow.jar_spec_revision` config entry exists, the combiner will use the specified JAR instead of the default vanilla 0.2.133 release QoB JAR.

We could incorporate this into `init_batch()`, but that comes from cpg-utils. Limiting the changes required to production-pipelines simplifies deployment of this quick hack.

(We will also need to generate a JAR containing the fix before we can set the config item!)

### DIFF
--- a/cpg_workflows/batch.py
+++ b/cpg_workflows/batch.py
@@ -2,6 +2,9 @@
 Extending the Hail's `Batch` class.
 """
 
+from hail.backend.service_backend import ServiceBackend
+from hail.utils.java import Env
+
 
 def make_job_name(
     name: str,
@@ -22,3 +25,20 @@ def make_job_name(
     if part:
         name += f', {part}'
     return name
+
+
+_override_revision = None
+
+
+class OverrideServiceBackend(ServiceBackend):
+    @property
+    def jar_spec(self) -> dict:
+        return {'type': 'git_revision', 'value': _override_revision}
+
+
+def override_jar_spec(revision: str) -> None:
+    global _override_revision
+    _override_revision = revision
+    backend = Env.backend()
+    if isinstance(backend, ServiceBackend):
+        backend.__class__ = OverrideServiceBackend
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -13,7 +13,9 @@ from typing import TYPE_CHECKING
 
 from hail.vds import new_combiner
 
+from cpg_utils.config import config_retrieve
 from cpg_utils.hail_batch import init_batch
+from cpg_workflows.batch import override_jar_spec
 from cpg_workflows.utils import can_reuse, to_path
 
 if TYPE_CHECKING:  # TCH002 https://docs.astral.sh/ruff/rules/typing-only-third-party-import/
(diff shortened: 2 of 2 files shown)


vivbak (approved):
Should we also include the change to the large_cohort.toml to enable this, in this PR

jmarshall:
I was guessing @Alexander-Stuckey might have a non-checked-in config file he could put it in…?

Either way, I'm going to generate the new JAR now, so we can't put it in a config file until we know the commit hash :smile:

Also it seems I screwed up the import. Thought I had run every edition of this locally…

jmarshall:
Okay, I've now built a d07bed2b40664316e6d98ffff25039794a191924 QoB JAR, which is our 0.2.133+local branch plus hail-is/hail#14750 (upstream's new proposed fix for this problem).

Once we merge this and put the following in a suitable config file:

```toml
[workflow]
jar_spec_revision = 'd07bed2b40664316e6d98ffff25039794a191924'
```

@Alexander-Stuckey will be able to retry his batch and see if we've finally broken the 20 million barrier!

(Alex, if you want to add that to _configs/defaults/large_cohort.toml_, feel free to add that as another commit on this PR.)

jmarshall:
Let's leave this for now until @Alexander-Stuckey has a chance to review it, as it affects his combiner code too.

Alexander-Stuckey:
> Let's leave this for now until @Alexander-Stuckey has a chance to review it, as it affects his combiner code too.

I'm happy to give this a try. Not sure that I can hit this problem with a smaller dataset, it needs to be of sufficient size to generate a 20MB JSON internally (certainly this never came up when I was testing the combiner code on a smaller dataset).

jmarshall:
Great — Let me just build one more JAR: vanilla 0.2.133+proposedfix, which will minimise the variables.

jmarshall:
Okay, let's add this to a suitable config:

```toml
[workflow]
jar_spec_revision = 'f5b65edb8f1ed8e431f1799546a8e96d0ddff7d9'
```

which is built from [this **fixed-133-qob-jar** branch](https://github.com/populationgenomics/hail/tree/fixed-133-qob-jar). Do you have a suitable config file, or shall I add that to _configs/defaults/large_cohort.toml_ or so in this PR?

(Also we'll need to merge this PR before you rerun it!)

Alexander-Stuckey:
> Okay, let's add this to a suitable config:
> 
> ```toml
> [workflow]
> jar_spec_revision = 'f5b65edb8f1ed8e431f1799546a8e96d0ddff7d9'
> ```
> 
> which is built from [this **fixed-133-qob-jar** branch](https://github.com/populationgenomics/hail/tree/fixed-133-qob-jar). Do you have a suitable config file, or shall I add that to _configs/defaults/large_cohort.toml_ or so in this PR?
> 
> (Also we'll need to merge this PR before you rerun it!)

I've got a config that I can use. I assume that I run with the Hail 0.2.133 container, and add this into the config?

jmarshall:
Yep. Let me merge this, and then that will be the production-pipelines commit to run from.

PR #979, opened 2024-11-07
Allow for merging only VDS, with no gvcf input

Alexander-Stuckey:
Had to change the default gvcf list into `None` as Hail was getting confused by `[]`


katiedelange (approved):
LGTM

PR #980, opened 2024-11-07
add checks to stop duplicate None entries

Alexander-Stuckey:
- Stop creating a list with multiple `None` values when there are samples that don't have a gvcf.
- Only try adding samples to the list of gvcfs to combine if `merge_only_vds = 'true'` is set in the combiner config

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -81,6 +81,7 @@ class Combiner(CohortStage):
             # Get VDS IDs first and filter out from this list
             print('toml false v python False')
             print(f'{combiner_config.get("merge_only_vds")}')
+            exit(1)
             cohort_sgs: list[SequencingGroup] = cohort.get_sequencing_groups(only_active=True)
             new_sg_gvcfs = [str(sg.gvcf) for sg in cohort_sgs if sg.gvcf is not None and sg.id not in sg_ids_in_vds]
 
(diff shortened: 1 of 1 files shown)


katiedelange (approved):
Just noting here that the direction in the code (add SGs if merge_only_vds is not true) is correct, rather than the comment in the PR (add SGs if true).

This looks functional to me, but the string comparison for Booleans is a bit icky. Is that definitely how our configuration data has to get returned/managed?

cpg_workflows/stages/large_cohort.py
```diff
@@ -68,19 +68,22 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+        new_sg_gvcfs: list[str] = []
 
         if combiner_config.get('vds_analysis_ids', None) is not None:
             for vds_id in combiner_config['vds_analysis_ids']:
                 tmp_query_res, tmp_sg_ids_in_vds = self.get_vds_ids_output(vds_id)
                 vds_paths.append(tmp_query_res)
                 sg_ids_in_vds = sg_ids_in_vds + tmp_sg_ids_in_vds
 
-        if not config_retrieve(['combiner', 'merge_only_vds'], None):
+        if combiner_config.get('merge_only_vds', False) is not True:
```
MattWellie, line 79: This is chaos, why not a Boolean? Lowercase unquoted true/false in the toml gets you a bool

MattWellie (changes requested):

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey: Yeah figured that out in testing just now, wasn't sure if that would happen

MattWellie (approved):

PR #982, opened 2024-11-08
Remove annotate cohort repartition

### LINKED ISSUE #981: Consider scrapping AnnotateDataset repartitioning
https://github.com/populationgenomics/production-pipelines/blob/8deea35191aff5c303f5bfab5f351ac2f11146c9/cpg_workflows/query_modules/seqr_loader.py#L229-L232

Low priority: Maybe we should consider removing this downsampling prior to AnnotateDataset

Original Issue: When AnnotateCohort was moved off DataProc, Hail's Query backend had issues ingesting VCFs above a certain size. We overcame this by setting a block size limit in `import_vcf` (the #MB per partition) which was pretty low. I don't remember all the reasoning here, just that I played with the number until the pipeline stopped breaking.

Next Issue: With a low block size and a high VCF size, we generated MASSIVELY fragmented datasets (115000+ partitions in our genome callsets). This did not cause problems in QoB, but became impossible to process in a local Hail Query environment (i.e. Talos). To improve usability, the block size override was removed (Hail Query has not broken), AND an extra patch was put into AnnotateDataset - when we break the huge callset into smaller per-project MTs, we do a proportional downsampling of the partitions.

In theory this is logical, with fewer partitions we gain downstream by giving Hail less busywork as there are fewer QoB jobs to keep track of. By repartitioning per-subset we're also making sure that each partition is data-dense, making good use of the spark parallelisation.

This downsampling does not break anything, but the process that extracts single-dataset MTs from the bigger MT takes hours. Contradicting Hail team comments/documentation, this is not a 'computationally free' process, it's expensive in terms of time and money 😬

This Issue: Now that we're not overriding the block sizes when we read in the 1TB+ VCF in AnnotateCohort (since https://github.com/populationgenomics/production-pipelines/pull/882), the whole-genome MT we're storing from AnnotateCohort is ~7700 partitions. That's still a lot, but way lower than the 115000 we were generating previously. Crucially, at 7700 partitions we have no issues using the data downstream (either in Talos, or when generating ES indexes), so there's no need to repartition for even more efficiency gains.

Example batch: https://batch.hail.populationgenomics.org.au/batches/529253

This is extracting 34 samples from a large callset, and has been running for 2+ hours. We haven't even created the MT to run AnnotateDataset on in that time :/ I haven't dug back into the archives to find out how long this ran for in DataProc, but that's probably not a fair comparison - DP jobs ran quickly, at great expense...

---

TL; DR - Now that AnnotateCohort creates MTs with a reasonable number of partitions, we can investigate whether the time/cost of downstream repartitioning is a waste. I don't think we have a clear test of AnnotateDataset with vs. without this repartitioning, but on the face of it this appears to be taking a massive amount of time.

MattWellie:
Closes #981 

See https://batch.hail.populationgenomics.org.au/batches/529254?q=&last_job_id=7850

7850 jobs, 3.5 hours, and $11 deep in this subsetting, almost every single job has been related to the repartitioning. This is the smallest subset cohort we have in the genome callset.

This just isn't a good use of time or money.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.30.3
+current_version = 1.30.4
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.30.3
+  VERSION: 1.30.4
 
 jobs:
   docker:
--- a/cpg_workflows/query_modules/seqr_loader.py
+++ b/cpg_workflows/query_modules/seqr_loader.py
@@ -220,39 +220,18 @@ def subset_mt_to_samples(mt_path: str, sample_ids: list[str], out_mt_path: str, 
     mt_sample_ids = set(mt.s.collect())
 
     if sample_ids_not_in_mt := unique_sids - mt_sample_ids:
-        raise Exception(
+        raise ValueError(
             f'Found {len(sample_ids_not_in_mt)}/{len(unique_sids)} IDs in the requested subset not in the callset.\n'
             f'IDs that aren\'t in the callset: {sample_ids_not_in_mt}\n'
             f'All callset sample IDs: {mt_sample_ids}',
         )
 
-    # work out the proportion of the original sample count that we'll keep
-    downsample = len(mt_sample_ids) // len(unique_sids)
-    # calculate the proportional number of partitions to generate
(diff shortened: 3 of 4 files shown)


EddieLF (approved):
LGTM, these code changes around removing the downsampling look like they only alter the process of generating the subset. The output should remain unchanged, in terms of the mt contents. The only thing this should change is the partition count of the dataset mt.

I think it's also fine to remove the `read_matrix_table` and `count_cols`/`count_rows` logging at the end there, as personally I have never needed to refer to this logging. Good to have one less read & evaluate operation in this job.

PR #984, opened 2024-11-11
RDA-3 fix the hap.py validation pipeline

MattWellie:
Latest run of this code on main:
- Mt2VCF and Hap.py: https://batch.hail.populationgenomics.org.au/batches/530048
- Final stage: https://batch.hail.populationgenomics.org.au/batches/530073

The hap.py validation pipeline doesn't _really_ need fixing, it works except for a couple of smelly bits:

- the stage which actually runs Hap.py fails when writing outputs unless it's done with `Full` permissions - Hail Batch tries to clean up temporarily staged files and fails.
- the stage which runs Hap.py has output locations as Strings, not Paths - this means the pipeline won't check their existence, so it will Always try and re-run this Stage... where it will fail again
- the final stage (parsing the Hap.py results and writing to Metamist) shouldn't be a stage, it should be a metadata decorator, like the one we use in stripy.

## Changes

- `query_modules/validation.py` (the code which extracts a single-sample VCF from a joint-call MT) is now a standalone script. This could have practical uses outside the validation pipeline, and I really hate query_command.
- `jobs.validation.validation_mt_to_vcf_job` doesn't need to exist any more, as calling the VCF generation is simple enough to put in a small Stage
- uses of `get_batch()` instead of a batch instance being passed around as an argument
- Instead of building a path to a MT based on config entries, I'm using the Talos method to find the latest dataset MT for the validation dataset
- expected outputs from ValidationHappyOnVcf are all Paths, so the workflow framework can test existence when building the workflow
- I've taken the separate Stage which parses the metadata, instead making it a metadata call attached to the previous stage


cpg_workflows/scripts/extract_single_sample_vcf_from_mt.py
```diff
@@ -0,0 +1,56 @@
...
+from argparse import ArgumentParser
+
+import hail as hl
+
+from cpg_utils.hail_batch import init_batch
+
+
+def single_sample_vcf_from_dataset_vcf(input_mt: str, sample_id: str, out_path: str) -> None:
+    """
+    takes the validation datatset VCF, filters to single sample
```
EddieLF, line 15:
```suggestion
    takes the validation datatset MatrixTable, filters to single sample
```

cpg_workflows/stages/happy_validation.py
```diff
@@ -4,75 +4,109 @@
...
 class ValidationMtToVcf(SequencingGroupStage):
     def expected_outputs(self, sequencing_group: SequencingGroup):
-        return {
-            'vcf': (
-                sequencing_group.dataset.prefix()
-                / 'validation'
-                / get_workflow().output_version
-                / f'{sequencing_group.id}.vcf.bgz'
-            ),
-            'index': (
```
EddieLF, line 26: is it ok to be removing the index output?

MattWellie: Yeah I think so. The index is still written, but the index entry here is never used. The only useful entries in this output section are things to be used in subsequent stages, and the location of the index is always implicit from the location of the VCF

EddieLF (approved):
Not extremely familiar with this pipeline, but these changes mostly just seem to be uplifting the code to use more modern CPG methods and infrastructure. Looks good as far as I can tell, thanks Matt

PR #985, opened 2024-11-12
Allow hail JAR override

MattWellie:
See [this thread](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1731047651300539) for my "what the hell is going on" write up. 

In an effort to confirm/rule out the exact Hail JAR file as a cause of our cost explosions, we'd like the ability to override the JAR file used in each of the AnnotateCohort, MT subsetting, and AnnotateDataset jobs. This is one revert-able commit that includes this change in all relevant places.

Q. will the interaction between the new (0.2.133) JAR file and the exact version of Hail installed in the cpg_workflows image have any interplay? i.e. to test that 133 is the cause, and 132 is _not_, should this build use a pinned version of Hail at 0.2.132 _and_ a JAR override to point to 132?

The test runs we already have are expensive and on 133, so we already have a baseline. The combination of things we need are:
- Hail version in the cpg_workflows image
- Hail version in the selected JAR file
- Current version of this codebase which will permit controlling the JAR file used

### DIFF
--- a/cpg_workflows/query_modules/seqr_loader.py
+++ b/cpg_workflows/query_modules/seqr_loader.py
@@ -26,6 +26,12 @@ def annotate_cohort(
     annotations.
     """
 
+    from cpg_utils.config import config_retrieve
+    from cpg_workflows.batch import override_jar_spec
+
+    if jar_spec := config_retrieve(['workflow', 'jar_spec_revision'], False):
+        override_jar_spec(jar_spec)
+
     # tune the logger correctly
     logging.getLogger().setLevel(logging.INFO)
 
@@ -203,6 +209,13 @@ def subset_mt_to_samples(mt_path: str, sample_ids: list[str], out_mt_path: str, 
         exclusion_file (str, optional): path to a file containing samples to remove from the
                                         subset prior to extracting
     """
+
+    from cpg_utils.config import config_retrieve
+    from cpg_workflows.batch import override_jar_spec
+
+    if jar_spec := config_retrieve(['workflow', 'jar_spec_revision'], False):
+        override_jar_spec(jar_spec)
+
     logging.basicConfig(level=logging.INFO)
 
     mt = hl.read_matrix_table(mt_path)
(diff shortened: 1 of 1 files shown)


Alexander-Stuckey (approved):
looks good to me

jmarshall:
The usual JAR selection mechanism is to use the JAR in the bucket corresponding to the hail Python package that is running, i.e. if you pinned 0.2.132 Hail then you would also get the 0.2.132 (namely 678e1f5…) QoB JAR. So you could run that test just by pinning.

jmarshall:
In Alex's _combiner.py_ code, the `override_jar_spec()` call occurs just after the explicit `init_batch()` call, which is the scenario for which I developed this hack. Is there an explicit `init_batch()` call that occurs before this code? It might work without, on the default batch or something, but 🤷‍♂️

jmarshall (approved):
Minor comments: not sure why the imports are inside the functions; feel free to use different config keys in the different places to be able to finely adjust.

Not sure if this hack will work without an explicit `init_batch()` call, but :+1: let's try it and see…

MattWellie:
> The usual JAR selection mechanism is to use the JAR in the bucket corresponding to the hail Python package that is running, i.e. if you pinned 0.2.132 Hail then you would also get the 0.2.132 (namely 678e1f5…) QoB JAR. So you could run that test just by pinning.

Aha, ok. That's what I thought... per my comment on that thread I _thought_ I'd set that workflow off (older version of the image back on 0.2.132), but now that I look at that run, the driver job ran on an older image, the worker job used the latest cpg_workflows. I've re-triggered that run, but correctly, which might negate the need for this whole PR

> Is there an explicit init_batch()

because this whole Stage runs using query_command (🤮) there is an init_batch call made before the code is executed as part of the transcribed code, so we'd be alright in this case. The imports didn't need to be inside the method, I was mixing up my python_jobs and query_commands...

MattWellie:
Closing as possibly irrelevant if we do get this just through pinning to an older version

PR #986, opened 2024-11-13
Manually override aneuploidies, Optimise output paths

MattWellie:
Non-functional change:

Relevant to https://github.com/populationgenomics/production-pipelines/pull/958 - where we generate a file prefix by hashing all the input filepaths, don't do this per-file, do it once per stage.

---

The functional change here is to TrimOffSexChromosomes - we have a massive issue in gCNV that GATK will fail badly if it thinks it detects an unidentified aneuploid sample in the callset (e.g. [here](https://batch.hail.populationgenomics.org.au/batches/530038/jobs/222))

As far as we can tell there are two complicating factors:
- these aren't aneuploids as far as we can tell, by either their metadata or GATK's own ploidy inference
- the error message is garbage, and identified [here](https://github.com/broadinstitute/gatk/issues/8834) points to either the wrong sample, the wrong variant, or both. As such, we can't really do targetted removal of samples from the callset without trial and error.

The solution so far has been to wait for these jobs to fail, remove all samples in the failing job from the Cohort, then re-run on the new cohort. This involves completely regenerating the whole callset at great expense, and often we just find new failures. This is painful in time and cost, and ultimately we can't recover any CNVs from these samples. 

---

This change adds the ability to use a configurable list of CPG IDs, in addition to those detected as aneuploid by GATK itself. For all aneuploidies, from config or inference, we strip off all X/Y calls - we won't have any calls on these contigs, but we should succeed across the rest of the exome.

### DIFF
--- a/configs/defaults/gcnv.toml
+++ b/configs/defaults/gcnv.toml
@@ -16,6 +16,12 @@ gncv_max_pass_events = 30
 # update this when cpg_workflows.scripts.get_gencode_gtf.sh is re-run
 gencode_gtf_file = 'gs://cpg-common-main/references/hg38/v0/gencode_47.gtf.gz'
 
+[gCNV]
+# add any CPG IDs to this list to strip their XY Calls from the final callset
+# this is a mitigation for gCNV's fiddly handling of aneuploidies
+# see https://github.com/populationgenomics/production-pipelines/pull/986
+aneuploid_samples = []
+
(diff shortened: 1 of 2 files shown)


EddieLF (approved):
Looks good to me. Only suggestion might be to add the `gCNV.aneuploid_samples` config entry to the default gcnv config toml.

PR #988, opened 2024-11-14
Only trim XY on SG IDs in this cohort

MattWellie:
Quick follow-on to https://github.com/populationgenomics/production-pipelines/pull/986

That PR allows us to add a list of CPG IDs in config which will have all X/Y calls removed. Because we run as multiCohorts, that list can contain SG IDs from a range of cohorts.

The method that uses the contents of that file now needs to limit its behaviour to only SG IDs which belong to this exact Cohort, otherwise it will try and pull in samples which aren't part of this cohort, and will fail to find input paths

### DIFF
--- a/cpg_workflows/stages/gcnv.py
+++ b/cpg_workflows/stages/gcnv.py
@@ -313,10 +313,17 @@ class TrimOffSexChromosomes(CohortStage):
         germline_calls = inputs.as_dict_by_target(GermlineCNVCalls)
         jobs = []
         for sgid, new_vcf in expected.items():
-            if sgid == 'placeholder':
+            if sgid == 'placeholder' or sgid not in cohort.get_sequencing_group_ids():
                 continue
             sg_vcf = germline_calls[sgid]['segments']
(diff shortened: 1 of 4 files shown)


cpg_workflows/stages/gcnv.py
```diff
@@ -313,10 +313,17 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
         germline_calls = inputs.as_dict_by_target(GermlineCNVCalls)
         jobs = []
         for sgid, new_vcf in expected.items():
-            if sgid == 'placeholder':
+            if sgid == 'placeholder' or sgid not in cohort.get_sequencing_group_ids():
```
EddieLF, line 316: Is the call to `cohort.get_sequencing_group_ids()` cached? If not, this could be slow as it executes for each sg in the loop. Might be better to take it out of the for loop in this case.

MattWellie: nice spot. In cpg-flow it will be a dict, but for now it's not so efficient. I've updated it!

EddieLF (approved):

PR #989, opened 2024-11-15
Force list of Cohort IDs to be unique on startup

MattWellie:
Cast the list of input cohorts as a set, forcing unique contents. This means that on startup we'll never set the same cohort up twice.

The expected type of the cohort list in `get_sgs_for_cohorts` is updated to be `set[str]`


vivbak (approved):
Thanks Matt!!

vivbak:
Actually could we also log a warning here so a user knows there's duplication in their input cohorts

PR #991, opened 2024-11-17
add restore functionality

Alexander-Stuckey:
Add the ability to load a combiner save path and restore from that save point.

### DIFF
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -27,6 +27,7 @@ def run(
     sequencing_type: str,
     tmp_prefix: str,
     genome_build: str,
+    save_path: str | None,
     gvcf_paths: list[str] | None = None,
     vds_paths: list[str] | None = None,
     specific_intervals: list[str] | None = None,
@@ -43,6 +44,10 @@ def run(
         if jar_spec := config_retrieve(['workflow', 'jar_spec_revision'], False):
             override_jar_spec(jar_spec)
 
+        # Load from save, if supplied
+        if save_path:
+            logging.info(f'Loading combiner plan from {save_path}')
+
         if specific_intervals:
             logging.info(f'Using specific intervals: {specific_intervals}')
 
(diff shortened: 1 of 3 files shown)


MattWellie:
I did not know this was a feature

Alexander-Stuckey:
> I did not know this was a feature

It is pretty cool. I need to think of a good way to store the save path (and delete the save path data if the combiner succeeds).

MattWellie:
Ok so working out what this does, it saves the intended reference/VDS/gVCF paths that are going to be used in making the new VDS? I don't know how much time/processing this saves, and I'm a little worried about this causing some weird continuation errors. Taking the save path from config leads to an extra thing an operator needs to track and change manually, which I'm not wild about.

If you want to use this I would drop the logic that picks up the plan save path from config and use:
- output_version (so the plan is not resumed if the gVCFs being used as input change)
- tmp bucket (I don't think this plan is worth keeping once the combiner exists)

Alexander-Stuckey:
Save path is used like this (automatically by Hail):

`WARNING:backend.service_backend:generated combiner save path of gs://cpg-tenk10k-main-tmp/large_cohort/b7c2217b7b689e58f5aa7b0d115367cdf97f4d_367/combiner/tenk10k-genome-2-0/combiner-plans/vds-combiner-plan_e7bdaaee7ef9c42fc5b903593614d70f3c1638a605e0b4da8315f17456c50d51_0.2.133.json
`

It's to ONLY be used in resuming / retrying runs that fail in the middle for some reason (e.g. like the QoB issue we just had) where the underlying data does not change. It should never be used otherwise.

MattWellie:
> Basically, if the input data is changed in any way, then save path should not (and I think probably can't) be used to try resume from.

Yup, that's what I mean. If you overwrite save_path from config you introduce a way of overriding the behaviour hail already has by allowing it to resume from the wrong plan. By standard the plan involves hashed contents to make sure it matches the data. If save_path is provided as a path it'll just blindly use it

MattWellie (approved):

PR #992, opened 2024-11-18
Add combiner restore functionality mw edit

MattWellie:
A thought on how I'd do this 

expected_outputs returns a dict of 2 entries - the VDS path itself, and the combiner plan path

- Combiner plan JSON path contains the `output_version` (hash made of CRAM paths), so it's explicitly linked to a set of Sequencing Groups. We'll never accidentally resume the wrong run.
- It also includes all the VDS Analysis IDs we're mixing into the plan

This makes the full plan path something unique, deterministic, and not something that can be manually overridden.

- `self.expected_outputs` is called once and saved as a very mild efficiency benefit.

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -25,15 +25,18 @@ if TYPE_CHECKING:
 HAIL_QUERY: Final = 'hail query'
 
 
-@stage(analysis_type='combiner')
+@stage(analysis_type='combiner', analysis_keys=['vds'])
 class Combiner(CohortStage):
-    def expected_outputs(self, cohort: Cohort) -> Path:
+    def expected_outputs(self, cohort: Cohort) -> dict:
         workflow_config = config_retrieve('workflow')
         combiner_config = config_retrieve('combiner')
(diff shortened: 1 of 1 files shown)


cpg_workflows/stages/large_cohort.py
```diff
@@ -25,15 +25,25 @@
...
             f"{workflow_config['cohort']}-{workflow_config['sequencing_type']}-{combiner_config['vds_version']}",
         )
-        return cohort.analysis_dataset.prefix() / 'vds' / f'{output_vds_name}.vds'
+
+        # include the list of all VDS IDs in the plan name
+        if vds_ids := config_retrieve(['combiner', 'vds_analysis_ids']):
+            ids_list_as_string = '_'.join(sorted(vds_ids))
+            combiner_plan_name = f'combiner_{ids_list_as_string}'
+        else:
+            combiner_plan_name = 'combiner'
```
Alexander-Stuckey, line 42: I'd add the cohort name in here as well (as soon as `cohort.name` is implemented)

Alexander-Stuckey (approved):
Looks pretty good, just one change for the future

PR #994, opened 2024-11-18
Patch old vds

Alexander-Stuckey:
Just assume that all reference blocks are diploid


MattWellie (approved):
Will this break anything later if we split multiallelics later and the LGT array doesn't match the number of alleles on the row?

Alexander-Stuckey:
This is just operating on the reference data for the samples in the VDS, so hopefully not. It's just assuming that all ref blocks are diploid for all samples.

PR #995, opened 2024-11-19
Patch old vds

Alexander-Stuckey:
Patch an old VDS to include LGT for reference blocks.

If a sample is female, or the ref block falls inside the PAR region, LGT is set to (0,0). Otherwise, it is set to (0).


MattWellie (approved):
Are these now multiple separate patch/upgrade scripts?

Alexander-Stuckey:
No, I keep using the same branch name, which is probably bad practice (and I have to keep fixing merge conflicts, so I really should stop)

katiedelange (approved):
LGTM

PR #996, opened 2024-11-20
Create first stages of RD combiner

MattWellie:
First two candidate stages of the RD usage of the gVCF combiner. 

Motivation: we can use the current Combiner stage of the LargeCohort pipeline, but that's it. We'd still need to write a new VDS -> Mt stage, as the Large Cohort one does an intersection/filter against known QC sites, not relevant to us. Whilst we can generate a VDS, we don't have anything existing in the pipeline to make it into a MT, so we can't engage with the MT -> VCF -> Hap.py pipeline.

This creates a new workflow, rd_combiner, which wraps the existing Combiner job, and adds a new second stage for the VDS -> MT translation. It stops there, I think that's all we need for now to test Hap.py.

We can feed a set of intervals into the gVCF -> VDS combiner. That will create a pre-partitioned dataset based on those intervals, which is pretty exciting. We would be able to drop all the get_intervals logic we currently have (fragment the genome/exome into n fragments using GATK), and instead output perfectly balanced partitioned dataset by exporting separate VCFs in parallel from each MT partition.

By default we can provide no manual intervals, and use Hail's genome defaults (2568 separate pre-defined intervals) or exome defaults, depending on the sequencing type


cpg_workflows/scripts/densify_VDS_to_MT.py
```diff
@@ -0,0 +1,112 @@
...
+
+    vds = hl.vds.read_vds(vds_in)
+
+    get_logger(__file__).info('Splitting multiallelics')
+    vds = hl.vds.split_multi(vds)
+
+    get_logger(__file__).info('Densifying data...')
+    mt = hl.vds.to_dense_mt(vds)
+
+    # remove
```
EddieLF, line 54: remove what

MattWellie: remove this damn comment

EddieLF (approved):
LGTM as a start to get some validation results from Hap.py. Thanks Matt!

PR #998, opened 2024-11-21
Pin hail in the image back to 0.2.132

MattWellie:
Based on some recent cost explosions I'd like to try setting the default version of Hail installed with the cpg_workflows package back to 0.2.132 (prior to the attempted patch for the StreamConstraintsException, which wasn't actually fixed in 133)


vivbak (approved):
Crossing everything🤞

PR #1003, opened 2024-11-21
Package ClinvArbitration results

MattWellie:
The process so far is to run this workflow for usage internally, then I localise all the component pieces, zip them up, and attach the data bundle to the [ClinvArbitration releases](https://github.com/populationgenomics/ClinvArbitration/releases)

This is a mild pain, so this change adds an extra stage on the end to pre-package this data for release as a tarball.

The added benefit is that instead of `gcloud cp`'ing the two Hail tables during Talos runs we can instead copy in a single TARchive and unpack it in the image. Makes the whole process easier.

--

I chucked two other changes onto here. 

- One is a simple correction to the Hap.py stages. The Hap.py stage expected a different output format from the previous stage

- The other is a simplification to the VDS -> MT/VCFs script to get the ball rolling on the Combiner pipeline. 

Neither are relevant to this PR, both have been tested separately, only god will judge me.


EddieLF (approved):
Looks like a good addition to the pipeline

PR #1004, opened 2024-11-22
Patch usage of VCF extractor

MattWellie:
Not sure what to say here... I used this process last week to run e2e validation, then somehow changed the interface. Not sure how or why that happened. 

Going from a query_command/pythonJob to a separate script, I enforced `--flag` arguments using argparse. This now needs to be used when the script is being called. 🤦 

Previously arguments were just passed as 3 non-keyword args. Argparse could do that, but I'd prefer to be explicit

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.31.2
+current_version = 1.31.3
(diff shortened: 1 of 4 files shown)


EddieLF (approved):
Agreed that explicit is better

PR #1006, opened 2024-11-22
Add new ClinvArbitration stage to the known workflows

MattWellie:
Without adding the last stage this isn't run by default

... also resolves yet another goof in the hap.py stages


EddieLF (approved):
last_stages = ['Approved']

PR #1007, opened 2024-11-22
Resume sample qc from checkpoints

katiedelange:
As per the example in this batch [here](https://batch.hail.populationgenomics.org.au/batches/530586), SampleQC was first pre-empted, and then on the rerun it 1) repeated the slow/costly sample QC annotation due to some extra `overwrite=True` settings, and then 2) failed as it tried to overwrite a tmp checkpoint file that already exists.

These changes remove the `overwrite=True` so that checkpointed files can be used for sample QC and sex imputation, and adds additional checks for existing checkpoint files to the interval filtering section.


MattWellie (approved):
couple of optional nits, looks good

Issue #611, opened 2024-02-06
Inputs.as_X methods are unintuitive


MattWellie:
https://github.com/populationgenomics/production-pipelines/blob/f70c393faea03f3c382439708918159a3e89e5f5/cpg_workflows/workflow.py#L160-L187

These methods are used to cycle the outputs of previous stages into the next stage as input. Almost all our Stage outputs are Paths (required for the existence checking logic to trigger), but often we need String versions of those paths when using them as input. This creates an odd situation where the `inputs.as_str()` method should be perfect, but in practice is pretty pointless.

What I think it should do: Take the input to this method, and extract a requested value as a string

What it actually does: Take the input to this method, if it's not already a `str`, throw an exception. If it is a `str`, re-cast it as a `str`, then return that.

This feels like a functionality gap in the pipeline - in MANY places we have `str(inputs.as_path(...)`, while `inputs.as_str` only has one usage which wouldn't break if we designed this properly.

It's a pretty core bit of functionality, but I would like to alter `inputs.to_str` to actually cast the relevant variable to a string, and fail if that's not possible, rather than the (silly, IMO) "check if X is a str, then cast as a str" logic we have.

vivbak:
Vivian to investigate:
Categorise under improving the developer experience.

MattWellie:
I don't really care which team owns the pipeline, I want a better solution to this problem, and I'm happy to draft it myself.

MattWellie:
See #782

MattWellie:
Closed, re-issued as https://github.com/populationgenomics/cpg-flow/issues/62

PR #1009, opened 2024-11-24
All ClinvArb. outputs as Path objects

MattWellie:
Returning all outputs from the ClinvArbitration workflow as Path objects to make the interface easier.

https://batch.hail.populationgenomics.org.au/batches/530638/jobs/1

Relevant to https://github.com/populationgenomics/production-pipelines/issues/611

I have a Stage which returned a String and a Path, and i wanted to use them downstream. I have to use as_path and as_str to pull each argument from a single stage. Very dumb. 

The inputs.as_X methods should be replaced with a generic "get all outputs from this stage" without any stupid half-baked type casting middle managing methods. They literally fail unless the variable is already the given type, so why are we re-casting variables with their current type. I'll be recreating #611 in the new cpg-flow repo.

---

Sliding in a sneaky additional change - the default gCNV config file should not actively set the aneuploidies list or exome capture path. These must be provided by the cohort-specific config.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.31.4
+current_version = 1.31.5
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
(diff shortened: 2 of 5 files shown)


EddieLF (approved):
Looks fine to me, making types more consistent in the clinvarbitration stage 👍

PR #1023, opened 2024-11-28
Handle missing data when imputing sex

katiedelange:
Samples within unknown sex following imputation are currently given a LGT of 'NA', which causes problems downstream with re-extracting reference homozygous calls. Instead, we assign samples in the F-stat 'grey zone' as male (based on threshold observations in our data), and set samples with insufficient data as female.


Alexander-Stuckey (approved):
Looks good to me

PR #1025, opened 2024-11-29
Checking Samples To Remove Are Actually in Matrix Table

michael-harper:
As per [this Slack thread](https://centrepopgen.slack.com/archives/C03FA2M1MR9/p1732846694353049?thread_ts=1732772079.666649&cid=C03FA2M1MR9).

In the original code, the number of samples to drop (`samples_to_drop`) was determined by counting the rows in `sample_to_drop_ht`. However, this count included all related samples, not just those present in the matrix table (`mt`). This led to an incorrect calculation of the remaining samples, resulting in a negative number of samples after removal.

This PR updates the PCA analysis script to ensure that only samples present in the matrix table are considered for removal.


Alexander-Stuckey (approved):
Looks good to me

PR #1027, opened 2024-11-30
RD; Combiner Style

MattWellie:
Replaces 
- #1022  
- #1024 
- #1026 

End2End test run on Validation here: 
- https://batch.hail.populationgenomics.org.au/batches/531987 (Combiner)
- https://batch.hail.populationgenomics.org.au/batches/532005 (VQSR training)
- https://batch.hail.populationgenomics.org.au/batches/532015 (previous WF plus a patch)

- Validation run: https://batch.hail.populationgenomics.org.au/batches/532098

Comparison run from scratch using seqr_loader: https://batch.hail.populationgenomics.org.au/batches/532065

This implements an e2e Combiner-using RD workflow. The new workflow needs to be run in two separate stages. First run the VDS and MT creation steps, then run everything else. One downside of Hail for workflow planning is that is needs to know an exact number of jobs in advance of starting. This workflow implementation experiments with using the Hail per-partition VCF export - this exports a whole pile of VCFs, and a manifest file which can be used to locate them all. 

The first 2 steps of the workflow create the manifest, the rest uses the manifest file for per-fragment stages (VQSR, VEP...).

---

There's a few tricky bits of logic in this PR

- `gcloud condense`: the `gcloud storage condense` command takes up to 32 separate files, concatenates into a single file, and writes the result. This is perfect for our use case of "whole genome in ordered fragments, with a separate header file". In the validation run there are 2587 separate partitions, and we do a rolling merge of these files to form a single whole-genome VCF. Due to the 30-file-per-command cap the first round creates 87 intermediates (2587 // 30). They are merged to create 3 intermediates (87 // 3), then the 3 are merged to create a single file. This file is then copied to long term storage, and indexed.
- Stacked jobs in a single VM. This logic is reused from [Exomiser](https://github.com/populationgenomics/production-pipelines/blob/aef01e14c17f1b095fe21ee6a266e152f9eab067/cpg_workflows/jobs/exomiser.py#L344). For VQSR training, the time taken to read in resource files was 95% of the runtime, so I've read in resources once, then piled a huge number of fragments into the same job. Seems pretty reasonable - the complication is that Hail requires you to namespace the outputs uniquely... If each job just wrote to `job.output`, this would be overwritten by each command. Instead we need to write to `job['something_unique']`, so we can unpack all the results accurately at the end. This looks a little messy, but is much more cost effective than a separate VM for each menial task. The parallelisation rates are all drawn from config... What works for a 6-sample validation dataset won't scale to large cohorts, but the principle works.

---

# Stages

`CreateVdsFromGvcfsWithHailCombiner`: Hail gVCF Combiner (shared code with PA). This can be used to create a new VDS from scratch (gVCFs) or combine an existing VCF (from Metamist) with some new gVCFs.

`CreateDenseMtFromVdsWithHail`: Reads the VDS, and writes it out in 3 ways:

- As a MatrixTable (this will be read back in later in AnnotateCohort, where we currently read a VCF as a MT)
- As a collection of sites-only VCF fragments, with a separate header per shard. These VCFs are exported per-MT-partition, which saves us the effort of generating genomic intervals and subsetting from a larger VCF to get fragments.
- A second sites-only VCF, but this time with a separate header file. This is for the VQSR use-case; we need a whole-genome VCF for QC training, but exporting a single VCF from thousands of fragments is not time-effective. Instead we can export in this format and rebuild a VCF using concatenation of the separate files. These files are kept in temp.

`ConcatenateVcfFragmentsWithGcloud`: Use gcloud compose to concatenate all the sites-only separate-header VCF fragments into a single VCF. This sites-only VCF is around 1GB in size, even in massive callsets, so this is quick and cheap.

`TrainVqsrIndelModelOnCombinerData`: uses the whole-genome sites-only VCF for training VQSR's Indel model

`TrainVqsrSnpModelOnCombinerData`: uses the whole-genome sites-only VCF for training VQSR's SNP model

`TrainVqsrSnpTranches`: this is where it starts to get fun. Using the SNP trained model, this applies the recalibration to each separate fragment in turn, generating per-fragment tranches. This is a deviation from the current process (one job per fragment), instead loading 100 fragments into each job. The reasoning here is that the jobs each took 99% of their runtime/cost pulling the docker image and loading the reference data, only spending 5 seconds doing work. This approach drops the costs of this stage by ~98%. 98% of not very much, but it's a saving nonetheless.

`GatherTrainedVqsrSnpTranches`: Gathers the separate results from the previous stage into a single VCF. This might... not be necessary? Revisit whether we can pass each fragment its own tranche fragment, removing the need for this step. This only needed to be a separate Stage because Hail couldn't handle tracking this many batch objects, so we couldn't run the scatter and gather using the same objects.

`RunTrainedSnpVqsrOnCombinerFragments`: Similar to TrainVqsrSnpTranches, this is a fragmented application of the trained model from before, one task per fragment. I've loaded 60 jobs into each VM for a massive cost benefit. This Stage runs the application of VQSR model and joining of all fragments into a single VCF.

`RunTrainedIndelVqsrOnCombinedVcf`: Uses the previous VCF, and applies the Indel VQSR model as well. This is done genome-wide as Indels can potentially span fragments, which means that having arbitrary boundaries can cause dodgy overlaps. The final VCF created here contains all variants, as well as all SNP and Indel VQSR annotations

`AnnotateFragmentedVcfWithVep`: Annotate each fragment of the whole genome/exome with VEP. This was done using query_command but I hate that guy.

`AnnotateCohortSmallVariants`: AnnotateCohort, without query_command

`SubsetMatrixTableToDataset`: Required between AnnotateCohort and AnnotateDataset... unless there's only one dataset in the run, in which case this step is skipped. This can save a pile of time/money for single-dataset runs. 

`AnnotateDatasetSmallVariantsWithHailQuery`: AnnotateDataset, without query_command. Takes input from either of the last two Stages, depending on whether a subset was required.


MattWellie:
Still not running smoothly, but under investigation

cpg_workflows/jobs/vcf.py
```diff
@@ -57,6 +57,38 @@ def subset_vcf(
...
+    j.image(image_path('bcftools'))
+    res = STANDARD.set_resources(j, storage_gb=storage_gb)
+
+    # declare a resource group for the concatenated output
+    j.declare_resource_group(output={'vcf.gz': '{root}.vcf.gz', 'vcf.gz.tbi': '{root}.vcf.gz.tbi'})
+    j.command(
+        f"""
+    bcftools concat --threads {res.get_nthreads() -1 } -a {" ".join(vcf["vcf.gz"] for vcf in input_vcfs)} \
+    -Oz -o {j.output['vcf.gz']}
+    tabix -p vcf {j.output['vcf.gz']}
```
Alexander-Stuckey, line 86: It's probably better future compatibility to use bcftools index (to make `csi` indexes) over tabix, unless you really want `tbi` indexes for some reason

MattWellie: mmm, fair point. I'll edit this to make both. Hail can export VCFs but only with `tbi`s, so I've just tried to keep up with that throughout.

MattWellie:
I'm putting this whole PR in the bin - it's been reimplemented as 11 discrete single-stage PRs

PR #1029, opened 2024-12-02
Dragen ica realignment pipeline

Alexander-Stuckey:
Pipeline for realigning CRAMs with Dragen in ICA.

Addition in `cpg_workflows` Dockerfile to install `icasdk`

There are a number of areas where this pipeline could be improved, but I wanted to get a working version out for review before I go on leave.

There are three empty stages at the moment: `GvcfMlrWithDragen`, `MonitorGvcfMlrWithDragen` and `CancelIcaPipeline`. The two MLR stages I am unlikely to get to before I go on leave. I may be able to work on the cancel stage before then. A cancel stage is needed, as this pipeline mostly makes API requests to Illumina's ICA platform, so hitting cancel in Hail Batch won't actually cancel any running pipeline in ICA.

There are a few places where checking for statuses could be more robust, again I just wanted something working in a PR before I go on leave.


cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+                f'gs://{sg_bucket}/{GCP_FOLDER_FOR_ICA_PREP}/{sequencing_group.name}.cram.crai_ica_file_id',
+            ),
+            'analysis_output_fid': cpg_utils.to_path(
+                f'gs://{sg_bucket}/{GCP_FOLDER_FOR_ICA_PREP}/{sequencing_group.name}_dragen_output_folder_id',
+            ),
+        }
+        return output_dict
+
+    def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> StageOutput | None:
+        cram_path_components: dict[str, str] = get_path_components_from_gcp_path(path=str(object=sequencing_group.cram))
```
MattWellie, line 75: There's a lot of use of `get_path_components_from_gcp_path`, the method is fine, but it feels like a long way round.

Instead of breaking into component parts and getting the "file" part, you could just use `sequencing_group.cram.name` for instance, which is much easier to read.

In general the Path objects have all the pieces you're looking for as instance attributes

```
cram_path_components['file'] == sequencing_group.cram.name
cram_path_components['bucket'] == sequencing_group.cram.drive
{cram_path_components["suffix"]}{cram_path_components["file"]} == sequencing_group.cram.blob
```

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+        ica_download_job.storage(storage=calculate_needed_storage(cram=cram, bucket_name=bucket_name))
+        ica_download_job.image(image=image_path('ica'))
+
+        # Get secrets and folder ID needed for runtime
+        ica_analysis_folder_id = batch_instance.read_input(ica_analysis_folder_id_path)
+
+        # Download an entire folder with ICA. Don't log projectId or API key
+        authenticate_cloud_credentials_in_job(ica_download_job)
+        ica_download_job.command('set +x')
```
MattWellie, line 324: The performance of this 12-command block would be the same if it was all done as one multiline string block like this [example](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/jobs/vqsr.py#L558). There's probably as much boilerplate as there is actual  commands so it's n t trivial to see the whole command sequence at a glance

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+        bucket_name: str = cram_path_components['bucket']
+        logging.info(bucket_name)
+
+        prepare_ica_job: PythonJob = get_batch().new_python_job(
+            name='PrepareIcaForDragenAnalysis',
+            attributes=(self.get_job_attrs() or {}) | {'tool': 'ICA'},
+        )
+        prepare_ica_job.image(image=image_path('cpg_workflows'))
+
+        prepare_ica_job.call(
```
MattWellie, line 86: As a general note having an explicit connection between the output of `expected_outputs` (generating the result paths) and the process that writes those files makes a lot of sense to me. Here you're building the paths to output files in `expected_outputs` from components, then building the same path elsewhere from components in a different way. 

The end result is that the outputs are created, but it's difficult to trace why/how they're created at the right paths due to the indirection.

MattWellie: This is probably the biggest deviation from every other stage in this pipeline framework. I'm guessing that's because the helper methods you've created (like `register_output_to_gcp`) require various parts of filename to use the google storage client, when you could use the Path object on the exact Path objects which are being created in `expected_ouputs`, e.g.

```
output_path = self.expected_outputs()

# do whatever to get the result
workflow_id = <some method call>

# open the Path for writing directly
with output_path.open('w') as handle:
    handle.write(workflow_id)
```

MattWellie (commented):
It works, and I acknowledge that being the only target at this point, but there's quite a few deviations from the way (almost?) all our other stages have been written using this framework.

Some of that feels like a little bit of reinventing the wheel, or maybe just not finding some of the obscure helper methods which can attack the same problems using pretty tight syntax

cpg_workflows/stages/realign_genotype_with_dragen.py
MattWellie: Cloudpathlib (the result of `to_path('gs://something')`, returned by most of our helper methods, and PathLib have the same interface, so this would be a decent doc for all the various path elements these Path objects will have available as properties:

https://docs.python.org/3/library/pathlib.html#methods-and-properties

cpg_workflows/stages/realign_genotype_with_dragen.py
Alexander-Stuckey: Can do that, it's much cleaner

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+    storage_client = storage.Client()
+    gcp_bucket = storage_client.bucket(bucket_name=bucket_name)
+    blob_to_upload_size_bytes: int = gcp_bucket.get_blob(blob_name=f'{cram}').size
+    storage_size: int = ceil(ceil((blob_to_upload_size_bytes / (1024**3)) + 3) * 1.2)
+    logging.info(f'Calculated storage is {storage_size}Gi.')
+    return f'{storage_size}Gi'
+
+
+# No need to register this stage in Metamist I think, just ICA prep
+@stage(analysis_keys=['cram_fid', 'cram_index_fid', 'analysis_output_fid'])
```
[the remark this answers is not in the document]
Alexander-Stuckey: Makes sense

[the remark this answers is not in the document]
Alexander-Stuckey: Ah, except that I need them as inputs for a couple of other stages, so not sure what the practice here is?

cpg_workflows/stages/dragen_ica/ica_utils.py
```diff
@@ -0,0 +1,210 @@
...
+    Raises:
+        Exception: Any exception, as we want to fail if we can't get the credentials for any reason
+
+    Returns:
+        dict[str, str]: A dictionary with the keys projectId and apiKey
+    """
+    SECRET_CLIENT = secretmanager.SecretManagerServiceClient()
+    SECRET_PROJECT = 'cpg-common'
+    SECRET_NAME = 'illumina_cpg_workbench_api'
+    SECRET_VERSION = 'latest'
```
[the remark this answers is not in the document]
Alexander-Stuckey: Done

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+GCP_FOLDER_FOR_ICA_PREP: Final = f'ica/{DRAGEN_VERSION}/prepare'
+GCP_FOLDER_FOR_RUNNING_PIPELINE: Final = f'ica/{DRAGEN_VERSION}/pipelines'
+GCP_FOLDER_FOR_ICA_DOWNLOAD: Final = f'ica/{DRAGEN_VERSION}/output'
+ICA_REST_ENDPOINT: Final = 'https://ica.illumina.com/ica/rest'
+
+
+coloredlogs.install(level=logging.INFO)
+
+
+def calculate_needed_storage(
```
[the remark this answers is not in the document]
Alexander-Stuckey: Neat, that's much nicer

Dockerfile
```diff
@@ -6,4 +6,6 @@ COPY setup.py .
 COPY cpg_workflows cpg_workflows
 COPY seqr-loading-pipelines/hail_scripts hail_scripts
 COPY gnomad_methods/gnomad gnomad
-RUN pip install .
+RUN pip install . \
+ && pip install git+https://github.com/Illumina/ica-sdk-python.git \
+ && pip install typing-extensions --upgrade
```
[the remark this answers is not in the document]
Alexander-Stuckey: Moved `icasdk` install to beginning of Dockerfile

configs/defaults/realign_genotype_with_dragen.toml
```diff
@@ -0,0 +1,35 @@
...
+[ica.reference_ids]
+# Dragen hash table for 3.7.8
+dragen_ht_id = 'fil.854d49a151a24edae5d708da2935b1b0'
+
+# CRAM reference used to align the CRAM file that you want to realign with Dragen
+# Dragmap reference FASTA
+cram_reference_id = 'fil.159e94790e5c4bdfcbef08dd148ff548'
+
+# GATK reference FASTA
+# cram_reference_id = 'fil.7e16c32f65e9468d31b908dcefab1cbc'
```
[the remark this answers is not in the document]
Alexander-Stuckey: Can do

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,357 @@
...
+            'cram_fid': cpg_utils.to_path(
+                f'gs://{sg_bucket}/{GCP_FOLDER_FOR_ICA_PREP}/{sequencing_group.name}.cram_ica_file_id',
+            ),
+            'crai_fid': cpg_utils.to_path(
+                f'gs://{sg_bucket}/{GCP_FOLDER_FOR_ICA_PREP}/{sequencing_group.name}.cram.crai_ica_file_id',
+            ),
+            'analysis_output_fid': cpg_utils.to_path(
+                f'gs://{sg_bucket}/{GCP_FOLDER_FOR_ICA_PREP}/{sequencing_group.name}_dragen_output_folder_id',
+            ),
+        }
```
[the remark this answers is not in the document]
Alexander-Stuckey: I'll give this a try, I think it was having some trouble with `-test` earlier, which is why I had swapped to the current syntax. However, that problem seems to have reoccurred when using `bioheart-test` instead of `fewgenomes-test` so 🤷

cpg_workflows/stages/realign_genotype_with_dragen.py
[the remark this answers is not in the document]
MattWellie: yeah no worries that's cool. Answer iss..... depends how you want to access them. The analysis_keys in the `@stage` are only relevant for the metamist registration. If you want to pull from Metamist later, they have to be in there.

If you want to pull them directly from the previous stage as inputs the `inputs.as_dict(self, Stage_Name)` (or whatever the syntax is) will always get you the full result of the expected_outputs method, the stage decorator isn't involved

MattWellie:
Just cross-posting this PR: https://github.com/populationgenomics/production-pipelines/pull/1068. The fix adopted in this PR was the one from #1063. #1068 is a more succinct fix for the `-test` suffix/`dataset.prefix()` issue, which leaves the metamist.py script unchanged.

I've made [this](https://github.com/populationgenomics/production-pipelines/pull/1069) stacked on top of this branch to update to the better version of the fix, and check that the fix works as intended before adopting.

MattWellie (commented):
Looks good! There's a stacked PR to prevent this merge to main circumventing #1068, but this looks like a ruddy good start

cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
```diff
@@ -0,0 +1,50 @@
...
+            time.sleep(600 + randint(-60, 60))
+            pipeline_status = ica_utils.check_ica_pipeline_status(
+                api_instance=api_instance,
+                path_params=path_params | {'analysisId': ica_pipeline_id},
+            )
+        if pipeline_status == 'SUCCEEDED':
+            logging.info(f'Pipeline run {ica_pipeline_id} has succeeded')
+            return {'pipeline': 'success'}
+        else:
+            raise Exception
```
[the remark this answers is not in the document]
Alexander-Stuckey: Yeah for sure. This was again just something to get working in case I had to disappear on leave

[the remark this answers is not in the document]
Alexander-Stuckey: Cleaned this up now, it raises an explicit exception if you cancel the pipeline (in ICA or otherwise), and otherwise informs you to check ICA for any other reason that the pipeline failed.

cpg_workflows/stages/realign_genotype_with_dragen.py
Alexander-Stuckey: I believe I've resolved all of these, except where strictly necessary

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,329 @@
...
+        output_fids = prepare_ica_job.call(
+            prepare_ica_for_analysis.run,
+            ica_analysis_output_folder=config_retrieve(['ica', 'data_prep', 'output_folder']),
+            api_root=ICA_REST_ENDPOINT,
+            sg_name=sequencing_group.name,
+            bucket_name=bucket_name,
+        ).as_json()
+
+        get_batch().write_output(
+            output_fids,
```
MattWellie, line 87: just out of curiosity is this result being kept as a `{key: result_ID}` dictionary because it needs to be passed that way in future? If the json'ification isn't important I would have assumed just returning/storing the ID in isolation would be easier (not needing JQ/other to read back later)

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,329 @@
...
+            {ICA_CLI_SETUP}
+            cram_status=$(icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram --match-mode EXACT -o json | jq -r '.items[].id.details.status')
+            if [[ $cram_status != "AVAILABLE" ]]
+            then
+                mkdir {sequencing_group.name}
+                gcloud storage cp {sequencing_group.cram} .
+                gcloud storage cp {sequencing_group.cram}.crai .
+                icav2 projectdata upload {sequencing_group.name}.cram /{bucket}/{upload_folder}/{sequencing_group.name}/
+                icav2 projectdata upload {sequencing_group.name}.cram.crai /{bucket}/{upload_folder}/{sequencing_group.name}/
+            fi
```
MattWellie, line 128:
```suggestion
        cram_group = get_batch().read_input_group(
            cram=str(sequencing_group.cram.path), 
            crai=str(sequencing_group.cram.index_path),
        )
        
        # Check if the CRAM already exists in ICA before uploading. If it exists, just return the ID for the CRAM and CRAI
        upload_job.command(
            f"""
            {ICA_CLI_SETUP}
            cram_status=$(icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram --match-mode EXACT -o json | jq -r '.items[].id.details.status')
            if [[ $cram_status != "AVAILABLE" ]]
            then
                icav2 projectdata upload {cram_group.cram} /{bucket}/{upload_folder}/{sequencing_group.name}/
                icav2 projectdata upload {cram_group.crai} /{bucket}/{upload_folder}/{sequencing_group.name}/
            fi
```

just a suggestion to delegate localisation to hail batch, which would make it easier to drop the gcloud cli in this job image later

MattWellie: This diff is a bit messy given that it includes the unchanged lines in between. It's the read_input_group to create the local files from cram/crai, and dropping the two gcloud cp lines

cpg_workflows/stages/dragen_ica/run_align_genotype_with_dragen.py
```diff
@@ -0,0 +1,125 @@
...
+    analysis_output_fid_path: str,
+    dragen_ht_id: str,
+    cram_reference_id: str,
+    dragen_pipeline_id: str,
+    user_tags: list[str],
+    technical_tags: list[str],
+    reference_tags: list[str],
+    user_reference: str,
+    api_root: str,
+) -> dict[str, str]:
```
MattWellie, line 91: If I'm being a picky bugger, seeing as so many of these methods are all called `run`, a docstring explanation of what this `run` runs would be cool. In the job call it's clearer as `run_align_genotype_with_dragen.run` but in this review I keep bouncing up to the file name to know what the method does

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -0,0 +1,329 @@
...
+        ica_download_job.storage(storage=calculate_needed_storage(cram=str(sequencing_group.cram)))
+        ica_download_job.memory('8Gi')
+        ica_download_job.image(image=image_path('ica'))
+
+        # Download an entire folder with ICA. Don't log projectId or API key
+        authenticate_cloud_credentials_in_job(ica_download_job)
+        ica_download_job.command(
+            f"""
+            {ICA_CLI_SETUP}
+            icav2 projectdata download $(cat {ica_analysis_folder_id_path} | jq -r .analysis_output_fid) {sequencing_group.name} --exclude-source-path
```
MattWellie, line 320: Aha, this the corresponding reading in of that JSON output. Seems fine, not sure if `$(cat analysis_id)` would be less work, but you might need to worry about line endings. 

Basically I've never gotten used to JQ, it scares me.

MattWellie (approved):
LGTMatt

cpg_workflows/stages/realign_genotype_with_dragen.py
Alexander-Stuckey: Yeah that'd be a better solution. Earlier this method had multiple key: value pairs, so json did make sense, but that changed once I had to redo the upload method. Will change

Alexander-Stuckey: I will look into changing this (as that would be better) in a new branch, because it'd be good to get something working merged in first.

cpg_workflows/stages/realign_genotype_with_dragen.py
Alexander-Stuckey: So I know that it doesn't take a huge amount of time to load data from GCP to the VM, but the current solution only does that if the CRAM is not already present in ICA, instead of always. Maybe I'm solving for a problem that won't really come up, but 🤷

cpg_workflows/stages/dragen_ica/run_align_genotype_with_dragen.py
Alexander-Stuckey: Yeah can do this

jmarshall:
I wish you would have announced that you were planning to merge this imminently and put out a final call for review, either here on the PR or on slack. You have asked a number of people to review this, many of whom have may have done so if it had been clear that you now considered this ready for merging.

I have two comments, but as they are somewhat process-related I had been leaving them until I had time to do a more useful review of the PR's actual content as well.

1. Because everything is imported in _main.py_, in this unified codebase the use of `icasdk` will lead to trouble for other workflows and local dry-running:

```
File "/Users/johnm/production-pipelines/cpg_workflows/stages/dragen_ica/cancel_ica_pipeline_run.py", line 6, in <module>
    import icasdk
ModuleNotFoundError: No module named 'icasdk'
```

This may be annoying to work around in your source files, but it's probably not reasonable to expect all driver images and local laptops to install this SDK. Perhaps you could just do something like `try: import icasdk; except ImportError: pass # can't actually use this workflow then`.

2. Merging this PR has applied a variant of the `-test` fixes to _cpg_workflows/inputs.py_. This work should have been left to PR #1068 and this PR rebased on top of that / merged onto from **main** once the full fix for that had been merged to **main**.

Alexander-Stuckey:
> I wish you would have announced that you were planning to merge this imminently and put out a final call for review, either here on the PR or on slack. You have asked a number of people to review this, many of whom have may have done so if it had been clear that you now considered this ready for merging.

This is fair, yet I will point out that this PR was marked ready for review last week, and only had reviews from one person in that time. Given that there had been no other apparent interest in that time, I had made the assumption that nobody else was wanting to review it.

Next time I will tag people for a final round of review before merging.


> I have two comments, but as they are somewhat process-related I had been leaving them until I had time to do a more useful review of the PR's actual content as well.
> 
> 1. Because everything is imported in _main.py_, in this unified codebase the use of `icasdk` will lead to trouble for other workflows and local dry-running:
> 
> ```
> File "/Users/johnm/production-pipelines/cpg_workflows/stages/dragen_ica/cancel_ica_pipeline_run.py", line 6, in <module>
>     import icasdk
> ModuleNotFoundError: No module named 'icasdk'
> ```

This is a very good solution, I will make a quick PR to ad this in so that this no longer breaks for others that have different workflows / images.

> This may be annoying to work around in your source files, but it's probably not reasonable to expect all driver images and local laptops to install this SDK. Perhaps you could just do something like `try: import icasdk; except ImportError: pass # can't actually use this workflow then`.



> 2. Merging this PR has applied a variant of the `-test` fixes to _cpg_workflows/inputs.py_. This work should have been left to PR [Removing test suffix from dataset names for multicohorts. #1068](https://github.com/populationgenomics/production-pipelines/pull/1068) and this PR rebased on top of that / merged onto from **main** once the full fix for that had been merged to **main**.

I apologise for this one, I'd used it for testing and forgotten to roll it back before merging, like I had for some other changes that were only in for testing. Would you like me to remove this, so that [Removing test suffix from dataset names for multicohorts. #1068](https://github.com/populationgenomics/production-pipelines/pull/1068) can still proceed properly?

jmarshall:
Was adding `icasdk` to the driver image also part of just stopping this import from breaking for people running other workflows, or was it intentional and significant to add this SDK to the default driver image?

If the former, reverting this change to _Dockerfile_ would also be appreciated.

PR #1030, opened 2024-12-03
RD Combiner: 1

MattWellie:
https://docs.google.com/document/d/18RHLlZmWjjv9OZNHAqpuAjf0w0y9NEb50Zjwdfiow2I/edit?usp=sharing

First two stages polished - Combine gVCFs and Densify to MT representation.

This leaves the original RD combiner stage unchanged (already in `main`, but it does rename it to conform to the more complete new naming convention)

It substantially alters the second step, densifying the VDS to a MT, and several sites-only VCFs.

The initial implementation of this Stage ([here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/scripts/densify_VDS_to_MT.py)) did just enough to export a sites-only VCF for validation purposes, but it didn't do enough to use that data downstream, e.g. in VQSR. [This](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/large_cohort/site_only_vcf.py) was a better model (the Large_cohort pipeline's sites-only VCF from VDS module). This generates all the INFO fields which VQSR will rely on, emulating the GATK Variant Caller. There was a bit of complexity here because the gnomad methods are designed to run on the Sparse representation, so to go from VDS to a MT with Split multiallelics we have to:

- Read the VDS
- Densify the VDS to a single MT (entry fields are still the sparse representaton at this stage, e.g. `LPL` instead of `PL`)
- Compute all the INFO fields based on the sparse entry fields using the gnomad methods
- Annotate this info content back into the MT
- Run the sparse_split_multi method, which will both split the multiallelic fields into separate rows, and recompute all the INFO fields to be specific to the individual alt allele on each row. 

At this point we have all the info fields populated, the alleles split, and the entry fields as an accurate MT representation. This is written to file as the main output of the Stage. This MT will be read in later during AnnotateCohort (where seqr_loader writes a joint-called whole-genome VCF, then reads it into a MT, we'll just read this MT)

We then also create two separate clusters of sites-only VCF files. By using Hail's export-vcf-per-shard we get the benefit of writing the MT data out as a VCF, without the Multi-hour export times that come from sequentially writing the thousands of separate partitions to a single file result. This is the Hail recommended route for generating a VCF from a MT. This behaviour replaces all the get_intervals usage in our current pipeline, instead using Hail's per-partition fragments to generate an almost perfectly balanced set of intervals:

1. `parallel='header_per_shard'`: Hail exports the MatrixTable as a collection of sites-only VCFs, one per partition, each with its own header. These are fully standalone VCFs, and we can feed them into VQSR training and VEP annotation.
2. `parallel='separate_header'`: Hail exports the MT as a collection of sites-only VCFs, with a separate header file. We can concatenate all these files later into a whole-genome VCF without having to localise all the fragments for a `bcftools concat`.

In both cases the VCF fragments are in genomic coordinate order, and Hail writes a manifest file which contains the names of all the ordered fragments.

--- 

These 2 stages together represent a stopping point for this workflow. Hail Batch needs to know the exact number of jobs in the workflow prior to starting. In the seqr_loader pipeline we started out with the number of intervals to split the data up in, and applied those intervals throughout the workflow: VEP, VQSR, Joint-Calling. Because we knew the number of intervals in advance, the whole workflow could be planned continuously.

This implementation of the per-partition export uses Hail's default partitioning intervals, based on exome or genome depending on sequencing_type. We can also overwrite the partitions, or repartition the finished dataset if we want to change this later. For reasons that will make sense in the next couple of Stages, it's important that all the sites-only VCFs we have are written, with a corresponding manifest file, for planning the next Stages of the workflow.


MattWellie:
I've rebased this branch to absorb the cohort.name vs. cohort.id changes. Shouldn't need to update each other stacked branch separately

cassimons (approved):

PR #1031, opened 2024-12-03
RD Combiner: 2 - gcloud manifest condenser

MattWellie:
This PR adds the next step of the workflow - using `gcloud storage compose` to take the per-partition VCF exports and rebuild them into a single continuous whole-genome VCF. We need a whole-genome sites-only VCF for VQSR training, but as it's sites-only the total data footprint is pretty small. Because we only need this for VQSR training, the fragments and intermediates are all kept in temp, with only the final file being written to main storage.

Because Hail writes out data-per-partition (fast, highly parallel) in a guaranteed ordered way, we can safely concatenate all the fragments in the order they sit in the manifest file.

This approach makes best use of the per-partition export from MT (rapid) and the GCloud condense (combine resulting partitions without ever having to localise files)

---

This step needs to run after the prior stage has completed - see the config entry:

`last_stages = ['CreateDenseMtFromVdsWithHail']`

To know how many fragments we have to combine, the manifest file needs to exist. There may be ways to screw with Hail Batch job scheduling to fit this into one continuous workflow, but the simplest approach is to generate the MT and fragments, then run the rest of the workflow continuously. 

The first operation of this Stage is checking that the manifest file already exists on file - if it doesn't exist yet, a message will tell you to run up to the previous Stage.


cpg_workflows/jobs/gcloud_composer.py
```diff
@@ -0,0 +1,153 @@
...
+            'vcf.bgz.tbi': '{root}.vcf.bgz.tbi',
+            'vcf.bgz.csi': '{root}.vcf.bgz.csi',
+        },
+    )
+    final_job.image(image_path('bcftools'))
+    final_job.storage(config_retrieve(['gcloud_condense', 'storage'], '10Gi'))
+
+    # we specifically need block-gzipped output for some downstream tools
+    # otherwise a mv or a bcftools view with --write-index=tbi would be neater
+    final_job.command(f'bcftools view {input_vcf} | bgzip -c > {final_job.output["vcf.bgz"]}')
```
MattWellie, line 146: can swap this over to a mv, _should_ still be block-gzipped

cpg_workflows/stages/rd_combiner.py
```diff
@@ -163,3 +164,44 @@ def queue_jobs(self, multicohort: MultiCohort, inputs: StageInput) -> StageOutpu
...
+@stage(analysis_keys=['vcf'], analysis_type='vcf', required_stages=[CreateDenseMtFromVdsWithHail])
+class ConcatenateVcfFragmentsWithGcloud(MultiCohortStage):
+    """
+    Takes a manifest of VCF fragments, and produces a single VCF file
+    This is disconnected from the previous stage, but requires it to be run first
+    So we check for the exact same output, and fail if we're not ready to start
+    """
+
+    def expected_outputs(self, multicohort: MultiCohort) -> dict[str, Path]:
+        return {'vcf': self.prefix / 'gcloud_composed_sitesonly.vcf.bgz'}
```
MattWellie, line 178: probably should be in tmp

cassimons (approved):
Combining is cool, we should do more of that.  The [bgzip docs](https://www.htslib.org/doc/bgzip.html#BGZF_FORMAT) support the direct concatincation generates a valid bgz file.

PR #1033, opened 2024-12-03
RD Combiner: 4

MattWellie:
Adds the next Stage - a scattered per-fragment training of VQSR SNP Tranches, using the genome-wide model generated in the previous stage.

During testing I found that the time taken to read the necessary resources into each of these single-fragment training jobs was ~90% of the runtime. As a result I've re-worked the existing logic to break up the separate fragments into chunks (configurable), and create a single VM per chunk. Within that VM I run a large number of fragment training jobs, to optimise the cost associated with the reference data localisation.

I was playing with a very small dataset, so the required storage and chunk size are all configurable in case we run into issues with larger datasets.

This is the first use of the VCF fragments! How exciting! I've used a cached method to read all the fragments into the batch. That prevents any unnecessary re-localisation (e.g. once per fragment for VEP, a separate localisation for VQSR..)


cpg_workflows/jobs/rd_combiner/vqsr.py
```diff
@@ -176,3 +177,136 @@ def train_vqsr_snps(sites_only_vcf: str, snp_model: str, job_attrs: dict):
...
+    # to hold the resulting parts, as Resources
+    snp_tranche_fragments: list[ResourceFile] = []
+
+    # if we start this as -1, we can increment at the start of the loop, making the index counting easier to track
+    vcf_counter = -1
+
+    # iterate over all fragments, but in chunks of FRAGMENTS_PER_JOB
+    for chunk_counter, fragment_chunk in enumerate(chunks(vcf_resources, TRAINING_PER_JOB)):
+        # NB this VQSR training stage is scattered, and we have a high number of very small VCF fragments
+        # 99.5% of the time and cost of this task was pulling the docker image and loading reference data
```
cassimons, line 224: Are these time estimates from the tiny test cohort, or did you look at the equivalent jobs from the previous pipeline? It will be interesting to see how this scales.

cassimons (approved):

cpg_workflows/jobs/rd_combiner/vqsr.py
MattWellie: These were the timings from the test cohort. From the genome runs in `main` the ratio was a bit different... maybe 2 mins reading references to 40 seconds running the process.

MattWellie: In seqr-loader (lower scatter count at ~1500) the balance was ~2 mins to run the input step, ~40 seconds of work
https://batch.hail.populationgenomics.org.au/batches/530803/jobs/6177. 140GB storage provisioned but only 20GB used, almost entirely references

PR #1035, opened 2024-12-03
RD Combiner: 6

MattWellie:
Almost there - this adds the next VQSR step

- Runs VQSR on each of the VCF fragments using the global SNP model, and the per-fragment trained tranches.
- Combines the resulting VCFs into one whole-genome VCF, including the VQSR INFO data

Again, this operates in chunks, with multiple fragments being processed in each chunk, and the per-fragment VCFs being processed into a smaller number of intermediate files. The final operation is to `bcftools concat` the intermediates into one whole-genome VCF file.


cpg_workflows/jobs/vcf.py
```diff
@@ -9,11 +9,48 @@
...
+    concat_job.declare_resource_group(
+        output={
+            'vcf.gz': '{root}.vcf.gz',
+            'vcf.gz.tbi': '{root}.vcf.gz.tbi',
+            'vcf.gz.csi': '{root}.vcf.gz.csi',
+        },
+    )
+    concat_job.command(
+        f"""
+    bcftools concat --threads {res.get_nthreads() -1 } -a {" ".join(vcf["vcf.gz"] for vcf in input_vcfs)} \
```
cassimons, line 45: Can you use `-native` here or are the headers not helpful?

cassimons (approved):

cpg_workflows/jobs/vcf.py
MattWellie: I've not come across that flag, what does `-native` do? I couldn't see it in the [manual](https://samtools.github.io/bcftools/bcftools.html)

PR #1037, opened 2024-12-03
RD Combiner: 8

MattWellie:
Adds fragmented VEP annotation, building into a single Hail Table

This localises the per-fragment VCFs, and runs VEP annotation on each one.

This simplifies some of the existing VEP logic - it only runs on GRCh38, on VEP 110, and outputs JSON. Because of this it can drop all conditional logic for running on VEP 105 vs 110, etc.


MattWellie:
This is a definite candidate for running `gcloud compose` on the individual JSONL files, instead of expecting Hail Batch to aggregate 2500 separate JSON files when building the unified HT. These files are all a single JSON variant annotation per row, 1:1 with the variant MT, so we shouldn't have issues concatenating them all together into one fat JSON file

cassimons (approved):

PR #1050, opened 2024-12-05
Add the ability to update the storage and memory requirements via config

katiedelange:
MakeSiteOnlyVcf requires more storage, as per: https://batch.hail.populationgenomics.org.au/batches/532410/jobs/124630


silkm (approved):
Looks good!

PR #1055, opened 2024-12-08
Add non-preemptible vm option for es-index creating job

EddieLF:
Allow the elasticsearch index export job to run on non-preemptible (or "not spot") VMs, to stop them getting preempted and cancelling the index. Should only be used for when time is of the essence and multiple days worth of preemptions is not worth the savings.

Non-preemtible VM cost to export a 22GB index over 1hr 25m: [$0.4062](https://batch.hail.populationgenomics.org.au/batches/532710/jobs/163) - according to Hail Batch

Preemptible VM cost in the same time: [~$0.20](https://batch.hail.populationgenomics.org.au/batches/532774/jobs/1) (it got preempted 😆  so this is a guesstimate)

### DIFF
--- a/cpg_workflows/stages/seqr_loader.py
+++ b/cpg_workflows/stages/seqr_loader.py
@@ -226,7 +226,7 @@ class MtToEs(DatasetStage):
 
         job = get_batch().new_bash_job(f'Generate {index_name} from {mt_path}')
         if not config_retrieve(['workflow', 'es_index', 'spot_instance'], default=True):
-            job = job.spot(is_spot = False)
+            job = job.spot(is_spot=False)
 
         req_storage = tshirt_mt_sizing(
             sequencing_type=config_retrieve(['workflow', 'sequencing_type']),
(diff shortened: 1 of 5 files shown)


cpg_workflows/stages/seqr_loader.py
```diff
@@ -225,6 +225,8 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
         flag_name = str(outputs['done_flag'])
 
         job = get_batch().new_bash_job(f'Generate {index_name} from {mt_path}')
+        if not config_retrieve(['workflow', 'es_index', 'spot_instance'], default=True):
+            job = job.spot(is_spot=False)
```
MattWellie, line 229:
```suggestion
        job = job.spot(is_spot=config_retrieve(['workflow', 'es_index', 'spot_instance'], default=True))
```

This is a nit... current code is fine

MattWellie (approved):
LGTMatt

cpg_workflows/stages/seqr_loader.py
EddieLF: Fair to comment on, something was bugging me about the way it was 
`if not x, then y = false` - its just a bit confusing at a glance, IMO

I've changed it slightly, so it's now 
`if x is false, then y = false` 
And added an inline comment just so it's extra clear

PR #1057, opened 2024-12-09
Implement init_batch_args in Frequencies Stage for Better Resource Management

michael-harper:
Another [driver job](https://batch.hail.populationgenomics.org.au/batches/532713/jobs/9) is running out of memory, this time when running the `Frequencies` stage of large cohorts (parent batch [here](https://batch.hail.populationgenomics.org.au/batches/532712)). 

This PR implements the `init_batch_args` within the Frequencies stage to allow more granular control over the worker and driver jobs independently so as to avoid further crashes during larger runs.


katiedelange (approved):
Looks great, thanks Harper!

PR #1063, opened 2024-12-13
Strip -test from dataset name 

### LINKED ISSUE #1047: SG.dataset.prefix() doesn't work with access-level == test
https://batch.hail.populationgenomics.org.au/batches/532274/jobs/1

The config file generated in test runs doesn't have a storage.dataset-test section, but the SGs are being populated with a dataset called "dataset-test", e.g. fewgenomes-test

vivbak:
Fixes #1047 

Currently, users are unable to run workflows in test, hitting the following error 

`cpg_utils.config.ConfigError: Storage section for dataset "project-test" not found in config. Please check that you have permissions to the dataset. Expected section: [storage.project-test]`

It seems as though, the dataset.prefix() method is returning a dataset name with -test appended at the end. This conflicts with the current naming convention of the config file storage section. 

```
[storage.project]
analysis = "gs://cpg-project-test-analysis"
```
rather than 
```
[storage.project-test]
analysis = "gs://cpg-project-test-analysis"
```

An alternative solution would be to modify how we generate the config, which I personally think is a better approach. 

I am suggesting this approach because I am unfamiliar with modifying the infrastructure this way, this is urgent, and I'm going on leave for the year.


vivbak:
@milo-hyben has been briefed

milo-hyben (approved):
As Vivian mentioned, this is a quick and dirty solution.
I had look at the cpg-infra implementation of the config and it is slightly more complicated, e.g. all depending datasets config would need to have the right suffix as well.
As this is just before everyone goes on holidays any significant change to CPG-Infra would not be wise.
I am going to create a JIRA ticket to revisit in a new year and implement the proper solution as Vivian suggested.
For now I am going to approve this so people are not blocked.

jmarshall:
Thanks for the suggestions but we don't believe changing the output of functions in _metamist.py_ is the best approach. We will continue based on PR #1068.

PR #1067, opened 2024-12-15
Second attempt at -test correction

MattWellie:
We also need to drop the -test suffix in this dictionary structure to match the later `-test` trimming.

The creation of the `SGs per project` dictionary uses the project name in the metamistresults: `SG.Sample.project`. In test projects, this project string also has `-test` in.

This behaviour won't be required in cpg-flow as we've already scrapped this structure


Alexander-Stuckey (approved):
Can confirm that this change, in addition to the others, fixes the issue with `-test` datasets

PR #1068, opened 2024-12-16
Removing test suffix from dataset names for multicohorts.

milo-hyben:
Datasets name gets populated from metamist project name for multicohorts, the old way was from the config file.
Metamist contain suffix for test projects. 
This PR removes them when creating relevant Dataset by name.

### DIFF
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -105,7 +105,7 @@ def _populate_cohort(cohort: Cohort, sgs_by_dataset_for_cohort, read_pedigree: b
     TODO (mwelland): so we can lose a layer of the structure here
     """
     for dataset_name in sgs_by_dataset_for_cohort.keys():
-        dataset = cohort.create_dataset(dataset_name)
+        dataset = cohort.create_dataset(dataset_name[:-5] if dataset_name.endswith('-test') else dataset_name)
         sgs = sgs_by_dataset_for_cohort[dataset_name]
 
         for entry in sgs:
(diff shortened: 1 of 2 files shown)


cpg_workflows/inputs.py
```diff
@@ -105,7 +105,7 @@ def _populate_cohort(cohort: Cohort, sgs_by_dataset_for_cohort, read_pedigree: b
     TODO (mwelland): so we can lose a layer of the structure here
     """
     for dataset_name in sgs_by_dataset_for_cohort.keys():
-        dataset = cohort.create_dataset(dataset_name)
+        dataset = cohort.create_dataset(dataset_name[:-5] if dataset_name.endswith('-test') else dataset_name)
```
jmarshall, line 108: I reckon clearer just as

```suggestion
        dataset = cohort.create_dataset(dataset_name.removesuffix('-test'))
```

MattWellie:
This issue is solved by https://github.com/populationgenomics/production-pipelines/pull/1063 - to fully solve we need to remove the `-test` suffix from the Dataset, and also from the result of the `SG.Sample.Project` retrieved from Metamist. 

The `.removesuffix('-test')` syntax is cleaner, but this an incomplete solution, just FYI.

milo-hyben:
> This issue is solved by #1063 - to fully solve we need to remove the `-test` suffix from the Dataset, and also from the result of the `SG.Sample.Project` retrieved from Metamist.
> 
> The `.removesuffix('-test')` syntax is cleaner, but this an incomplete solution, just FYI.

Hey @MattWellie, modifying metamist.py is not good idea.
The original Vivian fix modified dataset_name for the whole scope of the function, this PR only modifies it for cohort.create_dataset, so the issue Alex was experiencing in the second try would not occur.

jmarshall (approved):
Confirmed that the new test fails without the _inputs.py_ change, and succeeds with it.

Thanks @milo-hyben, looks good to me!

PR #1069, opened 2024-12-16
Dragen ica pipeline 1068

MattWellie:
Union of #1029 and #1068

#1029 currently uses a fix from #1063 which works, but also requires a change in `metamist.py`.

This updates #1029 to swap to the #1068 change for testing, which is the most concise required fix.


Alexander-Stuckey (approved):
Looks good to me

PR #1070, opened 2024-12-16
RD: Resume from previous combiner

MattWellie:
This PR looks pretty chunky, but it's actually quite minor

- In the shift from Joint-calling to gVCF combining we need the ability to cost-effectively resume from a previous partial run failure
- Hail permits this by providing the combiner plan from an initial attempt
- The currently Combiner implementation is being shared with PA, and in that the `force=True` argument to the Combiner is hard coded
- The force argument prevents hail from reusing the combiner plan, and instead guarantees a from-scratch attempt with each run

The solution here is to duplicate the combiner.py code into an `rd_combiner` folder, and modify the force argument to be modifiable from config (defaulting instead to being False).

The only substantive change being made is the addition of a separate configurable parameter for `force`, and the logging to clarify both if a save_path was provided, and if Hail will be able to use it. To keep the changes separate from PA's usage it's been duplicated away.

---

Now that there are 5 different stages where we might want to override the Hail JAR file I've re-ordered the config parameters. Instead of the config having `rd_combiner.STAGE. jar_spec_revision` I've swapped it around to `workflow. jar_spec_revisions.STAGE`, which means that all the 5 config entries are at the same level in the hierarchy, which to me reads much better.

I've also set these defaults to be the latest prod-ready fix containing the two patches we're interested in.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.32.0
+current_version = 1.32.1
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.32.0
+  VERSION: 1.32.1
 
 jobs:
   docker:
--- a/configs/defaults/rd_combiner.toml
+++ b/configs/defaults/rd_combiner.toml
@@ -23,7 +23,7 @@ snps_recal_disc_size = 20
 
 # choices of jar specs to use for the various steps
 # without this setting we would default to matching the native Hail version in the container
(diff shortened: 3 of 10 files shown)


cassimons (approved):
Great. Yea, I think having a RD specific version of the stage is the right way to go in any case.

EddieLF (approved):
Looks good to me, nothing is substantially different from [combiner.py](https://github.com/populationgenomics/production-pipelines/blob/5bcd771829a1ba6f40ee1a1a0967cffabe59631a/cpg_workflows/large_cohort/combiner.py), just the addition of the force argument managed via the config. Approved!

MattWellie:
I've also tacked on ability to modify the driver/worker resources from config. I think they were raised to highmem/high cores while we were chasing the StreamConstraints issues, and now we're stuck with each Worker provisioning 30GB but only using ~300MB of memory. The defaults will either bring costs way down or cause failures, and if it fails we can modify from config

PR #1079, opened 2024-12-18
Combine Exomiser variant results

MattWellie:
Relevant to https://github.com/populationgenomics/talos_manuscript/pull/1/files

This takes all the exomiser result files for a family and combines them into 
- a JSON file
- a Hail Table

The JSON file has the shape

```json
{
   "FAMILY_ID": {
        "chr1:1234:A:C": [
             {
                 "rank": 1,
                 "moi": "AD"
             },
             {
                 "rank": 8,
                 "moi": "AR"
             }
         ]
    },
    "FAMILY_ID_2": {
        "chr1:5678:A:C": [
             {
                 "rank": 1,
                 "moi": "AR"
             }
         ]
    }
}
```

which contains the 4 main(?) pieces of information about the variant:
- the variant itself
- each family where that was a flagged variant
- for each event, the rank and MOI

This list structure can feature the same variant multiple times per family, e.g when they have the same variant flagged with multiple MOIs. The list-of-dictionaries could also be extended easily to include mode information from the Exomiser result file.

It also reprocesses the data and writes a Hail Table:

| locus | alleles | family_details |
| ----- | ------ | -------------- |
| chr1:1234 | [A, C] | FAMILY_RANK_MOI::FAMILY_RANK_MOI |
| chr1:5678 | [G, T] | FAMILY_RANK_MOI |

This Hail table would be useful as an annotation in Talos
- new hail category
- join the variant MT with the Exomiser HT
- if there's a match, keep the variant
- downstream we can optionally process the data in a more detailed way, e.g. only keep flags which are above rank `n`

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.32.4
+current_version = 1.32.5
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.32.4
+  VERSION: 1.32.5
 
 jobs:
   docker:
--- a/configs/defaults/exomiser.toml
+++ b/configs/defaults/exomiser.toml
@@ -1,4 +1,5 @@
 [workflow]
+name = 'exomiser'
 status_reporter = 'metamist'
 
 # attributes to provision the Exomiser VMs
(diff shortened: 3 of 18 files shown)


cassimons (approved):
Looks good. The only potential issue I see with this format is that it makes it hard test if a given family is represented in this result file (ie was exomiser run). Is it worth adding a meta field that is simply the list of IDs used to generate the file?

MattWellie:
Yup, or I could just doubly index the whole json - on family and on variant. It would duplicate the data a bit, but these files are tiny anyway. Orrrrr... make the json output nicer, indexed on family and having details stored better, and save the compressed representation just for the hail table

PR #1081, opened 2024-12-18
Don't always force a new combiner to run

Alexander-Stuckey:
As done in the RD combiner, don't always force a new combiner to run from scratch, if a plan exists for it already.

### DIFF
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -14,6 +14,17 @@ highmem_workers = true
 # Calling intervals (defauls to whole genome intervals)
 #intervals_path =
 
+# Override for QoB JAR to fix teh StreamConstraints issue
+jar_spec_revision = 'f5b65edb8f1ed8e431f1799546a8e96d0ddff7d9'
+
+[combiner]
+force_new_combiner = false
+vds_version = "0.1"
+memory = "8Gi"
+storage = "10Gi"
+vds_analysis_ids = []
+merge_only_vds = false
+
 [vqsr]
 # VQSR, when applying model, targets indel_filter_level and snp_filter_level
 # sensitivities. The tool matches them internally to a VQSLOD score cutoff
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -96,8 +96,8 @@ class Combiner(CohortStage):
 
         j: PythonJob = get_batch().new_python_job('Combiner', (self.get_job_attrs() or {}) | {'tool': HAIL_QUERY})
         j.image(image_path('cpg_workflows'))
-        j.memory(combiner_config.get('memory', "4Gi"))
-        j.storage(combiner_config.get('storage', "5Gi"))
+        j.memory(combiner_config.get('memory', '4Gi'))
+        j.storage(combiner_config.get('storage', '5Gi'))
 
         # Default to GRCh38 for reference if not specified
         j.call(
@@ -109,6 +109,7 @@ class Combiner(CohortStage):
             gvcf_paths=new_sg_gvcfs,
             vds_paths=vds_paths,
             save_path=output_paths['combiner_plan'],
+            force_new_combiner=config_retrieve(['combiner', 'force_new_combiner'], False),
         )
 
         return self.make_outputs(cohort, output_paths, [j])
@@ -337,8 +338,8 @@ class MakeSiteOnlyVcf(CohortStage):
         sitesvcf_config = config_retrieve('sitesvcf')
 
         j.image(image_path('cpg_workflows'))
-        j.memory(sitesvcf_config.get('memory', "4Gi"))
-        j.storage(sitesvcf_config.get('storage', "5Gi"))
+        j.memory(sitesvcf_config.get('memory', '4Gi'))
+        j.storage(sitesvcf_config.get('storage', '5Gi'))
 
         init_batch_args: dict[str, str | int] = {}
         for config_key, batch_key in [('highmem_workers', 'worker_memory'), ('highmem_drivers', 'driver_memory')]:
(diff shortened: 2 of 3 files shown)


configs/defaults/large_cohort.toml
```diff
@@ -14,6 +14,17 @@ highmem_workers = true
 # Calling intervals (defauls to whole genome intervals)
 #intervals_path =
 
+# Override for QoB JAR to fix teh StreamConstraints issue
+jar_spec_revision = 'f5b65edb8f1ed8e431f1799546a8e96d0ddff7d9'
```
jmarshall, line 18: I'd strongly recommend not checking this in. In particular, [this one is now 404](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1734320188057349?thread_ts=1732046164.229849&cid=C030X7WGFCL).

Alexander-Stuckey: Yes, was just in the process of checking that I had the correct jar. Will update

jmarshall: Okay, I'll just leave it as: I'd strongly recommend not checking this in.

cpg_workflows/stages/large_cohort.py
```diff
@@ -96,8 +96,8 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
 
         j: PythonJob = get_batch().new_python_job('Combiner', (self.get_job_attrs() or {}) | {'tool': HAIL_QUERY})
         j.image(image_path('cpg_workflows'))
-        j.memory(combiner_config.get('memory', "4Gi"))
-        j.storage(combiner_config.get('storage', "5Gi"))
+        j.memory(combiner_config.get('memory', '4Gi'))
+        j.storage(combiner_config.get('storage', '5Gi'))
```
katiedelange, line 100: I think it's confusing to have both defaults - if you set 8Gi/10Gi in the large_cohort.toml, then this 4Gi/5Gi can be removed (it's always overwritten anyway, and it will just lead to confusion about where 8Gi is coming from when 4Gi is apparently the default). Or, set the default here so it's easy to find in the code, but don't overwrite it in the toml.

It doesn't look like we have a consistent approach through the large cohorts pipeline, not sure if others (@MattWellie?) have been more consistent elsewhere.

cpg_workflows/stages/large_cohort.py
```diff
@@ -109,6 +109,7 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
             gvcf_paths=new_sg_gvcfs,
             vds_paths=vds_paths,
             save_path=output_paths['combiner_plan'],
+            force_new_combiner=config_retrieve(['combiner', 'force_new_combiner'], False),
```
katiedelange, line 112: As above, either have the default here or in the large_cohort.toml, but probably not both?

configs/defaults/large_cohort.toml
katiedelange: Echoing John's comment, I don't think we want to hard code _any_ jar_spec_revision in. This is going to change over time, and we want the person running it to have to actually make a deliberate decision.

cpg_workflows/stages/large_cohort.py
MattWellie: I've been crap at this in the past, sometimes leaving the in-code default slightly different to the in-config default as things evolve. I'm trying to fix all those instances, and I'd generally recommend the code and config defaults be identical.

I think there's utility in this in-code default being there so that you have the at-a-glance estimate of how big (expensive) each job is just from reading the code, without having the config open alongside

MattWellie, line 100: as a side note, any storage request between 0-10Gi will be inflated to 10Gi. That's hail's minimum actual storage if you request any storage

MattWellie: I personally like the in-code defaults to get a quick estimate of job size without looking into the config, but it can muddle things. A comment would be equally clear, and would make the pipeline fail if a default isn't set in config (for pipelines it always is). The pipeline is woefully short of code conventions to follow, maybe we should discuss some of those syntax (in)consistencies when we come to fragmenting the codebase.

There's a tangent to go down about scripts which can run in the pipeline context or standalone, but that's not relevant to any pipeline stage code.

MattWellie (approved):

cpg_workflows/stages/large_cohort.py
MattWellie: I personally like the in-code defaults to get a quick estimate of job size without looking into the config, but it can muddle things. A comment would be equally clear, and would make the pipeline fail if a default isn't set in config (for pipelines it always is). The pipeline is woefully short of code conventions to follow, maybe we should discuss some of those syntax (in)consistencies when we come to fragmenting the codebase.

There's a tangent to go down about scripts which can run in the pipeline context or standalone, but that's not relevant to any pipeline stage code.

configs/defaults/large_cohort.toml
Alexander-Stuckey: Fair points, I will take it out so that it has to be specified by the user.

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey: Changed so now the user must set it via the config

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey: These have been removed.

katiedelange: +1 for discussing some of these conventions as part of the codebase fragmentation.

katiedelange: Edited to use the configuration defaults, as per the force_new_combiner decision below (the latest revision had removed the setting of storage and memory in the combiner entirely).

katiedelange (approved):

PR #1082, opened 2024-12-18
make icasdk optional and dont preempt other pr

Alexander-Stuckey:
Make the icasdk import optional, so that it doesn't break other workflows that use images that don't have icasdk installed in it.

Also revert the change that removed the test suffix from the dataset name for multicohorts, so that https://github.com/populationgenomics/production-pipelines/pull/1068 is not preempted. This change was left in the original merge by mistake, rather than reverting after testing.

### DIFF
--- a/cpg_workflows/inputs.py
+++ b/cpg_workflows/inputs.py
@@ -107,7 +107,7 @@ def _populate_cohort(cohort: Cohort, sgs_by_dataset_for_cohort, read_pedigree: b
     for dataset_name in sgs_by_dataset_for_cohort.keys():
         # If the dataset_name string ends in -test, strip test from the end
         # This is to ensure compatibility with the structure of our storage configs.
-        dataset = cohort.create_dataset(dataset_name.removesuffix('-test'))
+        dataset = cohort.create_dataset(dataset_name)
         sgs = sgs_by_dataset_for_cohort[dataset_name]
 
         for entry in sgs:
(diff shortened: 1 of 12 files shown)


cpg_workflows/inputs.py
```diff
@@ -107,7 +107,7 @@ def _populate_cohort(cohort: Cohort, sgs_by_dataset_for_cohort, read_pedigree: b
     for dataset_name in sgs_by_dataset_for_cohort.keys():
         # If the dataset_name string ends in -test, strip test from the end
         # This is to ensure compatibility with the structure of our storage configs.
```
jmarshall, line 109: Also remove this comment to revert the change.

Alexander-Stuckey: oops

jmarshall:
Unfortunately the `try … except` thing didn't do the trick for local dry-runs:

```
  File "/Users/johnm/production-pipelines/cpg_workflows/stages/dragen_ica/ica_utils.py", line 52, in <module>
    api_instance: project_analysis_api.ProjectAnalysisApi,
                  ^^^^^^^^^^^^^^^^^^^^
NameError: name 'project_analysis_api' is not defined
```

However I have an alternative suggestion that does work — remove all that `try … except` around the `import icasdk` lines, and instead do this (also a bit of a hack) in _main.py_:

```python
try:
    from cpg_workflows.stages.realign_genotype_with_dragen import CancelIcaPipelineRun, DownloadDataFromIca
except ImportError:
    # If icasdk is not installed, ignore error to allow other workflows to be used
    class CancelIcaPipelineRun:
        pass

    class DownloadDataFromIca:
        pass
```

jmarshall (approved):
Thanks!

PR #1083, opened 2024-12-19
Update image for dragen ica

katiedelange:
Including DRAGEN in the main production pipelines file leads to problems when trying to use old cpg_workflow images that don't contain the ICA SDK. This will no longer be an issue following on from the CPG Flow refactor, but to avoid breaking changes until then, we separate the DRAGEN pipeline out with it's own runner script.

[Putting in the PR on behalf of Alex, as he's locked out of GitHub :( ]


jmarshall (approved):
Looks good to me. What a pain “maybe there” SDKs turn out to be in Python!

You could adjust the comment on line 4 of _dragen_378_realignment_runner.py_ to “Entry point to run Dragen-offloaded workflows” or something.

PR #1084, opened 2024-12-19
Split usage of two output types

MattWellie:
I'm now tracking the exomiser Variant-level and Gene-level outputs from the same stage. Each type of output is directed exclusively to one downstream stage, so I just need to add some specificity.

This condition ensures that the `FAMILY_variants` output is only directed to the `variants` aggregator, not the gene-level aggregator.

Example error: 
```
    'vcf': vcf_inputs[family],
KeyError: 'A1_variants'
```


annacuomo (approved):
lgtm

PR #1085, opened 2024-12-19
Update VDS schema to drop data

MattWellie:
Add a script to do a VDS-editing 2-step

- take a VDS, and the list of `entry.gvcf_info` fields to remove
- remove fields from schema and write to a new location
- read back in, and overwrite original location

This is a pretty invasive operation, but AFAIK the data we have issues with is not currently recoverable, and there's no way to provide parameters during combining to remove specific fields, so we need to either edit the VDSs or gVCFs in-place to save our joint-call.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.32.5
+current_version = 1.32.6
 commit = True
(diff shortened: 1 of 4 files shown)


EddieLF (approved):
Nice simple way to patch this, looks good to me

PR #1086, opened 2024-12-19
Revert cpg_workflows pin to Hail 132 now that issues are understood

MattWellie:
After the roller-coaster of Hail JAR iteration, we now understand the underlying cause of the inflated expense issues - there is no "expensive version of Hail", per se, so we can release this pin to an earlier formal release version.

This is relevant for VDS work - it appears that even if the QOB JAR file is an earlier hail version, the codecs used to write data (VDSs) are determined by this version. By pinning to this earlier version we write VDSs which natively have an older specification, and would have less efficient interval queries.

This is evident if you read any of the recently written VDSs with Hail 133+, you get a logging notice suggesting that your VDSs be patched in place to gain the improved functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,7 @@ setup(
         'cpg-utils>=5.1.1',
         'cyvcf2==0.30.18',
         'analysis-runner>=2.43.3',
-        # Pin Hail to the release prior to 0.2.133
-        # see https://centrepopgen.slack.com/archives/C018KFBCR1C/p1731047651300539
-        'hail==0.2.132',
+        'hail>=0.2.133',
         'networkx>=2.8.3',
         'obonet>=0.3.1',  # for HPO parsing
         'grpcio-status>=1.48,<1.50',  # Avoid dependency resolution backtracking
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         'cpg-utils>=5.1.1',
         'cyvcf2==0.30.18',
         'analysis-runner>=2.43.3',
-        'hail>=0.2.133',
+        'hail==0.2.133',  # Pin Hail at CPG's installed version
         'networkx>=2.8.3',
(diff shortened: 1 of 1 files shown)


EddieLF (commented):
Sounds good, though I'm not as familiar with all the jar related issues as the other requested reviewers are so I'll let someone else approve.

The exact message logged can be seen in this batch [here](https://batch.hail.populationgenomics.org.au/batches/557100/jobs/1)
```
WARNING:backend.service_backend:You are reading a VDS written with an older version of Hail.
  Hail now supports much faster interval filters on VDS, but you'll need to run either
  `hl.vds.truncate_reference_blocks(vds, ...)` and write a copy (see docs) or patch the
  existing VDS in place with `hl.vds.store_ref_block_max_length(vds_path)`.
```

setup.py
```diff
@@ -16,9 +16,7 @@
         'cpg-utils>=5.1.1',
         'cyvcf2==0.30.18',
         'analysis-runner>=2.43.3',
-        # Pin Hail to the release prior to 0.2.133
-        # see https://centrepopgen.slack.com/archives/C018KFBCR1C/p1731047651300539
-        'hail==0.2.132',
+        'hail>=0.2.133',
```
jmarshall, line 19: Best to put the previous comment back too:

```suggestion
        'hail>=0.2.133',  # Pin Hail at CPG's installed version
```

Traditionally we use `==` for this particular dependency. I don't know that it matters too much, but perhaps we should stick with tradition…

katiedelange (approved):
Agree with this change - we definitely want to shift to the newer VDS specification, and this hail release version turned out to be unrelated to the cost issue. Just noting that the VDS specs are changing fast: there was a major patch-requiring change in 131, and apparently now again in 133, so this is going to be something we need to keep an eye on as we start to use the Combiner more broadly.

jmarshall (approved):

PR #1087, opened 2024-12-20
Put partition command in correct method call

MattWellie:
Not sure why my local linter and the CI check both thought this was ok... The `n_partitions` argument was placed in the wrong method call
https://batch.hail.populationgenomics.org.au/batches/557481/jobs/3


annacuomo (approved):
lgtm

PR #1088, opened 2024-12-30
Couple of exomiser tweaks

MattWellie:
I needed to adjust a behaviour in the gene-level aggregation Stage - the gene-level aggregation ignores all '-variants' files, the variant aggregation only uses '-variants' files. 

This felt like a decent opportunity to change the way this Stage works in general, to make it behave the same as the other stages (use an entrypoint instead of finding a script by guessing the path to it)


annacuomo (approved):
LGTM

PR #1090, opened 2025-01-03
Exomiser scripts were swapped

MattWellie:
Absolute despair at this point

I've messed this up a HIGH number of times

During the last PR where I was renaming scripts to make them consistent I somehow swapped the script/entrypoint. So dumb.


annacuomo (approved):
😆

Issue #1092, opened 2025-01-08
Align stage - Checkpoint the sorted BAM produced before MarkDuplicates to allow for its reuse


EddieLF:
## Problem

Aligning genomes (and exomes) can be a long and expensive process, depending on the alignment input file(s), the resources required for the jobs, the VM preemptions that unavoidably occur when using spot instances, and the transient errors that force reruns. 

The `Align` stage in `cpg_workflows` operates in two main steps. In summary, the first step aligns the inputs into a sorted BAM, and the second step identifies and removes duplicates reads in the aligned BAM, producing the final CRAM analysis output (and a .crai index file). 

During the first step, extra resources might be required to align the inputs for especially large sequencing groups. Using higher memory machines and additional disk might be necessary depending on the inputs. These jobs may also take up to a day to run to completion, sometimes costing upwards of $35 if the job is using extra resources AND faces multiple unfortunate preemptions. 

Not only can costs quickly rise for the first step, but they may even have to be repeated if errors interrupt the second step, MarkDuplicates. This is because the sorted BAM produced in the first step is not saved to cloud storage; it exists only within the context of the batch and is not accessible if the batch fails or is cancelled. If Markduplicates is interrupted for any reason, the pipeline must run again from the first step before MarkDuplicates can be re-attempted. 

## Solution

We should checkpoint the BAM produced by the first step of the `Align` stage. This would involve writing the BAM to the dataset's `tmp` bucket, where it would exist for up to 7 days before being deleted. The MarkDuplicates job could then read the BAM from this temp storage location, rather than reading it in through Hail Batch resource file (`job.sorted_bam`). 

Then, when the `Align` stage is queued, the jobs (align / merge) in the first step can be skipped if the sorted BAM exists in the expected location. If it does, then we should only queue a MarkDuplicates job which uses the sorted BAM as input.

## Potential issues

- The `Align` stage and/or jobs may need significant refactoring to achieve the above.
- The `production-pipelines`/`cpg_workflows`/`cpg_flow` refactoring is currently in flight and this work may add unnecessary complications.
- Uploading the sorted BAM to cloud storage and then downloading it for MarkDuplicates will incur more costs than using the Hail Batch resource groups.
  - The download costs could be mitigated by still using the Hail Batch resource file if the two steps are completed in the same batch, and only downloading from the temp bucket when MarkDuplicates needs to be re-run.
[...]

PR #1095, opened 2025-01-08
Add option to write sorted bam to temp and reuse it for markdups

EddieLF:
Addresses #1092 

Adds some logic that modifies the align job so that the sorted BAM created by the align / merge jobs can be re used by MarkDuplicates if MarkDuplicates fails or is cancelled.
 
- **NEW CONFIG OPTION**: `workflow.checkpoint_sorted_bam` - If this is set and the file does not already exist, write the sorted BAM to temp storage prior to MarkDuplicates. 
  - The sorted BAMs are large and it would be expensive to store them in temp for all sequencing groups being aligned. Therefore, it's best to keep this as a config option to be toggled on for runs with expensive align jobs, where checkpointing the BAMs would be wise. 

- **NEW LOGIC**: Re-use the sorted BAM in temp storage as input to MarkDuplicates (if it exists), skipping the align / merge jobs. The config option `workflow.check_intermediates` must be set to allow reuse.

---

This works by adding a new path arg to the `Align` stage & jobs - `sorted_bam_path`, which is a path in the dataset's temp bucket. 

Then, if set in the config, the sorted BAM output by the align / merge jobs will be written to this temp path. This will allow MarkDuplicates to read the sorted BAM from the temp storage if the MarkDuplicates job needs to be re-attempted (which is not uncommon, due to memory requirements & transient errors). 

If the sorted BAM exists in the temp path on a rerun can be reused, the align jobs will be skipped, and only the MarkDuplicates job will be executed. A "dummy job" will be created in place of the align / merge job which usually creates the sorted BAM. This is required due to the structure of the `finalise_alignment` function in the align job module. 

---

### Test run

1. Sorted BAM is checkpointed by aligning fastqs and writing the merged and sorted BAM to tmp storage
[Driver](https://batch.hail.populationgenomics.org.au/batches/588601/jobs/1) and [Batch](https://batch.hail.populationgenomics.org.au/batches/588602?q=&last_job_id=50)
2. Sorted BAM is re-used to skip straight to MarkDuplicates, which reads the sorted BAM from temp
[Driver](https://batch.hail.populationgenomics.org.au/batches/588626/jobs/1) and [Batch](https://batch.hail.populationgenomics.org.au/batches/588627)
3. Genotype jobs are correctly queued after sorted BAM reuse
[Driver](https://batch.hail.populationgenomics.org.au/batches/588628/jobs/1) and [Batch](https://batch.hail.populationgenomics.org.au/batches/588629)

### DIFF
--- a/cpg_workflows/jobs/align.py
+++ b/cpg_workflows/jobs/align.py
@@ -131,7 +131,7 @@ def align(
         - for dragmap, submit an extra job to extract a pair of fastqs from the cram/bam,
         because dragmap can't read streamed files from bazam.
 
-    - if the sorted bam already exists, skip the alignment job(s) and go straight to markdup.
+    - if the sorted bam can be reused, skip the alignment job(s) and go straight to markdup.
 
     - if the markdup tool:
         - is biobambam2, deduplication and alignment/merging are submitted within the same job.
@@ -173,7 +173,6 @@ def align(
     sorted_bams = []
 
     if can_reuse(sorted_bam_path):
-        # If the sorted BAM can be reused, skip the alignment job(s) and go straight to markdup.
         logging.info(f'{sequencing_group.id} :: Re-using sorted BAM: {sorted_bam_path}')
         # Its necessary to create this merge_or_align_j object to pass it to finalise_alignment,
         # and to declare the sorted_bam_path as a resource, so it can be written to the checkpoint.
@@ -622,7 +621,6 @@ def finalise_alignment(
         align_cmd += f' > {j.sorted_bam}'
 
     if not can_reuse(sorted_bam_path):
-        # Submit the alignment job(s) if we're not reusing the sorted BAM
         j.command(command(align_cmd, monitor_space=True))  # type: ignore
 
     if (
@@ -630,7 +628,7 @@ def finalise_alignment(
         and not sorted_bam_path.exists()
         and config_retrieve(['workflow', 'checkpoint_sorted_bam'], False)
     ):
-        # Only write the sorted BAM to the checkpoint if it doesn't already exist and the config is set
+        # Write the sorted BAM to the checkpoint if it doesn't already exist and the config is set
         logging.info(f'Will write sorted bam to checkpoint: {sorted_bam_path}')
         b.write_output(j.sorted_bam, str(sorted_bam_path))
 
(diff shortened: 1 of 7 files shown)


cpg_workflows/jobs/align.py
```diff
@@ -262,13 +265,20 @@ def align(
...
         overwrite=overwrite,
     )
     if md_j and md_j != merge_or_align_j:
         jobs.append(md_j)
 
+    if md_j:
+        if md_j.attributes.get('reusing_sorted_bam'):
+            # Remove all jobs except markdup if sorted bam is reused
+            logging.info(f'Reusing sorted bam from temp - only markdup job will be submitted for {sequencing_group.id}')
+            jobs = [md_j]
```
MattWellie, line 280: II think this is fine, but I'd be keen to leave this as a logging statement based on the md_j attributes, then flow down to the existing `return jobs` below. If the task is in a position to resume then we should never have more than one job to return, so this is fine, but I'm scared of the way that this structural change (only return one job, which _should_ be 1 of 1) could interfere with workflow scheduling (inter-stage dependencies are only set based on the jobs returned by the Stage, not the all scheduled jobs in Hail)

cpg_workflows/jobs/align.py
```diff
@@ -169,7 +172,21 @@ def align(
...
 
-    if not sharded:  # Just running one alignment job
+    if can_reuse(sorted_bam_path):
+        logging.info(f'{sequencing_group.id} :: Re-using sorted BAM: {sorted_bam_path}')
+        # Its necessary to create this merge_or_align_j object to pass it to finalise_alignment,
+        # and to declare the sorted_bam_path as a resource, so it can be written to the checkpoint.
+        job_attrs = (job_attrs or {}) | dict(label='Reusing sorted bam', tool='Reusing sorted bam')
+        merge_or_align_j = b.new_job('Reusing sorted bam', job_attrs or {})
+        merge_or_align_j.image(config_retrieve(['workflow', 'driver_image']))
+        merge_or_align_j.sorted_bam = b.read_input(str(sorted_bam_path))
```
MattWellie, line 182: I don't know if this is valid syntax 🤔 
Even if it is I hate the idea of making a fake job just to read a file into the batch. We can definitely re-work the interface here

MattWellie (approved):
I hate the overall structure of this Stage, but given the spaghetti we have to work with this is the least invasive version of this change.

If we plan to keep this Stage in its current form, this should be used as a demonstration of how tricky it is to alter even a relatively minor behaviour

michael-harper:
Happy with this change - but would prefer to hold off on merging until PA has finished running a dataset with current implementation :)

PR #1099, opened 2025-01-13
Change py version in lint workflow

nevoodoo:
This is to make it consistent with the versions defined in the pre-commit config and the rest of the codebase. Also stops the linting from failing.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -16,7 +16,7 @@ jobs:
 
     - uses: actions/setup-python@v5
       with:
-        python-version: '3.10.0'
+        python-version: '3.10.9'
         cache: 'pip'
         cache-dependency-path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     - uses: actions/setup-python@v5
       with:
-        python-version: '3.11.0'
+        python-version: '3.10.9'
```
jmarshall, line 19: Just use `'3.10'`, which will select the highest 3.10.x patch release.
There seems to have been a reason for specifying 3.11 — see 287c751c1f672054b47dc81c68e840d76aeafe75 — but 🤷

jmarshall (approved):

vivbak (approved):
yaaayy

PR #1102, opened 2025-01-14
Fix force new combiner param

michael-harper:
Updating the image has worked.

The was `Combiner` stage using a `cpg_workflows` image from the config `australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_workflows:1.32.1` which was over a month old, that image was built December 17th, the `force_new_combiner` argument was added on the 20th. 

Updating the code so that the Combiner job runs inside the more recently built `driver_image` that contains the new parameter works.

Also noting that, the code printed out in the Combiner driver job itself was not the code being run by the Combiner workers. Which is why, despite the `force_new_combiner` parameter being visible in the print out in [this](https://batch.hail.populationgenomics.org.au/batches/588883/jobs/1) job, it didn't actually exist in the job that was doing the work.

So because the driver and worker jobs are using different images, they actually have a different `combiner.run` method description, but both are using the same arguments (from the driver job).

Shoutout to @MattWellie for providing context

### DIFF
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -24,6 +24,8 @@ memory = "8Gi"
 storage = "10Gi"
 vds_analysis_ids = []
 merge_only_vds = false
+# if false, use non-preemptible VMs
+preemptible_vms = false
 
 [vqsr]
 # VQSR, when applying model, targets indel_filter_level and snp_filter_level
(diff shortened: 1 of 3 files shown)


michael-harper:
Also adding RDs implementation to use non-premptible workers for the Combiner driver jobs. [Working batch here](https://batch.hail.populationgenomics.org.au/batches/588902)

MattWellie (approved):
LGTM, I think all these changes deliver more predictable performance

PR #1104, opened 2025-01-15
Remove doubled printing during workflow planning

MattWellie:
These lines are logged twice - once each for Dataset, Cohort, and MultiCohort Stages. These same lines are already printed in get_action, and make the logging during the driver job look a bit silly.

Should be fixed in cpg-flow, but why wait


vivbak (approved):
Thanks Matt! @nevoodoo, do you have time to port this change over to CPG Flow today?

PR #1105, opened 2025-01-15
Re-work exomiser stages

MattWellie:
Firstly - this keeps the bulk of the content the same, so that's cool. Exomiser is run in the same way.

Dodgy things:
- within each dataset I previously collected all families with at least one affected member, then arbitrarily chose one of the affected participants and focused the analysis on them. This was usually _the_ proband, but didn't address multi-proband families, or affected parents.
- The ID chosen for the results was the external ID for the chosen family member. I also called this `family_id`, when it wasn't. Getting this pipeline going was a nightmare, I blame general delerium.
- Generating these analyses on the basis of an external sample ID makes it really tricky to aggregate them later, e.g. for use in Talos and Seqr. One Ext. Sample ID can have multiple CPG IDs, but each CPG ID has only one Sample ID. By carrying out analyses and storing results on the CPG SG ID, we can unambiguously know which member has been analysed as a proband, and link this through more easily to other systems.

Changes:
- do the same filtering to find every family with at least one affected member
- for each family:
    - check if there are any affected members with parents in the family (in which case we assume at least one affected parent and child)
    - if there was an affected child, generate a separate analysis for every affected child only (do not analyse parents as probands)
    - if there was no affected child, generate a separate analysis for every affected family member (singletons, sibling duos, de novo trios)
    - each of these analyses uses the primary member's CPG ID, not external ID
   
- we don't use Exomiser 13, so the output path `exomiser_14_[input|output]` is now just `exomiser_[input|output]`
- Exomiser 13 job scheduling methods are deleted
- per-analysis family gVCFs are now in `-tmp`

This includes a novel little follow-on SequencingGroupStage `RegisterSingleSampleExomiserResults` to register the results of a previous DatasetStage. This was needed because `RunExomiser` is a DatasetStage, but it generates per-SG results, which can't be easily registered. There's a separate helper method to pull these analysis entries, which is used to prevent VCFs being regenerated when the final results already exist.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.33.2
+current_version = 1.34.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.33.2
+  VERSION: 1.34.0
 
 jobs:
   docker:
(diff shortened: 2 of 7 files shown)


cassimons (approved):
Very cool thanks

PR #1108, opened 2025-01-16
Exomiser Analysis Outputs can be a string or a dict

MattWellie:
I failed to account for the situation where a **first class files** section can be either a dictionary or a string.

This tests for the exact output type, then processes accordingly.

Specific cause here (a guess based on results) is that the `outputs` dictionary attached to the analysis entry is only generated if the output file path is a _file_, not a _directory_. In situations where we can register either a HT/MT or an exact single file as the given type, we need to account for both data structures when parsing metamist results. @nevoodoo / @vivbak  can you confirm if this is the parsing logic in Metamist for FCF? That would make sense given that directories won't have `size` or `file_checksum` attrs.

### DIFF
--- a/cpg_workflows/stages/exomiser.py
+++ b/cpg_workflows/stages/exomiser.py
@@ -10,7 +10,7 @@ As a dataset Stage
 
 from functools import cache
 
-from cpg_utils import Path
+from cpg_utils import Path, to_path
 from cpg_utils.config import config_retrieve
 from cpg_utils.hail_batch import get_batch
 from cpg_workflows.jobs.exomiser import (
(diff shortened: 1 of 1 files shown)


cassimons (approved):

nevoodoo:
@MattWellie that's correct, because of the complex structures of `.mt` or `.ht` "files", it was difficult to do this and we had those as a stretch goal. At the moment, it will treat those files as if they were strings, not file paths:

https://github.com/populationgenomics/metamist/blob/f569d8228225cf80b79b6192007b02fdfc9bffdf/models/models/output_file.py#L186-L193

I would love to accommodate MT/HT files but I haven't gotten my head around how, just yet. If this ends up being a strong blocker, or is causing issues, I will happily pick this up and look into it!

MattWellie:
@nevoodoo nah, fine by me. In the use-case I have I'll only need to pick up files anyway, and they should all be dicts. Determining size of these complex objects is.. non trivial. Something like a `gcloud storage du -s -r gs://PATH.vds` to get a total recursive size summary can take hours to run on highly partitioned data :/

PR #1111, opened 2025-01-16
Change the VDS Combiner defaults to combine in smaller batches 

katiedelange:
Each step consists of 5 intermediate VDSs, each combining 50 gVCFs


MattWellie (approved):
LGTMatt

PR #1118, opened 2025-01-24
Improve ICA Pipeline Robustness Against Failures

michael-harper:
This PR addresses challenges in improving the robustness and usability of the DRAGEN ICA pipeline, particularly when managing stage dependencies and output file checks. The current system’s reliance on the existence of all outputs from a previous stage makes it difficult to handle scenarios like canceling a pipeline run when only partial outputs (e.g., pipeline ID) are available.

Slack thread [here](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1737690966683519)

### Problem

When monitoring the status of an ICA pipeline run the monitor stage wrote it's own output. However, in the event of a pipeline failure by ICA, the Monitor stage would write it's output and when we went to re-run that stage, the output would exist and so the stage would not get triggered to run again. This leaves us in a position where we have to manually go back and delete a stages outputs in GCP just so we can re-run a pipeline...not ideal even if the failure rates are low. So the option to merge the monitor and align+genotype stages into one stage and have them be job dependencies was adopted and this combined stage have two `expected_outputs`. However, this precipitated a similar issue with the cancellation stage (a stage designed to cancel a pipeline run if the user wants to). This is detailed as follows:

For example, the CancelIcaPipelineRun stage depends on the AlignGenotypeWithDragen stage. This means it won’t execute unless all outputs from AlignGenotypeWithDragen—including `_success.json` and `_pipeline_id.json`—exist. This behavior prevents the cancellation of a pipeline when only the pipeline ID is available.

### Splitting Downloading of Outputs to Allow Registering in Metamist
To register cram and gvcf files outputted by the ICA pipeline separately the easiest way was to split the downloading of these files into separate stages with the stage decorator parameter `analysis_type` as `cram` and `gvcf` respectively. The has resulted in three separate stages for downloading outputs, the `DownloadCramFromIca`, `DownloadGvcfFromIca`, and `DownloadDataFromIca` which downloads the rest of the outputs.

This was necessary because of the necessity of knowing the pipeline ID in order to download as well as the existence checks that occur by using the stage decorator.

### Proposed Solutions

This PR implements a solution which involves merging CancelIcaPipelineRun into AlignGenotypeWithDragen (and renaming the stage to `ManageDragenPipeline`). The combined stage now:
	1.	Handles pipeline aligning+genotyping, monitoring and cancellation together.
	2.	Triggers cancellation only if the configuration enables it (e.g., [ica.pipelines.cancel_cohort_run] = false).
	3.	Simplifies dependencies, converting the former stage relationship into job dependencies, making it easier to manage file existence checks.


cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -142,90 +140,137 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
 
 @stage(required_stages=[PrepareIcaForDragenAnalysis, UploadDataToIca])
-class AlignGenotypeWithDragen(SequencingGroupStage):
+class ManageDragenPipeline(SequencingGroupStage):
     # Output object with pipeline ID to GCP
     def expected_outputs(
         self,
         sequencing_group: SequencingGroup,
-    ) -> cpg_utils.Path:
+    ) -> dict[str, cpg_utils.Path | str]:
```
MattWellie, line 148:
```suggestion
    ) -> dict[str, cpg_utils.Path]:
```

both these outputs are Paths

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -142,90 +140,137 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
+                name='CancelIcaPipelineRun',
+                attributes=(self.get_job_attrs() or {}) | {'tool': 'Dragen'},
+            )
+            cancel_job.image(image=image_path('ica'))
+            cancel_pipeline_result = cancel_job.call(
+                cancel_ica_pipeline_run.run,
+                ica_pipeline_id_path=str(outputs['pipeline_id']),
+                api_root=ICA_REST_ENDPOINT,
+            ).as_json()
+            logging.info(f'Pipeline cancelled: {cancel_pipeline_result}')
```
MattWellie, line 181:
```suggestion
```

You can't print the output of a job that hasn't run yet

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -142,90 +140,137 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
+            get_batch().write_output(
+                cancel_pipeline_result,
+                str(
+                    sg_bucket
+                    / GCP_FOLDER_FOR_RUNNING_PIPELINE
+                    / f'{sequencing_group.name}_pipeline_cancelled_at{slugify(str(datetime.now()))}.json',
```
MattWellie, line 187: This file includes the datetime, so it's guaranteed to be unique, but doesn't include the pipeline ID which was cancelled. 

If you swap this to be `f'{sequencing_group.name}_pipeline_{pipeline_id}_cancelled.json'` then you can still know the datetime from the file metadata, but you can also check for existence of this file (don't attempt to cancel the same workflow ID twice)

cpg_workflows/stages/dragen_ica/run_align_genotype_with_dragen.py
```diff
@@ -139,4 +139,7 @@ def run(
         )
 
         logging.info(f'Submitted ICA run with pipeline ID: {analysis_run_id}')
-        return {'pipeline_id': analysis_run_id}
+        with cpg_utils.to_path(output_path).open('w') as f:
+            f.write(analysis_run_id)
```
MattWellie, line 143: Some tidy-up needed here I think. This method writes the String pipeline ID to a file, but returns a pipeline ID as a dict. 
On a re-run it reads a String, so the downstream monitor job has to tolerate a string or a dict. I don't think passing a dict rather than a string has any tangible benefit, but that was the original implementation and now it's stuck between the two changes having to handle both.

michael-harper: During testing I thought I ran in to issues with a `PythonJob` returning a dictionary but I've just done some testing [here](https://batch.hail.populationgenomics.org.au/batches/589699) and after cancelling and re-running [here](https://batch.hail.populationgenomics.org.au/batches/589700/jobs/1) with this stage returning just a string it looks like it's working :)

Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
 
-RUN pip install metamist
+RUN pip install metamist \
+&& pip install git+https://github.com/Illumina/ica-sdk-python.git \
+&& pip install typing-extensions --upgrade
```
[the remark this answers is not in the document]
michael-harper: Removing this from the Dockerfile results in import errors in jobs, see [here](https://batch.hail.populationgenomics.org.au/batches/590487/jobs/1). Even though the `cancel_job` is being told to use the `ica` image (`cancel_job.image(image=image_path('ica'))`) it fails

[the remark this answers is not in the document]
MattWellie: Isn't that what this is for? https://github.com/populationgenomics/images/blob/main/images/ica/Dockerfile

The image in the job you're pointing at is `cpg_workflows:ica`, built from this branch after the ICA install was removed. 

What you actually need is a new version of the actual ICA image, built from [this branch](https://github.com/populationgenomics/images/tree/build-ica-from-cpg_workflows-ica) (on top of the cpg_workflows:ica image, install ICASDK)

Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
 
-RUN pip install metamist
```
katiedelange, line 3: I don't think we want to drop this metamist install from the main production pipelines Dockerfile?

cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
```diff
@@ -37,29 +36,28 @@ def run(
...
                 api_instance=api_instance,
-                path_params=path_params | {'analysisId': ica_pipeline_id},
+                path_params=path_params | {'analysisId': pipeline_id},
             )
         if pipeline_status == 'SUCCEEDED':
-            logging.info(f'Pipeline run {ica_pipeline_id} has succeeded')
-            return {'pipeline': 'success'}
+            logging.info(f'Pipeline run {pipeline_id} has succeeded')
+            return {'pipeline': pipeline_id, 'status': 'success'}
         elif pipeline_status in ['ABORTING', 'ABORTED']:
```
katiedelange, line 60: Just noting that "ABORTING" will never get hit in this elif; however, I think it's correct behaviour to wait until the monitor gets confirmation that the run was successfully "ABORTED".

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -141,91 +141,153 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
     def expected_outputs(
         self,
         sequencing_group: SequencingGroup,
-    ) -> cpg_utils.Path:
+    ) -> dict[str, cpg_utils.Path]:
         sg_bucket: cpg_utils.Path = sequencing_group.dataset.prefix()
-        return sg_bucket / GCP_FOLDER_FOR_RUNNING_PIPELINE / f'{sequencing_group.name}_pipeline_id.json'
+        return {
+            'output': sg_bucket / GCP_FOLDER_FOR_RUNNING_PIPELINE / f'{sequencing_group.name}_pipeline_success.json',
+            'pipeline_id': sequencing_group.dataset.prefix(test=True)
```
katiedelange, line 169: I think we want to drop the "test=True" here?

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -249,47 +311,170 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
+            cram_md5=$(icav2 projectdata list --parent-folder /{bucket_name}/{ica_analysis_output_folder}/{sequencing_group.name}/{sequencing_group.name}-$pipeline_id/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram.md5sum --match-mode EXACT -o json | jq -r '.items[].id')
+            icav2 projectdata download $cram_id {sequencing_group.name}/{sequencing_group.name}.cram --exclude-source-path
+            icav2 projectdata download $crai_id {sequencing_group.name}/{sequencing_group.name}.cram.crai --exclude-source-path
+            icav2 projectdata download $cram_md5 {sequencing_group.name}/{sequencing_group.name}.cram.md5sum --exclude-source-path
+            gcloud storage cp {sequencing_group.name}/{sequencing_group.name}.cram gs://{bucket_name}/{GCP_FOLDER_FOR_ICA_DOWNLOAD}/{sequencing_group.name}/
+            gcloud storage cp {sequencing_group.name}/{sequencing_group.name}.cram.crai gs://{bucket_name}/{GCP_FOLDER_FOR_ICA_DOWNLOAD}/{sequencing_group.name}/
+            gcloud storage cp {sequencing_group.name}/{sequencing_group.name}.cram.md5sum gs://{bucket_name}/{GCP_FOLDER_FOR_ICA_DOWNLOAD}/{sequencing_group.name}/
```
katiedelange, line 380: Ideally, it would be good if we could do an automatic check on this checksum after download.

cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -249,47 +311,170 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
+            target=sequencing_group,
+            data=self.expected_outputs(sequencing_group=sequencing_group),
+            jobs=ica_download_job,
+        )
+
+
+@stage(
+    analysis_type='gvcf',
+    analysis_keys=['gvcf', 'gvcf_tbi'],
+    required_stages=[ManageDragenPipeline],
```
katiedelange, line 394: I think we need to make an equivalent stage that is dependent on GvcfMlrWithDragen, so we download and register the recalibrated gVCF (once we have that stage implemented). Until that recalibration is set up, it's fine if this stage never gets reached (i.e. we won't add it as a pipeline dependency downstream).

Ideally we would download and register that full chain in Metamist 
- CRAM (produced by DRAGEN 3.7.8) -> gVCF (base produced by DRAGEN 3.7.8) -> recalibrated gVCF (produced by DRAGEN-ML).

katiedelange (commented):
Looking good, thanks for all this hard work Harper - I've added a few minor comments.

A general comment that is not for this PR, but so I don't forget: I think all the files under stages/dragen_ica/* are jobs not stages, and actually belong in jobs/dragen_ica/*? It will get shifted later with cpg_flow anyway, but we may want to make a second PR to restructure those if it's not too painful.

cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
Alexander-Stuckey: I think that I added this code in for the chance that someone cancels the pipeline in ICA, and then tries to use this pipeline to cancel it (could happen if there is an issue with our infra and we need to cancel a run I guess)

cpg_workflows/stages/realign_genotype_with_dragen.py
michael-harper: Have added some bash commands to the end to check md5's, latest [run here](https://batch.hail.populationgenomics.org.au/batches/590638)

cpg_workflows/stages/realign_genotype_with_dragen.py
michael-harper: If I'm interpreting this correctly we should remove `DownloadGvcfFromIca` as a `required_stage` from `DownloadDataFromIca`?

katiedelange: I think this is all fine as-is, and the once we add the ML stage in we will also add in a `DownloadRecalibratedGVcfFromIca` (or some less chunky name) that is then another `DownloadDataFromIca` dependency. But we can leave `DownloadGvcfFromIca` alone for the base gVCF.

katiedelange (approved):
Looks great to me, thanks for all the hard work on this Harper, Alex and team!

PR #1122, opened 2025-01-29
Filtering tweaks for hap.py

MattWellie:
This change adds two main things:

- barfs if we try to extract a non-existent sample from a callset. At the moment Hail will continue and generate a header-only VCF

- enables us (via config) to keep all quality filtered variants in the VCF passed to Hap.py. 2 benefits:

1. Hap.py will correctly report ALL vs. PASS stats (currently these are the same)
2. The VCF exported by Hap.py contains all sites which were compared, and the truth/query sites identified during the comparison

It looks from the documentation that hap.py should only be writing this VCF if explicitly requested, but we're currently generating it anyway?

https://github.com/Illumina/hap.py/blob/master/doc/happy.md#additional-outputs

-> gs://cpg-validation-main/validation/bac0696ab7bb34b62a15619a98c8d3401d78b4_3577/HG004_NA24143__CPG255224.happy.vcf.gz

If this VCF is written out with Truth and Query columns, every row tagged as TP/FN/FP, and the relevant filters, we can potentially do a deeper dive on the VQSR tranches where we're losing real variants.


configs/defaults/validation.toml
```diff
@@ -1,9 +1,14 @@
...
 status_reporter = 'metamist'
 
+input_datasets = ['validation']
+
+# if True, the single sample VCF extracted from the MT will be filtered to PASS-only, otherwise it will be the full VCF
+vcf_pass_only = false
+
 # if this is populated, the validation workflow will start from this specific MatrixTable
-#matrixtable = "placeholder"
+#matrixtable = "gs://cpg-seqr-main/seqr_loader/d6cc67110be4b232067508037bd50b8212f2d5_4205/AnnotateCohort/cohort.mt"
```
cassimons, line 11: Do we wan this set by default?

cassimons (approved):

PR #1123, opened 2025-01-29
Long read seqr load pipeline

EddieLF:
Updates the seqr loader long read pipeline with several changes compared to the initial version merged in 2024.

This PR is intended to create SNV/Indel and SV callsets from the provided PacBio VCFs, all the way up to the point they can be loaded into Seqr. There are a number of other changes being trialled in the draft PR https://github.com/populationgenomics/production-pipelines/pull/1115 which are not in scope for this PR, mentioned at the bottom.

---

**BAM to CRAM**
- Rename the stage to `ConvertPacBioBamToCram`
- Minor changes to stage and job config options, allowing user to change resourcing requirements and input reference fasta.

**SNPs / Indels VCFs Annotation**
- Update Metamist query to account for new file name formats
- Normalise the sample VCFs before merging and joining by adding a `bcftools norm -m -any -f ref.fa -c s in.vcf` command to the VCF processing step in the `ReFormatPacBioSNPsIndels` stage. This is required because merging un-normalized VCFs was not compatible with the downstream Hail stage to annotate the cohort mt. See this [Slack discussion](https://centrepopgen.slack.com/archives/C082NA58R7S/p1737442347259089?thread_ts=1737410276.815149&cid=C082NA58R7S) for more detail.
- New arg `remove_invalid_contigs` - will filter the matrix table to remove entries with a contig not in the standard 22XYM set. See [example job](https://batch.hail.populationgenomics.org.au/batches/589417/jobs/1) where not removing invalid contigs causes failure.
- Set the `mt.info.AC` and `mt.info.AF` rows to take the array values at index `mt.a_index` rather than `mt.a_index - 1`. This is a bit involved, but basically because we have to normalise the sample VCFs before merging, and then overwrite the `AF`/`AC` values with those from `hl.variant_qc` before calling `hl.split_multi_hts`, it means `a_index` is always `1` for every row. But `mt.info.AC` / `mt.info.AF` arrays always have length 2, with the WT allele values at index 0, and the alt allele values at index 1. Since we want the alt allele values, we need to remove the `-1` from the indexing. See this [Slack discussion](https://centrepopgen.slack.com/archives/C037XB40XFW/p1737526958523809?thread_ts=1737502555.496289&cid=C037XB40XFW) for more detail.

**SV VCF annotation**
- Update Metamist query to account for new file name formats
- When checking for faidx / sniffles REF base consistency, make sure that the sniffles base is uppercase. This is because the hg38 masked reference only contains upper case bases, while the input VCF may have lower case bases, indicating masking/repetition. See [Slack discussion here](https://centrepopgen.slack.com/archives/C082NA58R7S/p1738189204370249) for more detail.

---

**NOT IN SCOPE**
- Adding read group information during the BAM > CRAM conversion
- Reheadering the CRAMs with the hg38 masked reference contig information
- Adding CramQC / Somalier stages

### DIFF
--- a/cpg_workflows/jobs/bam_to_cram.py
+++ b/cpg_workflows/jobs/bam_to_cram.py
@@ -45,17 +45,17 @@ def bam_to_cram(
     res = STANDARD.set_resources(
         j,
         ncpu=nthreads,
-        storage_gb=config_retrieve(['resource_overrides', 'bam_to_cram_storage_gb'], 50),
+        storage_gb=config_retrieve(['resource_overrides', 'bam_to_cram', 'storage_gb'], 50),
     )
 
     j.declare_resource_group(
         sorted_cram={
             'cram': '{root}.cram',
-            'cram.crai': '{root}.cram.crai',
+            'crai': '{root}.cram.crai',
         },
     )
 
-    cmd = f'samtools view -@ {res.get_nthreads() - 1} -T {fasta.fasta} -C {input_bam.bam} | tee {j.sorted_cram["cram"]} | samtools index -@ {res.get_nthreads() - 1} - {j.sorted_cram["cram.crai"]}'
+    cmd = f'samtools view -@ {res.get_nthreads() - 1} -T {fasta.fasta} -C {input_bam.bam} | tee {j.sorted_cram["cram"]} | samtools index -@ {res.get_nthreads() - 1} - {j.sorted_cram["crai"]}'
     j.command(command(cmd, monitor_space=True))
 
     return j, j.sorted_cram
--- a/cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py
+++ b/cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py
@@ -139,7 +139,11 @@ def modify_sniffles_vcf(file_in: str, file_out: str, fa: str, new_id: str | None
 
             # a quick check, if we can
             if l_split[3] != 'N':
-                assert new_base == l_split[3][0], f'Discrepancy between faidx and Sniffles: {new_base}, {l_split[3]}'
+                # If using the hg38 masked reference, the base will always be upper case
+                # So make sure the compairson to the Sniffles REF is upper case too
+                assert (
+                    new_base == l_split[3][0].upper()
+                ), f'Discrepancy between faidx and Sniffles: {new_base}, {l_split[3]}'
 
             # replace the REF String
             l_split[3] = new_base
--- a/cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py
+++ b/cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py
@@ -140,7 +140,7 @@ def modify_sniffles_vcf(file_in: str, file_out: str, fa: str, new_id: str | None
             # a quick check, if we can
             if l_split[3] != 'N':
                 # If using the hg38 masked reference, the base will always be upper case
-                # So make sure the compairson to the Sniffles REF is upper case too
+                # So make sure the comparison to the Sniffles REF is upper case too
                 assert (
                     new_base == l_split[3][0].upper()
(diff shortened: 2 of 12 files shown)


cpg_workflows/jobs/bam_to_cram.py
```diff
@@ -45,17 +45,17 @@ def bam_to_cram(
...
     j.declare_resource_group(
         sorted_cram={
             'cram': '{root}.cram',
-            'cram.crai': '{root}.cram.crai',
+            'crai': '{root}.cram.crai',
         },
     )
 
-    cmd = f'samtools view -@ {res.get_nthreads() - 1} -T {fasta.fasta} -C {input_bam.bam} | tee {j.sorted_cram["cram"]} | samtools index -@ {res.get_nthreads() - 1} - {j.sorted_cram["cram.crai"]}'
+    cmd = f'samtools view -@ {res.get_nthreads() - 1} -T {fasta.fasta} -C {input_bam.bam} | tee {j.sorted_cram["cram"]} | samtools index -@ {res.get_nthreads() - 1} - {j.sorted_cram["crai"]}'
```
MattWellie, line 58: this syntax is cool, but given that there are two operations in stream each running with `res.get_nthreads() - 1` threads, would it be more resource/threadsafe to write the CRAM first and index as a separate operation? 

If it works it works, just wondering

cpg_workflows/scripts/long_read_sniffles_vcf_modifier.py
```diff
@@ -139,7 +139,11 @@ def modify_sniffles_vcf(file_in: str, file_out: str, fa: str, new_id: str | None
 
             # a quick check, if we can
             if l_split[3] != 'N':
-                assert new_base == l_split[3][0], f'Discrepancy between faidx and Sniffles: {new_base}, {l_split[3]}'
+                # If using the hg38 masked reference, the base will always be upper case
+                # So make sure the compairson to the Sniffles REF is upper case too
```
MattWellie, line 143:
```suggestion
                # So make sure the comparison to the Sniffles REF is upper case too
```

MattWellie (approved):
LGTM, when we migrate I expect the long and short reads will be broken into two pipelines, which makes the logic in cpg_workflows/query_modules/seqr_loader.py much clearer

cpg_workflows/jobs/bam_to_cram.py
jmarshall: This PR doesn't change this aspect of the code, so it's already working successfully even though it's using more threads than it perhaps should. I think because the two samtools commands are interdependent via one writing to the other, the buffering and throughput delays mean that in practice it's not going to soak up all the cores it's illegitimately trying to.

Looks like this pattern appears twice in this file and also once in _align_rna.py_.

It's probably okay, but it would be better to use `--write-index` as _picard.py_ does:

```python
cmd = f'samtools view -@ {res.get_nthreads() - 1} -T {fasta.fasta} --write-index -C -o {j.sorted_cram["cram"]} {input_bam.bam}'
```

However it may or may not be worth making that actual behaviour change, and I would be a bit more happy about it if our samtools image were more up-to-date (but I guess it works well enough for _picard.py_!)…

MattWellie: I'm aware it's unchanged, I just hadn't seen it before.

Our default samtools image `australia-southeast1-docker.pkg.dev/cpg-common/images/samtools:1.18` has `--write-index` functionality

cassimons (approved):
Looks good Ed 🙌

cpg_workflows/jobs/bam_to_cram.py
EddieLF: Thanks for the comments on this both @MattWellie and @jmarshall 

As noted, it is working fine so far in the testing environment. A future PR will alter this job to get these CRAMs, working with the Somalier and QC stages. I'll group this suggested change in the scope of that next PR.

PR #1124, opened 2025-01-30
Improve target as x methods

### LINKED ISSUE #611: Inputs.as_X methods are unintuitive

MattWellie:
Closes #611

Minimal modification:
### as_str

- gets the target value from StageInput in the same way
- checks if the target is a `str` or `os.PathLike` - other types are not tolerated
- convert the value to a String, and return it

### as_path

- gets the target value from StageInput in the same way
- checks if the target is a `str` or `os.PathLike` - other types are not tolerated
- return the `to_path` representation

---

I flirted with the idea of keeping the `cast` calls - they don't change the actual object, they just communicate the type to the checker. I don't really rate casts. 

The logic could be "if the object is the expected type, return it casted to the correct type, else convert it", but that adds an extra if condition, which isn't really necessary. Running `str('string')` has no computational burden, and running `to_path('PATH')` calls through to `to_anypath`, which returns the unchanged object if it passes `isinstance(s, (CloudPath, Path))`. 

The other aspect here is that the method has a return type of Path or String, which is enough to satisfy the type checker that a type has changed, without using cast.

Given that there's no burden in converting these objects to the type they already are, there's no need to use a typing-only transformation, and no real reason to add extra conditional lines to avoid the type conversion.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.35.1
+current_version = 1.36.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.35.1
+  VERSION: 1.36.0
 
 jobs:
   docker:
--- a/cpg_workflows/workflow.py
+++ b/cpg_workflows/workflow.py
@@ -171,8 +171,6 @@ class StageOutput:
         res = self._get(key)
         if not isinstance(res, (os.PathLike, str)):
             raise ValueError(f'{res} is not a str or valid Pathlike, will not convert.')
-        if isinstance(res, str):
-            return res
         return str(res)
 
     def as_path(self, key=None) -> Path:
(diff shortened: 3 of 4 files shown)


MattWellie:
Closed - exists in cpg-flow

PR #1127, opened 2025-01-31
Fix compound key in hail

MattWellie:
Correction to #1125 

VEP outputs don't have REF/ALF alleles as individual attributes, they only have them within the larger input string, so we need to do some string manipulation to get them

See https://github.com/populationgenomics/production-pipelines/pull/602/files#r1936638900

I previously addressed these issues in a concept PR/branch for making seqr_loader consistently normalise/split multiallelic variants 😅

### DIFF
--- a/cpg_workflows/scripts/vep_json_to_ht.py
+++ b/cpg_workflows/scripts/vep_json_to_ht.py
@@ -203,9 +203,7 @@ def vep_json_to_ht(vep_result_paths: list[str], out_path: str):
     ht = ht.transmute(vep=ht.f0)
     # we're using split multiallelics, so we need to compound key on chr/pos/ref/alt
     ht = ht.annotate(
-        locus=hl.locus(
-            ht.vep.seq_region_name, hl.parse_int(ht.vep.input.split('\t')[1])
-        ),
+        locus=hl.locus(ht.vep.seq_region_name, hl.parse_int(ht.vep.input.split('\t')[1])),
         alleles=ht.vep.allele_string.split('/'),
     )
     ht = ht.key_by(ht.locus, ht.alleles)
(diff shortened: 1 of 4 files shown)


EddieLF (approved):
Seems good

cpg_workflows/scripts/vep_json_to_ht.py
```diff
@@ -201,14 +201,11 @@ def vep_json_to_ht(vep_result_paths: list[str], out_path: str):
...
-    # happens for indels). So instead parsing POS from the original VCF line stored
-    # as ht.vep.input field.
-    original_vcf_line = ht.vep.input
-    start = hl.parse_int(original_vcf_line.split('\t')[1])
-    chrom = ht.vep.seq_region_name
-    ht = ht.annotate(locus=hl.locus(chrom, start))
     # we're using split multiallelics, so we need to compound key on chr/pos/ref/alt
+    ht = ht.annotate(
+        locus=hl.locus(ht.vep.seq_region_name, hl.parse_int(ht.vep.input.split('\t')[1])),
+        alleles=ht.vep.allele_string.split('/'),
```
cassimons, line 207: Is it easy to make this split on '/' OR '|'

Given our push to LR we are going to start seeing phased VCFs more and more so it would be good to build that into our muscle memory

MattWellie: A split on one of two different separators isn't explicitly available as a hail method, but we can do a replace for `|` -> `/`, then split on the latter 🤔

cassimons: Not a bit deal for this PR in particular - just one of those footguns that makes me nervous about basic string parsing a VCF.

MattWellie: I'll bookmark this and have a look for some phased calls in the final MT. I think this might be fine as this is the Alleles string, but the phasing is in the GT field. I would expect that even for `0|1` the alleles being passed to VEP are still in `A/T` format, with the option of being `A/C/T` for multiallelics

Issue #1066, opened 2024-12-13
Run gcloud compose on the VEP JSONs


MattWellie:
This is a definite candidate for running `gcloud compose` on the individual JSONL files, instead of expecting Hail Batch to aggregate 2500 separate JSON files when building the unified HT. These files are all a single JSON variant annotation per row, 1:1 with the variant MT, so we shouldn't have issues concatenating them all together into one fat JSON file

_Originally posted by @MattWellie in https://github.com/populationgenomics/production-pipelines/issues/1037#issuecomment-2540371099_

The VEP PythonJob has to aggregate all the separate JSON files (each being atomic and in chromosomal-position order). It's presumably easier to do this on a single file instead of thousands, but I can't back that up with much other than _vibes_.

There's template code in this new workflow to do a rolling concatenation of all files, so there's stuff we can reuse

MattWellie:
Closing because... it's just not necessary. Call me again if the JSON aggregation process fails

PR #1128, opened 2025-01-31
Add QOB params to VEP parsing

MattWellie:
Possibly relevant: https://github.com/populationgenomics/production-pipelines/issues/1066

See https://batch.hail.populationgenomics.org.au/batches/589951

The QOB aggregation and manipulation of the VEP JSONs is running into some memory issues I've not had previously. Maybe adding another single operation into the JSON parsing was a step too far 😬 

The issue ticket #1066 is a suggestion that we can use the rolling gcloud compose to concatenate all the VEP JSONs into a single file before parsing into a Hail Table. That innately feels like way less work in memory than combining 2500 separate files... If this quick patch still results in memory issues I'll investigate further, but this has previously worked fine with no additional resources

A semi-related update being bundled here is that the per-module JAR spec overrides are removed from scripts and config. These are hopefully now not needed, given the current main version contains patches and logging.

### DIFF
--- a/configs/defaults/rd_combiner.toml
+++ b/configs/defaults/rd_combiner.toml
@@ -28,14 +28,14 @@ indel_recal_disc_size = 20
 snps_gather_disc_size = 10
 snps_recal_disc_size = 20
 
-[workflow.es_index]
-# if false, use a non-preemptible instance to run the ES export
-spot_instance = false
-
 # without this setting we use the native Hail version in the container
 # not currently in use as the cpg-utils default contains these patches by default, plus extra logging
 # most recently set to 'a8be268326b2ac168035b629a78465c9d94fc7b9' - 0.2.133 + 400'retry + StreamConstraints patch
 #default_jar_spec_revision = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
+
+[workflow.es_index]
+# if false, use a non-preemptible instance to run the ES export
+spot_instance = false
 
 [combiner]
 # used to decide if we should resume from a previous combiner plan
--- a/configs/defaults/rd_combiner.toml
+++ b/configs/defaults/rd_combiner.toml
@@ -32,15 +32,10 @@ snps_recal_disc_size = 20
 # if false, use a non-preemptible instance to run the ES export
 spot_instance = false
 
-# choices of jar specs to use for the various steps
-# without this setting we would default to matching the native Hail version in the container
-# currently set to 'a8be268326b2ac168035b629a78465c9d94fc7b9' - 0.2.133 + 400'retry + StreamConstraints patch
-[workflow.jar_spec_revisions]
-annotate_cohort = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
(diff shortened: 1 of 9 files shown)


cassimons (approved):

configs/defaults/rd_combiner.toml
```diff
@@ -32,15 +32,10 @@ snps_recal_disc_size = 20
...
-[workflow.jar_spec_revisions]
-annotate_cohort = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
-annotate_dataset = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
-combiner = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
-densify = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
-subset = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
+# without this setting we use the native Hail version in the container
+# not currently in use as the cpg-utils default contains these patches by default, plus extra logging
+# most recently set to 'a8be268326b2ac168035b629a78465c9d94fc7b9' - 0.2.133 + 400'retry + StreamConstraints patch
+#default_jar_spec_revision = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
```
jmarshall, line 38: I'd recommend not putting in a config item called `default_jar_spec_revision`, as that'd be confusing. Nothing will be looking for it in `workflow.es_index` anyway, so uncommenting it would also confusingly do nothing.

jmarshall: Glad to see the per-module overrides being removed BTW, now that they're no longer necessary as the default one exists now. But with this being taken care of in cpg-utils and cpg-infrastructure, you shouldn't need any comment here at all really.

MattWellie: This was incorrectly nested inside `es_index`, and has been moved to the correct location

EddieLF (approved):
love how neat these init_batch statements look now 👌

configs/defaults/rd_combiner.toml
```diff
@@ -28,20 +28,15 @@ indel_recal_disc_size = 20
 snps_gather_disc_size = 10
 snps_recal_disc_size = 20
 
+# without this setting we use the native Hail version in the container
+# not currently in use as the cpg-utils default contains these patches by default, plus extra logging
+# most recently set to 'a8be268326b2ac168035b629a78465c9d94fc7b9' - 0.2.133 + 400'retry + StreamConstraints patch
+#default_jar_spec_revision = 'a8be268326b2ac168035b629a78465c9d94fc7b9'
```
jmarshall, line 34: Please don't check references to `workflow.default_jar_spec_revision` into production-pipelines config files. We want to be able to update that in cpg-infrastructure as necessary, and it will lead to confusion and unsupportable configurations if various workflows override it.

MattWellie: On the basis that this is a commented out section of the config, with an explanatory comment saying "if you need to change the JAR spec, this is how you'd do it", I'm going to leave it in...

We currently have an undocumented, optional, and occasionally essential behaviour in cpg-utils. Surely more information is better than less.

PR #1129, opened 2025-02-03
VEP patch to correct for InDel variants

MattWellie:
Previous process (seqr_loader) only used contig and position from the VEP data (`seq_region_name` and position from `hl.parse_int(ht.vep.input.split('\t')[1])`)

New process on split multiallelics requires the Alleles [REF, ALT] to be populated, as this is part of the table key when we join VEP data into the variant table.

I was populating the Allele Strings using `vep.allele_string`, which seemed logical. This was a bad idea, as [this comment](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/vep.py#L357-L359)   alluded to. As I've clarified in the extended comment attached to the new code block, VEP was doing a transformation of all Indels to a dodgy format (e.g. T/- instead of AT/A). By taking VEP.allele_string, ALL Indels in the VEP data were given an Allele string featuring `-` as either the REF or ALT.

- VEP HT was indexed on invalid Alleles
- Invalid alleles couldn't be joined to the variant data
- All variants had VEP completely missing VEP annotation
- Many downstream impacts, mainly evident to us as Talos filtering removing ALL indels (the first filter for "is this gene in the Green gene list" failed for every InDel - vep.gene_ids = `NA`)

---

The current performance of the RD_Combiner pipeline is now even worse than before 😅 

- Before we were incorrectly adding one set of annotation to all variants which were split from the same multiallelic row. Most variants were being correctly annotated.
- Now all SNVs are being correctly annotated, even if they were part of a multiallelic row
- **But** all Indels, multiallelic or not, are failing 😬

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.35.5
+current_version = 1.36.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.35.5
+  VERSION: 1.36.0
 
 jobs:
   docker:
--- a/cpg_workflows/scripts/vep_json_to_ht.py
+++ b/cpg_workflows/scripts/vep_json_to_ht.py
@@ -207,9 +207,15 @@ def vep_json_to_ht(vep_result_paths: list[str], out_path: str):
     ht = hl.import_table(paths=vep_result_paths, no_header=True, types={'f0': json_schema})
     ht = ht.transmute(vep=ht.f0)
     # we're using split multiallelics, so we need to compound key on chr/pos/ref/alt
+    # the ref and alt can't be taken from the VEP output, as they are manipulated
+    # e.g. ["T","-"] for a Deletion, or ["-","AC"] for an insertion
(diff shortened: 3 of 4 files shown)


EddieLF:
I think I understand the issue but I'm not sure about the implementation, what happens downstream with the `locy` and `alleles2` ht fields? Do these get used? Are the `locus` and `alleles` fields defined elsewhere?

MattWellie:
> what happens downstream with the `locy` and `alleles2` ht fields

😆 total ballsup on my part - I was playing with the syntax locally to make sure that it works, but I couldn't overwrite the `alleles` or `locus` attributes because they were keys. I altered the name to make sure it worked ok:

```
+-------------------+---------------+-----------------------------------+---------------+------------------------------------+
| vep.variant_class | locus         | alleles                           | locy          | alleles2                           |
+-------------------+---------------+-----------------------------------+---------------+------------------------------------+
| str               | locus<GRCh38> | array<str>                        | locus<GRCh38> | array<str>                         |
+-------------------+---------------+-----------------------------------+---------------+------------------------------------+
| "SNV"             | chr1:10026    | ["A","T"]                         | chr1:10026    | ["A","T"]                          |
| "insertion"       | chr1:10108    | ["-","AACCCT"]                    | chr1:10108    | ["C","CAACCCT"]                    |
```

EddieLF (approved):
Nice, looks good now

PR #1130, opened 2025-02-03
Bamtocram resourcegroup fix

EddieLF:
Changing the resource group to `".crai"` instead of `".cram.crai"` had the effect of saving the index files with just the `.crai` extension. This had downstream effects in Seqr, so the change is being reverted to how it was in [the prior commit](https://github.com/populationgenomics/production-pipelines/blob/536aa1b9942911d2415454856c88df912f1cf4b4/cpg_workflows/jobs/bam_to_cram.py).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.0
+current_version = 1.36.1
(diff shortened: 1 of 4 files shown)


MattWellie (approved):
Nice

PR #1131, opened 2025-02-04
Create index files script

EddieLF:
A script that will take a GCP folder and then find BAM/CRAM files within that don't have corresponding bai/crai index files. Any files missing an index will have a samtools index job queued. 

The script is intended to create indexes for reads received to ingest, because having an index file present means that the reads can be re-aligned easier. 

See a test example of the script running here:
https://batch.hail.populationgenomics.org.au/batches/590102/jobs/1

In this test, the directory has 2 bams and 4 crams. 2 of the crams already have index .crai files, so they are skipped. Samtools index jobs are then queued for the 2 bams and 2 crams without indexes, and the bai and crai files are written to the correct output path.

Question - Do we need to also add a "samtools sort" in this code?

### DIFF
--- a/scripts/create_index_files.py
+++ b/scripts/create_index_files.py
@@ -50,7 +50,7 @@ def index_with_samtools(
     """
     for file_to_index_path, index_file_path in input_files:
         input_file = b.read_input(file_to_index_path)
-        j = b.new_bash_job(b, f'samtools index {file_to_index_path}')
+        j = b.new_bash_job(f'samtools index {file_to_index_path}')
         j.image(image_path('samtools'))
         # Set resource requirements
         storage_gb = 20 if to_path(file_to_index_path).stat().st_size < 1e10 else 100
--- a/scripts/create_index_files.py
+++ b/scripts/create_index_files.py
@@ -34,8 +34,7 @@ def find_files_to_index(
         else:
             continue
 
-        index_blob = bucket.blob(index_blob_name)
-        if not index_blob.exists():
+        if index_blob_name not in blob_names:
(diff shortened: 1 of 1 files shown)


cassimons:
Thanks Ed. Should not need a sort. Unsorted bams are typically just an intermediate step in a pipeline and I can never remember them being stored as a final output by any pipeline.

cassimons (commented):
LGTM in passing but can you review pls @MattWellie

MattWellie (approved):
This will work - I think you can do it in about 1/3 the lines: 

 - find_files_to_index explicitly checks bam and cram files, and builds the path to the index. Instead of returning a list of Paths, then later checking which type each one is, it should return a list of String tuples `(BAM/CRAM, Index)`. Downstream you never need to re-check for either `.bam` or `.cram`, and you won't need to conditionally apply `.bai` or `.crai` depending on the file type, you already have the index path.
 - index_files can be deleted - you don't need the conditional, the files are processed the same whether they're BAM or CRAM
 - index_with_samtools should take the list of String pairs, read the first into the batch as a single file, then run `samtools index ... {bam_or_cram} -o {job.output}`, then write `job.output` to the index path. No need to make a resource group as you only need a single output file.

Functionally you end up in the same place, but with less Hail Boilerplate

EddieLF:
Thanks for the suggestions @MattWellie, I had been staring at it and I could tell there was much more code than there needed to be but I wasn't really sure what to shave off. Implemented your suggestions and it still works great.

https://batch.hail.populationgenomics.org.au/batches/590111

PR #1132, opened 2025-02-07
Shard exome bams/cram inputs during align stage

EddieLF:
This PR adds a config option `workflow.exome_realign_shards` which can be used to determine how many shards to split bams/crams into before aligning each shard in parallel with dragen-os.

Currently, for genomes, the input is sharded into 10 fragments with bazam. But for exomes, no sharding takes place. 

It seems like the dragen-os align job struggles with processing an unsharded BAM, even when a high memory worker is allocated for the align job. The job will report an out of memory error (code 137) as the docker container crashes, but then it will restart itself and reattempt the align job. This has lead to overspending, as jobs that should simply fail instead restart themselves and accrue costs without any chance of succeeding.

An extensive discussion has taken place in [this slack thread](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1738889941313949), which contains links to various batches showing the issue and how this sharding option resolves it.

**Tests to realign an 8.3gb exome CRAM**

Failing run with no sharding, using a high mem machine:
https://batch.hail.populationgenomics.org.au/batches/590285

Successful run with 5 shards, using high mem machines:
https://batch.hail.populationgenomics.org.au/batches/590291

Successful run with 2 shards, using standard mem machines:
https://batch.hail.populationgenomics.org.au/batches/590293

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.1
+current_version = 1.36.2
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.36.1
+  VERSION: 1.36.2
 
 jobs:
   docker:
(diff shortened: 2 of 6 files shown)


MattWellie (approved):
Weird issue!

PR #1133, opened 2025-02-11
Updates to RD_Combiner

MattWellie:
Slight change to config so that we always start from an existing VDS base.

---

Patch to an issue - the namespacing of `seq_type_subdir` in MultiCohorts has a condition which doesn't need to be there. This returns '' if the Dataset.cohort is None.

https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/targets.py#L450-L456

The analysis_dataset of a MultiCohort is always created without a Cohort ([line here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/targets.py#L155)), so this is always None, meaning that in CPG_workflows, all genome and exome results from a MultiCohort are namespaced into the same folder.

This condition has already been refactored out in cpg-flow

https://github.com/populationgenomics/cpg-flow/blob/main/src/cpg_flow/targets/helpers.py#L12-L17

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.2
+current_version = 1.36.3
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.36.2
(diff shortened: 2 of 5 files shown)


EddieLF (approved):

MattWellie:
I'll merge this once the current Exome Combiner run has completed - if I merge it now I'll need to re-densify the VDS to run the rest of the pipeline

PR #1136, opened 2025-02-13
Update config option for combiner es-index datasets

EddieLF:
Small change that fixes the config option that determines which datasets to create an es-index for, to align with the seqr_loader pipeline. No bumpversion because it's so minor but I could add one.


MattWellie (approved):
Man what the hell was I doing... using `write_mt` for both but stopping the pipeline at AnnotateDataset... sigh

I'd be Ok with a bump version, but until we enforce unique image tags it's not essential

PR #1137, opened 2025-02-13
Hard check on vds contents

MattWellie:
See https://batch.hail.populationgenomics.org.au/batches/590461

Looks like some of the analysis entries for the VDSs we have on file suggest they include more SGs than they actually do. 

This check doesn't trust the metamist records, instead guaranteeing the VDS contents by checking directly.

Hopefully this is all just a teething problem


EddieLF (approved):
cool fix

PR #1139, opened 2025-02-18
Re-point images-dev docker images

MattWellie:
It's me, hi, I'm the problem it's me.

These are the two instances I know of where an `images-dev` Docker image is being used in routine work.

The GATK-SV one was being trialed before a merge to main in images, and the image path was never updated after it was tested successfully. The main and `-dev` images are from the same Dockerfile, and contain the same patch.

The ClinvArbitration one was built from [this branch](https://github.com/populationgenomics/images/pull/154) when there was no consensus on how/if it should be moved forward. The same image is now built with a different name (bcftools_120), but is not needed in this pipeline. Going forward I'm hoping this pipeline will be run outside of production-pipelines (see [here](https://github.com/populationgenomics/ClinvArbitration/pull/14))

### DIFF
--- a/configs/defaults/clinvarbitration.toml
+++ b/configs/defaults/clinvarbitration.toml
@@ -1,6 +1,3 @@
 [workflow]
 sequencing_type = 'genome'
 site_blacklist = ["victorian clinical genetics services, murdoch childrens research institute"]
-
-[images]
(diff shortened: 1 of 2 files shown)


MattWellie:
changed elsewhere

PR #1141, opened 2025-02-18
Update ICA configuration 

katiedelange:
Update the ICA pipeline configuration to include additional QC report flags, hard trimming of PolyG artefacts, CYP2D6 calling and CNV self-normalisation. This brings our pipeline largely inline with the configuration used by UK Biobank and All of Us, as described here: https://developer.illumina.com/dragen/dragen-popgen

The specific differences between us and the published configurations are that we do not run STR calling (in favour of running ExpansionHunter via our internal pipeline with a custom catalog) and we only produce gVCF outputs (not gVCF and VCF).


configs/defaults/realign_genotype_with_dragen.toml
```diff
@@ -41,6 +41,11 @@ dragmap = 'fil.159e94790e5c4bdfcbef08dd148ff548'
...
 gatk = 'fil.7e16c32f65e9468d31b908dcefab1cbc'
 
+[ica.qc]
+cross_cont_vcf = 'fil.1a7a2d0442854127e5d608da2935b1b0'
+coverage_region_1 = 'fil.ad60897bc97e4646f28c08da33ba0a20'
+coverage_region_2 = 'fil.c4ade57f5ffe4baff28d08da33ba0a20'
+
 [images]
 # This image contains the ICA CLI used for transferring data from ICA -> GCP at the end of the pipeline.
 ica = "australia-southeast1-docker.pkg.dev/cpg-common/images/ica:2.31.0"
```
Alexander-Stuckey, line 51: Is this still the default image, or is there a newer version?

michael-harper:
```suggestion
ica = "australia-southeast1-docker.pkg.dev/cpg-common/images/ica:2.31.2"
```

michael-harper (approved):
This looks good and has worked [in this batch run](https://batch.hail.populationgenomics.org.au/batches/590776)

PR #1143, opened 2025-02-21
Allow user to specify the GatherSampleEvidence jobs to run

EddieLF:
Updates the GatherSampleEvidence stage to allow for the user to specify the jobs that should be run using a new config option `workflow.GatherSampleEvidence.only_jobs`. 

This option will also determine the outputs that are collected from the cromwell server and copied to GCP + the analyses written to Metamist.

To do this, we need to
1. Stop the jobs not in the `only_jobs` list from running as part of the Cromwell workflow
2. Stop the Hail batch driver job from submitting "collect outputs" jobs for the Cromwell jobs which are not running

---

### 1. Stop certain cromwell jobs from running within the GatherSampleEvidence workflow

The GatherSampleEvidence stage runs the [GatherSampleEvidence.wdl](https://github.com/populationgenomics/gatk-sv/blob/main/wdl/GatherSampleEvidence.wdl) workflow in Cromwell. 
This workflow will submit a bunch of Cromwell jobs for a sample, namely:
- coverage counts
- 'pesr'  jobs
- scramble calling
- whamg calling
- manta calling

These jobs can be set to not run if the workflow input is changed. 
- [The coverage counts and pesr jobs can be turned off by setting the evidence collection flags to `False`](https://github.com/populationgenomics/gatk-sv/blob/main/wdl/GatherSampleEvidence.wdl#L26)
  - i.e. `collect_coverage = false`, `collect_pesr = false`
- [The caller jobs can be turned off by unsetting each of the `{caller}_docker` variables](https://github.com/populationgenomics/gatk-sv/blob/main/wdl/GatherSampleEvidence.wdl#L148) 
  - i.e. `scramble_docker = null`, `wham_docker=null`, ... 
  - Note that `melt_docker` is undefined in our pipeline, hence that caller's jobs are never submitted

The workflow input can be changed by altering the `input_dict` defined in the cpg_workflows stage.

### 2. Don't try to collect outputs for jobs which didn't run

The cpg_workflows code will submit "collect outputs" jobs to the Hail batch run that submits and watches the cromwell workflow. These "collect outputs" jobs also need to be disabled if the related cromwell job is not running. 

To stop the "collect outputs" jobs from running, the stages `expected_outputs` dictionary must be updated before the jobs are submitted. 

---

A simple code block which parses the `only_jobs` config list can be used to update both the `input_dict` and the `expected_outputs`, effectively limiting the Cromwell workflow to just the jobs specified. 

---

Test batch running only Manta: https://batch.hail.populationgenomics.org.au/batches/590770

### DIFF
--- a/cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
+++ b/cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
@@ -35,6 +35,7 @@ def get_sv_callers():
         callers = [caller for caller in SV_CALLERS if caller in only_jobs]
         if not callers:
             raise ValueError('No SV callers enabled')
+        return callers
     return SV_CALLERS
 
 
--- a/cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
+++ b/cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
@@ -30,7 +30,18 @@ from cpg_workflows.workflow import (
 )
 
 
-@stage(analysis_keys=[f'{caller}_vcf' for caller in SV_CALLERS], analysis_type='sv')
+def get_sv_callers():
+    if only_jobs := config_retrieve(['workflow', 'GatherSampleEvidence', 'only_jobs']):
+        callers = []
+        for job in only_jobs:
+            for caller in SV_CALLERS:
+                if caller in job:
+                    callers.append(caller)
+        return callers
+    return SV_CALLERS
+
+
+@stage(analysis_keys=[f'{caller}_vcf' for caller in get_sv_callers()], analysis_type='sv')
 class GatherSampleEvidence(SequencingGroupStage):
     """
     https://github.com/broadinstitute/gatk-sv#gathersampleevidence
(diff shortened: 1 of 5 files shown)


cpg_workflows/stages/gatk_sv/gatk_sv_single_sample.py
```diff
@@ -116,6 +125,31 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
+                    caller = key.removesuffix('_docker')
+                    input_dict[key] = val if caller in only_jobs else None
+
+            # remove the expected outputs for the jobs that are not in only_jobs
+            new_expected = {}
+            for job in only_jobs:
+                for key, path in expected_d.items():
+                    if job in key:
+                        new_expected[key] = path
+            expected_d = new_expected
```
MattWellie, line 151: You can move this logic up into expected_outputs() for consistency - i.e. only return the outputs dictionary with the outputs we intend to generate in this run. That will mean that the `expected_outputs()` is always limited to the things we actually intend to run, which will be used in the DAG generation.

e.g. 
- only_jobs = ['manta']
- manta VCF already exists
- expected_outputs -> includes pesr, coverage and manta
- workflow.py spots that pesr.vcf doesn't exist, so it runs queue_jobs
- in queue_jobs we tell it to only run manta, and launch the WDL
- WDL runs, manta is regenerated, overwrites the manta file we already had

MattWellie (changes requested):
I think there's a little polishing in expected outputs which would make this consistent throughout

MattWellie (approved):
🤞

PR #1144, opened 2025-02-21
Add None as default for config_retrieve GatherSampleEvidence.only_jobs

EddieLF:
Hotfix to add a default to the config retrieve added in #1143 

Since without a default, any workflow fails... even the non gatk-sv workflows. Woops.
https://batch.hail.populationgenomics.org.au/batches/590777/jobs/1


MattWellie (approved):
Oops 😅

PR #1145, opened 2025-02-24
Cast the AD values to a comma separated string

EddieLF:
Makes the AD values exported to the annotated dataset consistent with seqr's expected format of a comma separated string, like the example data [here](https://github.com/broadinstitute/seqr/blob/cb32e5b5c490e9dd24f1961fad3be64047205bd8/seqr/fixtures/1kg_project.json#L1822).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.8
+current_version = 1.36.9
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


MattWellie (approved):
approving this change, but potentially holding on a sec to see why these aren't all just the normalised ADs

MattWellie:
Alright, from a little investigation it looks like the dodgy AD values are an artefact of the pipeline development phase, meaning that the data we have in -test for the validation cohort is not representative of the current pipeline. 

We can re-run the pipeline in full, but IMO that shouldn't necessarily hold this change back

PR #1146, opened 2025-02-25
Fix ICA upload and download commands

michael-harper:
## Issue 1:
[During the upload](https://batch.hail.populationgenomics.org.au/batches/590809/jobs/3) of crams to Batch VMs we are seeing `"No space left on device"` for some jobs. We think that changing the upload directory to be `/io` will avoid these storage issues.

## Issue 2:
Additionally, the cram files in GCP do not have md5 hashes calculated and so comparing the ICA md5sum to the once on GCP returns an error and the file still gets downloaded but the job fails and is not registered to metamist. Consulting the `gcloud storage` documentation [here](https://cloud.google.com/storage/docs/data-validation#cli) we see:

> For the [Google Cloud CLI](https://cloud.google.com/sdk/gcloud/reference/storage), data copied to or from a Cloud Storage bucket is validated. This applies to cp, mv, and rsync commands. If the checksum of the source data does not match the checksum of the destination data, the gcloud CLI deletes the invalid copy and prints a warning message.

So we are only performing md5sum matching after downloading from ICA via `icav2` and not when copying from the local directory to GCP.

### DIFF
--- a/cpg_workflows/stages/realign_genotype_with_dragen.py
+++ b/cpg_workflows/stages/realign_genotype_with_dragen.py
@@ -120,11 +120,11 @@ class UploadDataToIca(SequencingGroupStage):
             cram_status=$(icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram --match-mode EXACT -o json | jq -r '.items[].details.status')
             if [[ $cram_status != "AVAILABLE" ]]
             then
-                mkdir /io/{sequencing_group.name}
-                gcloud storage cp {sequencing_group.cram} /io/{sequencing_group.name}/
-                gcloud storage cp {sequencing_group.cram}.crai /io/{sequencing_group.name}/
-                icav2 projectdata upload /io/{sequencing_group.name}/{sequencing_group.name}.cram /{bucket}/{upload_folder}/{sequencing_group.name}/
-                icav2 projectdata upload /io/{sequencing_group.name}/{sequencing_group.name}.cram.crai /{bucket}/{upload_folder}/{sequencing_group.name}/
+                mkdir $BATCH_TMPDIR/{sequencing_group.name}
+                gcloud storage cp {sequencing_group.cram} $BATCH_TMPDIR/{sequencing_group.name}/
+                gcloud storage cp {sequencing_group.cram}.crai $BATCH_TMPDIR/{sequencing_group.name}/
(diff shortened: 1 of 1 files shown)


cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -120,11 +120,11 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
             cram_status=$(icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram --match-mode EXACT -o json | jq -r '.items[].details.status')
             if [[ $cram_status != "AVAILABLE" ]]
             then
-                mkdir {sequencing_group.name}
-                gcloud storage cp {sequencing_group.cram} .
-                gcloud storage cp {sequencing_group.cram}.crai .
-                icav2 projectdata upload {sequencing_group.name}.cram /{bucket}/{upload_folder}/{sequencing_group.name}/
-                icav2 projectdata upload {sequencing_group.name}.cram.crai /{bucket}/{upload_folder}/{sequencing_group.name}/
+                mkdir /io/{sequencing_group.name}
+                gcloud storage cp {sequencing_group.cram} /io/{sequencing_group.name}/
```
jmarshall, line 124: Currently `/io/` does not appear in any production-pipelines code at all. I'd recommend `$BATCH_TMPDIR/` instead.

michael-harper: Of course! Thanks @jmarshall

MattWellie:
FYI this has been tried before, and works... most of the time, but not always
https://centrepopgen.slack.com/archives/C030X7WGFCL/p1713828954747299?thread_ts=1709094928.689809&cid=C030X7WGFCL

michael-harper:
It's definitely `$BATCH_TMPDIR` right? Or is it `$BATCH_TMP_DIR` as Franklin is mentioning in that thread?

jmarshall:
It's definitely `$BATCH_TMPDIR`.

It'd be worth adding `pwd` and `df --si` as well, to get some logs of what the situation is when the job starts up. As Matt said and you observed, there seems to be an intermittent problem here, so this information would be helpful if the problem occurs.

michael-harper:
FYI working batch [here](https://batch.hail.populationgenomics.org.au/batches/590880)

Alexander-Stuckey (approved):
Looks good to me

PR #1151, opened 2025-03-07
Add option to subset matrixtable to only certain families and SGs

EddieLF:
- Add a config option for RD combiner `workflow.dataset.only_families`
  - populate the `only_families` list with the internal ~or external~ IDs of families to keep in the subset
- Find the sequencing groups within this list of families by ~querying metamist for them~ checking each sequencing group's `pedigree.fam_id` attribute
- Write a new matrix table subsetted from the cohort matrix table for just these SGs
- Include this family subsetting logic in the `expected_outputs` for the `SubsetMatrixTableToDatasetUsingHailQuery`, `AnnotateDatasetSmallVariantsWithHailQuery`, and `ExportMtAsEsIndex` stages for consistency
- Also modify the names of the outputs to include the SG count, family count, and SG hash in the output filenames, as a way to differente them from the outputs that include the full dataset (or different family subsets)

Subsetting a cohort mt to a single family from a test dataset: https://batch.hail.populationgenomics.org.au/batches/591428

### DIFF
--- a/cpg_workflows/stages/rd_combiner.py
+++ b/cpg_workflows/stages/rd_combiner.py
@@ -156,7 +156,6 @@ def get_family_sequencing_groups(dataset: Dataset) -> dict:
     Get the subset of sequencing groups that are in the specified families for a dataset
     Returns a dict containing the sequencing groups and a name suffix for the outputs
     """
-    dataset_name = dataset.name (+ '-test' if config_retrieve(['workflow', 'access_level']) == 'test' else '')
     only_family_ids = set(config_retrieve(['workflow', dataset.name, 'only_families'], []))
     # keep only the SG IDs for the families in the only_families list
     family_sg_ids = [sg.id for sg in dataset.get_sequencing_groups() if sg.pedigree.fam_id in only_family_ids]
--- a/cpg_workflows/stages/rd_combiner.py
+++ b/cpg_workflows/stages/rd_combiner.py
@@ -158,6 +158,9 @@ def get_family_sequencing_groups(dataset: Dataset) -> dict:
     """
     only_family_ids = set(config_retrieve(['workflow', dataset.name, 'only_families'], []))
     # keep only the SG IDs for the families in the only_families list
+    get_logger.info(f'Finding sequencing groups for families {only_family_ids} in dataset {dataset.name}')
+    for sg in dataset.get_sequencing_groups():
+        print(sg.id, sg.pedigree.fam_id)
     family_sg_ids = [sg.id for sg in dataset.get_sequencing_groups() if sg.pedigree.fam_id in only_family_ids]
     if not family_sg_ids:
         raise ValueError(f'No sequencing groups found for families {only_family_ids} in dataset {dataset.name}. Available sequencing groups: {[sg.id for sg in dataset.get_sequencing_groups()]}')
--- a/cpg_workflows/stages/rd_combiner.py
+++ b/cpg_workflows/stages/rd_combiner.py
@@ -54,26 +54,6 @@ LATEST_ANALYSIS_QUERY = gql(
 """,
 )
 
-FAMILY_SGS_QUERY = gql(
-    """
-    projectFamilySequencingGroups($dataset: String!, $sequencing_type: String!) {
-        project(name: $dataset) {
-            families {
-                id
-                externalId
-                participants {
-                    samples {
-                        sequencingGroups(type: {eq: $sequencing_type}) {
-                            id
-                        }
-                    }
-                }
-            }
-        }
-    }
-    """,
-)
-
 SPECIFIC_VDS_QUERY = gql(
     """
     query getVDSByAnalysisId($vds_id: Int!) {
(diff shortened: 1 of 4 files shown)


cpg_workflows/stages/rd_combiner.py
```diff
@@ -591,22 +642,41 @@ def queue_jobs(self, multicohort: MultiCohort, inputs: StageInput) -> StageOutpu
...
-        return {
-            'mt': dataset.tmp_prefix() / 'mt' / self.name / f'{get_workflow().output_version}-{dataset.name}.mt',
-            'id_file': dataset.tmp_prefix() / f'{get_workflow().output_version}-{dataset.name}-SG-ids.txt',
-        }
+        if only_families := config_retrieve(['workflow', dataset.name, 'only_families'], []):
+            family_sg_data = query_for_family_sequencing_groups(
+                dataset.name,
+                config_retrieve(['workflow', 'sequencing_type']),
+                only_families,
+            )
```
MattWellie, line 664: I think this use case should come before the one above, otherwise we'll never be able to apply this family-subset logic to single-cohort callsets like VCGS

That will mean we also need to apply the same decision downstream when deciding whether to use input from AnnotateCohort or the Subset step, even for single-project callsets

cpg_workflows/stages/rd_combiner.py
```diff
@@ -622,11 +692,21 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput:
...
+            )['family_sg_ids']
+        else:
+            family_sg_ids = None
+
         # write a list of all the SG IDs to retain
         if not config_retrieve(['workflow', 'dry_run'], False):
             with outputs['id_file'].open('w') as f:
                 for sg in dataset.get_sequencing_groups():
-                    f.write(f'{sg.id}\n')
+                    if not family_sg_ids or sg.id in family_sg_ids:
```
MattWellie, line 708:
```suggestion
                    if (family_sg_ids is None) or (sg.id in family_sg_ids):
```

It's possible with the current logic to provide a list of families, then fail to map them to any SGs (family_sg_ids is empty). In that case `not family_sg_ids` still evaluates to True, so we keep all SGs.

I think a warning (probably quitting pipeline setup) would be valid if we give a specific family subset and can't find any SGs for them, given that we tried to acquire a specific subset, but won't achieve that.

cpg_workflows/stages/rd_combiner.py
```diff
@@ -652,7 +732,22 @@ def expected_outputs(self, dataset: Dataset) -> Path:
...
+                only_families,
+            )
+            n_sgs = len(family_sg_data['family_sg_ids'])
+            n_families = len(family_sg_data['only_family_ids'])
+            return (
+                dataset.prefix()
+                / 'mt'
+                / self.name
+                / f'{get_workflow().output_version}-{dataset.name}-{n_sgs}_sgs-{n_families}_families.mt'
+            )
```
MattWellie, line 748: If we carry out this same process twice, each for a different trio from the same callset, the `{dataset.name}-{n_sgs}_sgs-{n_families}_families.mt` portion of the filename would be identical, so we wouldn't repeat the Stage.


Can we incorporate a partial hash of the CPG IDs in this to make sure we're keeping these outputs in parallel? Borrowing from [here](https://github.com/populationgenomics/production-pipelines/blob/20bfbf9b60dde97b942891bc96e4a5bde43328e7/cpg_workflows/targets.py#L82-L91) we could do

```suggestion
            import hashlib
            sg_ids = family_sg_data['family_sg_ids']
            h = hashlib.sha256(''.join(sorted(sg_ids).encode()).hexdigest()[:4]
            n_sgs = len(sg_ids)
            n_families = len(sg_ids)
            return (
                dataset.prefix()
                / 'mt'
                / self.name
                / f'{get_workflow().output_version}-{dataset.name}-{n_sgs}_sgs-{n_families}_{h}_families.mt'
            )
```

cpg_workflows/stages/rd_combiner.py
```diff
@@ -704,7 +799,17 @@ def expected_outputs(self, dataset: Dataset) -> dict[str, str | Path]:
...
-        index_name = f'{dataset.name}-{sequencing_type}-{get_workflow().run_timestamp}'.lower()
+        if only_families := config_retrieve(['workflow', dataset.name, 'only_families'], []):
+            family_sg_data = query_for_family_sequencing_groups(
+                dataset.name,
+                config_retrieve(['workflow', 'sequencing_type']),
+                only_families,
+            )
+            n_sgs = len(family_sg_data['family_sg_ids'])
+            n_families = len(family_sg_data['only_family_ids'])
+            index_name = f'{dataset.name}-{sequencing_type}-{n_sgs}_sgs-{n_families}_families-{get_workflow().run_timestamp}'.lower()
```
MattWellie, line 810: likewise here, including some fragment of a hash in case we ever want to export multiple families separately from the same callset

MattWellie (changes requested):

MattWellie (approved):
Looks rad as hell

Issue #1153, opened 2025-03-13
Check for existence of previous outputs


MattWellie:
https://github.com/populationgenomics/production-pipelines/blob/48dd2194571359493e90c29ede21ef481d534c72/cpg_workflows/stages/gatk_sv/gatk_sv_multisample.py#L230-L249

Because this stage is disconnected from the prior stage which creates the files, we don't guarantee that these files exist. 

Add a check that these files exists for real, otherwise don't start the workflow (e.g. [here](https://batch.hail.populationgenomics.org.au/batches/591320/jobs/2) failing once the file existence was required, 14 hours in)

PR #1158, opened 2025-03-17
Validate sv multisample inputs

EddieLF:
Resolves https://github.com/populationgenomics/production-pipelines/issues/1153

The `gatk_sv_multisample` workflow uses the outputs generated in the `gatk_sv_single_sample` workflow as inputs. 

Since these are two disconnected workflows, the outputs are not explicitly checked for existence as the stages are not explicitly linked.

This adds an explicit path existence check for every input path before the multisample workflow starts. If any inputs are not found, they will be printed in the logs and an error is raised.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.11
+current_version = 1.36.12
 commit = True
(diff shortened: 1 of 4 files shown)


MattWellie (approved):
LGTMatt

PR #1166, opened 2025-03-24
gnomAD Joint Freq/annotations into MT

MattWellie:
Adds the gnomAD joint data in during the AnnotateCohort Stage (of the combiner pipeline). Keen to grab some input on the exact fields we want to pull over here. 

I've left a lengthy docstring to explain the process used to find and display these data. I couldn't find a README on how to use the gnomAD HT, but I was remembering a conversation with Ben B in the Seqr team about how they stored keys in the globals dict and used that as an index to unpack the values in each row as a compression technique. 

Here I'm pulling out the `adj` (quality-adjusted calls across all popuations) as the per-allele AC/AN/AF/Hom values, then using the max-population FAF and AF into separate fields to match the ones we're already extracting from exac/gnomad2/3

I'm also trying to pull out 'Hemi', which should be missing except for X/Y chrom variants. I might need to put some flexibility here for 'Hemi if defined, else missing'.


MattWellie:
Cool, I think this is now good for a review - the gnomad4 frequencies are accurate (spot-checked against the gnomad website), and the latest change replaces the existing `mt.gnomad_genomes` with the V4 data. We still have `mt.gnomad_exomes` (V2, from the seqr reference data), and `mt.gnomad_genome_coverage` (also from the reference data... the gnomad V4 table doesn't have this attribute, and the V4 release only seems to have a coverage table for Exomes?)

MattWellie:
This also deletes a test which was submitting 100+ python jobs in a chain to try and hit a niche recursion error. That's probably not something we need a test case for, and it was the only breaking test

jmarshall:
The weird test problem is already addressed by #1240.

PR #1167, opened 2025-03-24
update pipeline error handling case

Alexander-Stuckey:
Remove the extra case where it only checks if we have set the resume flag, without having a pipeline ID. We need both pieces of information to resume, otherwise we need to start another pipeline.


michael-harper (approved):
Reviewed with @Alexander-Stuckey and agree not a case we should be catching

PR #1168, opened 2025-03-24
Gatk sv multisample disable qc

EddieLF:
Adds a new default config which will prevent the various QC plotting subworkflows from running during the MakeCohortVcf stage. These plots are unused so there's no need to run them.


MattWellie (approved):
I like

PR #1170, opened 2025-03-25
Update Hail Version

MattWellie:
I've just double checked, and the latest hail-driver image is still on 0.2.133, are we ready to release a 0.2.134 version?

[2 weeks ago](https://github.com/populationgenomics/analysis-runner/actions/runs/13845799761) was the last run of Analysis-runner's deploy server workflow, which coincided with the release of [Hail 0.2.143.cpg1](https://github.com/populationgenomics/hail/releases/tag/0.2.134.cpg1), but it looks like the Hail SHA wasn't updated - that workflow installed 0.2.133 into the driver image.

If the driver image had been updated to 134, the strict pin here would have reinstalled 133 over the top anyway.

Once the driver image version is updated, this should be updated (probably to hail==0.2.134.cpg1 to be specific)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.15
+current_version = 1.36.16
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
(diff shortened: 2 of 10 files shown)


vivbak (approved):

vivbak:
As prod pipes doesn't have strict version pinning and will soon be deprecated anyhow, I'm fine with the more flexible > pin. To be discussed with @violetbrina in the case of CPG Flow

PR #1171, opened 2025-03-25
Talos prep to dataset stages

MattWellie:
It makes sense to have this be a CohortStage workflow, but at the moment it's impossible to parallelise this. To generate each project's data correctly formatted, and the resulting analysis entry saved to the correct dataset, this has to be done by starting each cohort in a separate workflow, which is unpleasant.

When https://cpg-populationanalysis.atlassian.net/browse/SET-490 is closed this will be easier, but we can get the same functionality now (albeit with a bit less of a same-cohort guarantee) by reverting these down to Dataset Stages

Example use case - run rd_combiner in full, then use the same config in talos_prep to generate the talos-specific outputs. The Cohorts would be identical, and rd_combiner already contains DatasetStages so this is more uniform

I've had to start about 60 separate runs to operate this workflow, adn we're about to have a fresh round of raw data, I'm tired boss


cassimons (approved):
Sounds like a good plan. Sanity is to be protected - it is a limited resource.

PR #1172, opened 2025-03-25
Compress exomiser results

MattWellie:
Exomiser workflow generates a Tarball'd Hail Table instead of just a Hail Table

Talos is the only consumer of this Hail Table, and I want to completely separate Talos from GCloud.

This file can be read in as a regular input file, rather than a massive directory


EddieLF (approved):
LGTM

PR #1173, opened 2025-03-25
Add VQSR headers during MT Extraction

MattWellie:
We had an issue where during the MT -> VCF -> Annotation -> MT loop in talos_prep, the VCF being exported didn't contain the header description lines for all VQSR FILTERS. This was tracked to AnnotateCohort, where we join the MatrixTable with the VQSR VCF. During this join we take the INFO fields and FILTERS from the VQSR VCF, but not the corresponding headers. This means there are FILTER entries in the MT which have no description line. 

In Hail this is fine, but bcftools and echtvar both have issues processing data with incomplete headers. To get over this speedbump I've been removing all FILTER-flagged variants in the Talos-prep workflow. To add those variants back in, we need the `talos_prep` VCF to include a description header line for every FILTER. This can be done by adding a `metadata` dictionary to the `hl.export_vcf` method. Oh... but how to get that dictionary you ask? Glad you did!

1. Hail has a method [get_vcf_metadata](https://hail.is/docs/0.2/methods/impex.html#hail.methods.get_vcf_metadata), which is pointed directly at a VCF file, grabbing a nested dictionary of the VCF contents:

```
all_headers = hl.get_vcf_metadata('gs://cpg-seqr-main/rd_combiner/.../vqsr_snps_and_indels.vcf.gz')
```

we can use this method, but that involves passing both the MatrixTable and VQSR paths into the `talos_prep` Stage. The VQSR file path isn't recorded in metamist, and as `talos_prep` and `rd_combiner` are separate workflows it would be tricky to make sure you're accessing the paired VQSR VCF and Variant MT.

2. We can just hard-code the FILTERS block from a VQSR VCF. AFAIK these tranches are an internal feature of VQSR - we aren't manually setting them via config. So long as we're using VQSR these will be the tranches in our data, and this maintains a full separation between the two workflows. During this MT -> VCF extraction the additional VQSR header descriptions are added, and the resulting VCF will have a fully defined header.

There are other INFO fields we're taking from VQSR, but these are being thrown away here anyway to make sure we have a clean slate for annotation. As a result these don't cause any parsing issues downstream. 

I've gone with option 2, it solve the problem quick and dirty. It's easier to do this than pull the FILTER header block from a specific individual VCF.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.15
+current_version = 1.36.16
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.36.15
+  VERSION: 1.36.16
 
 jobs:
   docker:
--- a/cpg_workflows/scripts/talos_prep/extract_fragmented_vcf_from_mt.py
+++ b/cpg_workflows/scripts/talos_prep/extract_fragmented_vcf_from_mt.py
@@ -23,7 +23,6 @@ from cpg_utils.config import config_retrieve, genome_build
 from cpg_utils.hail_batch import init_batch
 from cpg_workflows.utils import get_logger
 
-
 # VQSR filters are added to the VQSR VCF, but we don't retain these header lines when we copy the VQSR info over
 # to the variant MatrixTable during AnnotateCohort.
 # If we export VCFs here without these header lines, there's a possibility that downstream tools won't be able to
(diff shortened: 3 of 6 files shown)


EddieLF:
Regarding option 1. 

Is the VCF created by the `RunTrainedIndelVqsrOnCombinedVcf` usable for the headers? It does get written to Metamist in the rd_combiner pipeline e.g. [here](https://batch.hail.populationgenomics.org.au/batches/593015/jobs/120)

EddieLF (approved):
The hardcoding seems fine though if it's just those 6 specifically named filter fields

MattWellie:
> Regarding option 1.
> 
> Is the VCF created by the `RunTrainedIndelVqsrOnCombinedVcf` usable for the headers? It does get written to Metamist in the rd_combiner pipeline e.g. [here](https://batch.hail.populationgenomics.org.au/batches/593015/jobs/120)

You're quite right, I was looking at the stage above it 🤦 The VQSR VCF path is registered in metamist. 

https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/rd_combiner.py#L501

I reckon I could still make a case for hard-coding the contents based on them not changing, and this being a Dataset-workflow as opposed to a MultiCohort VQSR registration - you'd have to know which project to query to find it

PR #1175, opened 2025-03-26
make ica even more robust

Alexander-Stuckey:
Add a retry loop to try fetch file IDs from ICA, fail if the file is still empty after 10 attempts.

### DIFF
--- a/cpg_workflows/stages/realign_genotype_with_dragen.py
+++ b/cpg_workflows/stages/realign_genotype_with_dragen.py
@@ -127,7 +127,7 @@ class UploadDataToIca(SequencingGroupStage):
 
                 function get_fids {{
                 # Add a random delay before calling the ICA API to hopefully stop empty JSON files from being written to GCP
-                sleep $((shuf -i 1-30 -n 1))
+                sleep $(shuf -i 1-30 -n 1)
                 icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram --match-mode EXACT -o json | jq -r '.items[].id' > cram_id
                 icav2 projectdata list --parent-folder /{bucket}/{upload_folder}/{sequencing_group.name}/ --data-type FILE --file-name {sequencing_group.name}.cram.crai --match-mode EXACT -o json | jq -r '.items[].id' > crai_id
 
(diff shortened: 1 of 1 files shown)


cpg_workflows/stages/realign_genotype_with_dragen.py
```diff
@@ -123,19 +123,38 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
                     icav2 projectdata upload $BATCH_TMPDIR/{sequencing_group.name}/{sequencing_group.name}.cram /{bucket}/{upload_folder}/{sequencing_group.name}/
                     icav2 projectdata upload $BATCH_TMPDIR/{sequencing_group.name}/{sequencing_group.name}.cram.crai /{bucket}/{upload_folder}/{sequencing_group.name}/
                 fi
+                }}
 
+                function get_fids {{
                 # Add a random delay before calling the ICA API to hopefully stop empty JSON files from being written to GCP
-                sleep $((random % 30))
+                sleep $((shuf -i 1-30 -n 1))
```
MattWellie, line 130: this syntax doesn't work for me in a docker-ubuntu box. A single-parenthes statement works
```
root@44b3ce1ba2a3:/# sleep $((shuf -i 1-30 -n 1))
  > bash: shuf -i 1-30 -n 1: syntax error in expression (error token is "1-30 -n 1")
root@44b3ce1ba2a3:/# sleep $(shuf -i 1-30 -n 1)
root@44b3ce1ba2a3:/#
```

Alexander-Stuckey: You can see that `random` worked here with double parantheses https://batch.hail.populationgenomics.org.au/batches/593045/jobs/728

MattWellie: This was a fun little detour...

- random is a utility at `/dev/random`, so i guess `/dev` is in the PATH in that container, but not in my local one
- $RANDOM is an internal Bash function to do the same, that did work for me: [here](https://tldp.org/LDP/abs/html/randomvar.html)

`sleep  $(($RANDOM % 30))` worked for me, but I think that's what you had originally in this script... and this _did_ require double brackets 🤷

MattWellie (approved):
LGTMatt. Never used `shuf` before, but in AS we trust

PR #1179, opened 2025-03-27
general improvements to ica pipeline

Alexander-Stuckey:
- Disable the `Dragen report` step in ICA, as this has been the cause of some transient errors in ICA.
- Delete pipeline run ID files in the case of a failure in ICA, as we are unable to recover if this happens. This saves having to perform this step manually afterwards.
- Make the ICA data upload stage more resilient, and guard against the `crai` not being uploaded, as this has happened in a handful of cases.
- Use `$BATCH_TMPDIR` for download of data as well, as there have additionally been a handful of cases of running out of disk space, as the job was not properly using this space.

Have not yet implemented a running log for failed ICA pipeline runs, but that may take more time than anticipated, and want to have the above in a working stage so that the pipeline can run over the weekend.

### DIFF
--- a/cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
+++ b/cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
@@ -64,5 +64,5 @@ def run(
             # Log failed ICA pipeline to a file somewhere
             # Delete the pipeline ID file
             logging.info(f'Deleting the pipeline run ID file {pipeline_id_file}')
-            subprocess.run(['gcloud', 'storage', 'rm', f'{pipeline_id_file}'])
+            subprocess.run(['gcloud', 'storage', 'rm', pipeline_id_file])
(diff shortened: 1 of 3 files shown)


cpg_workflows/stages/dragen_ica/monitor_align_genotype_with_dragen.py
```diff
@@ -60,4 +61,8 @@ def run(
         elif pipeline_status in ['ABORTING', 'ABORTED']:
             raise Exception(f'Pipeline run {pipeline_id} has been cancelled.')
         else:
+            # Log failed ICA pipeline to a file somewhere
+            # Delete the pipeline ID file
+            logging.info(f'Deleting the pipeline run ID file {pipeline_id_file}')
+            subprocess.run(['gcloud', 'storage', 'rm', f'{pipeline_id_file}'])
```
MattWellie, line 67:
```suggestion
            subprocess.run(['gcloud', 'storage', 'rm', pipeline_id_file])
```

MattWellie (approved):
seems reasonable

PR #1185, opened 2025-03-31
Prepare Coverage Hail Table for Browser

michael-harper:
This PR integrates the `compute_coverage_stats` utility function from the `gnomad_methods` submodule into the Large Cohort pipeline, enabling efficient coverage table generation for display in the OurDNA Browser.

## 1. Generate Reference Table and Calculate Coverage
As suggested in [this comment below by @katiedelange ](https://github.com/populationgenomics/production-pipelines/pull/1185#discussion_r2170119089) we are now calculating coverage over the calling intervals used in genotyping. The number of intervals used is determined by the workflow's `scatter_count`. The implementation will automatically detect the workflow's `sequencing_type` and pull the correct `.interval_list` accordingly. The subsequent `reference_ht` is built using these scattered interval files. Additionally, to avoid excessive I/O and memory costs we only read in the desired VDS at those specific intervals.

#### Reason for building a reference Hail Table
- The VDS object does not have per-base granularity inside the `vds.variant_data` Matrix Table. Because we are wanting to calculate coverage across every position in the genome (in WGS sequencing) we must artificially add 'variant sites' at every position in the genome inside the `vds.variant_data` Matrix Table. Thus, during densification we can pull the DP and GT fields from the `vds.reference_data`. 
- The `reference_ht` table therefore includes all loci in the regions of interest (i.e., every base pair in the genome for WGS).

Coverage metrics are calculated for each VDS shard using the `gnomad_methods` coverage functions.

Each shard is processed independently, allowing parallel computation.

## 2. Merge Coverage Tables
Coverage tables generated per shard are merged to produce the final coverage table for the entire genome.

## Scalability Considerations
The primary scalability factor is the number of genomic sites (rows) multiplied by the number of samples (columns). When we are extending the `vds.variant_data` to include every site in the `reference_ht` (i.e. ~3 billion sites in WGS sequencing) we expand memory requirements significantly.
```
t = t._localize_entries("__entries", "__cols")
```
This line of code will have it so that each row will store an array of the entries for all columns e.g. 1 million rows and 10,000 samples, each row will have an array of 10,000 entries which may lead to significant memory usage and excessive run times when extended to every site in the genome.

This notion was discussed with several members of the Data Team and the general consensus was that if gnomAD implemented it at scale then we will proceed with the current implementation and address any concerns as they arise. During testing a `shard_size=50_000_000` led to all coverage tables being calculated in ~1 hour.


cpg_workflows/stages/large_cohort.py
```diff
@@ -489,6 +490,195 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+        }
+
+    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
+        from cpg_workflows.large_cohort import generate_coverage_table
+
+        j = get_batch().new_job(
+            'ShardVds',
+            (self.get_job_attrs() or {}) | {'tool': HAIL_QUERY},
+        )
+        j.image(image_path('cpg_workflows'))
```
MattWellie, line 516: best not to use `image_path('cpg_workflows')`, it's rarely the right image. Use `config_retrieve(['workflow', 'driver_image'])`

cpg_workflows/stages/large_cohort.py
```diff
@@ -489,6 +490,195 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
         return self.make_outputs(cohort, data=self.expected_outputs(cohort), jobs=[j])
 
 
+@stage(required_stages=[Combiner])
+class ShardVds(CohortStage):
```
MattWellie, line 494: Is this important? Query on Batch already parallelises all operations by partition, so explicitly pre-sharding this object shouldn't do much, but will double the data we store in GCP

cpg_workflows/stages/large_cohort.py
```diff
@@ -489,6 +490,195 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+                ['large_cohort', 'output_versions', 'exome_reference_coverage'],
+                default=None,
+            )
+        else:
+            ref_cov_version = config_retrieve(
+                ['large_cohort', 'output_versions', 'genome_reference_coverage'],
+                default=None,
+            )
+
+        ref_cov_version = ref_cov_version or get_workflow().output_version
```
MattWellie, line 556:
```suggestion
        sequencing_type = config_retrieve(['workflow', 'sequencing_type'])
        ref_cov_version = config_retrieve(
                ['large_cohort', 'output_versions', f'{sequencing_type}_reference_coverage'],
                default=get_workflow().output_version,
            )

```

cpg_workflows/stages/large_cohort.py
```diff
@@ -489,6 +490,195 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+            )
+            jobs.append(j)
+
+        return self.make_outputs(cohort, data=self.expected_outputs(cohort), jobs=jobs)
+
+
+@stage(required_stages=[Combiner, ShardVds, GenerateReferenceCoverageTable])
+class GenerateCoverageTable(CohortStage):
+    def expected_outputs(self, cohort: Cohort) -> dict[str, Path]:
+        if coverage_version := config_retrieve(['large_cohort', 'output_versions', 'coverage'], default=None):
```
MattWellie, line 590: Just a curiosity here - each of these stages is pulling output_version from a different variable in config, but this pipeline is continuous, so they should all have a single output version?

cpg_workflows/stages/large_cohort.py
MattWellie: This workflow shards at one end, and merges at the other, but QOB already provides parallelised operations :/

cpg_workflows/stages/large_cohort.py
michael-harper: This approach allows us to pull previously expensive stage runs outputs to test/run downstream stages and have those later stages outputs versions versioned differently. This has proven useful recently in running subsets of production data that Josh and Katie have produced to investigate OurDNA data but which we wanted to check would load into the browser.

In general though this approach is seldom used due to inevitably running into provenance issues not knowing which version of later stage output was produced from earlier versions of previous stages outputs.

configs/defaults/large_cohort.toml
```diff
@@ -111,6 +111,9 @@ max_n_singletons = 800000  # maximum number of unique SNPs for a sample
 max_r_duplication = 0.3  # maximum rate of duplicated reads (from picard metrics)
 max_r_het_hom = 3.3  # maximum rate of heterozygous to homozygous SNPs
 
+[large_cohort.references]
+contig_lengths = "gs://cpg-bioheart-test/GRCh38_contig_lengths.json"
```
katiedelange, line 115: If this is going to be the default, it needs to live in references.

cpg_workflows/large_cohort/generate_coverage_table.py
```diff
@@ -0,0 +1,85 @@
...
+    :param start: Start position of the interval.
+    :param end: End position of the interval.
+    :param out_path: Path to save the coverage table.
+    :return: Coverage Hail Table.
+    """
+    from cpg_utils.hail_batch import init_batch
+    from gnomad.utils.reference_genome import add_reference_sequence
+
+    init_batch()
+    rg: hl.ReferenceGenome = hl.get_reference('GRCh38')
```
katiedelange, line 51: Nitpick, as we're definitely using GRCh38 for the foreseeable future, but this should ideally be configurable.

cpg_workflows/stages/large_cohort.py
```diff
@@ -552,6 +554,100 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+                    chrom,
+                    int(start),
+                    int(end),
+                    str(coverage_output_path),
+                    setup_gcp=True,
+                ),
+            )
+            converage_jobs.append(j)
+
+        return self.make_outputs(cohort, data=self.expected_outputs(cohort), jobs=converage_jobs)
```
katiedelange, line 611:
```suggestion
        coverage_jobs = []

        coverage_table_paths = self.expected_outputs(cohort)

        for shard, coverage_output_path in coverage_table_paths.items():
            chrom, start, end = shard.split('_')
            j = get_batch().new_job(
                f'GenerateCoverageTable_{shard}',
                (self.get_job_attrs() or {}) | {'tool': HAIL_QUERY},
            )
            j.image(config_retrieve(['workflow', 'driver_image']))
            j.command(
                query_command(
                    generate_coverage_table,
                    generate_coverage_table.run.__name__,
                    str(inputs.as_path(cohort, Combiner, key='vds')),
                    chrom,
                    int(start),
                    int(end),
                    str(coverage_output_path),
                    setup_gcp=True,
                ),
            )
            coverage_jobs.append(j)

        return self.make_outputs(cohort, data=self.expected_outputs(cohort), jobs=coverage_jobs)
```

cpg_workflows/stages/large_cohort.py
```diff
@@ -552,6 +554,100 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+        }
+
+    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
+        from cpg_workflows.large_cohort import generate_coverage_table
+
+        converage_jobs = []
+
+        coverage_table_paths = self.expected_outputs(cohort)
+
+        for shard, coverage_output_path in coverage_table_paths.items():
```
katiedelange, line 590: Can you just explain the logic of doing all the handling of sharding here, rather than letting hail handle it in `compute_coverage_stats`?

configs/defaults/large_cohort.toml
katiedelange: We will also need different versions for genomes and exomes, as we only need to run this across the calling intervals + a small padding region for exomes (handled by `interval_ht` in `compute_coverage_stats`, but we'll need to do it externally if we want to manage sharding and intervals ourselves).

katiedelange (commented):
Thanks for the really nice write up Harper, super helpful. Some minor comments and one more conceptual question for my own understanding.

cpg_workflows/stages/large_cohort.py
michael-harper: We need to explicitly handle sharding and output path generation at the workflow (Stage) level, rather than inside `compute_coverage_stats`, because the prod-pipes builds the DAG and tracks outputs based on the paths returned by `expected_outputs()`. This means that, for downstream stages (like `MergeCoverageTables`) to know which files to expect and depend on, the sharding logic and output paths must be determined up front, not dynamically within a Hail job.

Additionally, since the shard size is configurable, the number and names of output files can change. By generating all output paths in the Stage, we ensure that the workflow’s dependency graph is accurate and reproducible, and that all expected outputs are known before any jobs are launched.

cpg_workflows/large_cohort/generate_coverage_table.py
michael-harper: Now using `from cpg_utils.hail_batch import genome_build`

configs/defaults/large_cohort.toml
michael-harper: I spoke to Josh last week about this in regards to generating a sites table for exome data. I will work on this today and see if it can be brought in to this PR.

michael-harper: Thanks @katiedelange - I have gone ahead and implemented this now. We are pulling the `scatter_count` from the config that will then be used by `get_intervals()`.

This required some logic change within the `generate_coverage_table` to read `.interval_list` files and work with them to produce reference tables that can be used to calculate coverage over. 

This has worked:
[Here](https://batch.hail.populationgenomics.org.au/batches/626172/jobs/1) (where we generate split tables). The outputs of this batch actually live in `tmp` now so will only be viewable for a week. This batch ran with a bug in expected output paths so a [second batch](https://batch.hail.populationgenomics.org.au/batches/626266/jobs/1) was needed to perform the merge (output lives in non-tmp)

katiedelange: Confirming that Harper and I have reviewed the [latest version](https://batch.hail.populationgenomics.org.au/batches/626636/jobs/1) of the coverage table produced, after the most recent adjustments (using the existing reference data, fixing issues with reference blocks loading incorrectly at the boundary regions, and updating the `compute_coverage_stats` method), and we are happy that it looks reasonable on the test data.

katiedelange (approved):
LGTM, nice work Harper!

PR #1187, opened 2025-04-01
[Quick Fix] Multiple Active Analysis Inputs 

vivbak:
We've encountered a challenge where multiple active versions of an analyses output (like a cram of gvcf) should exist at once. 

At the moment, production pipelines pulls only the latest analyses output (probably). In any case, it will pull one output and users have no way to influence this behaviour. 

Further, a cohort is defined as a list of sequencing groups, so they don't exist as a suitable solution here. 

In this PR we aim to:
- Query specific gvcf’s based on a config param e.g. ‘dragen’ or ‘genotype’ — this relates to the stage that processed them which is stored in metamist
- Update the SequencingGroup object to also store this information alongside the existing sg.gvcf, e.g. sg.gvcf_stage_name

There was an additional dot point to:

- Update the analysis_meta field for the combiner to include this new provider information.

However, it seems that the only param I can give the method is the VDS output. I can create a new output file that stores the stage_name, but all I can do is register _that_ file with information within it.

The analysis_meta method will need further re-working to make the right params available, which will take more time. In the meantime we can explicitly use output_version, to help with readability so we are clear on the outputs.

**Note, these changes have not gone through an e2e test, and keen for analysis support to run one before merging**


cpg_workflows/metamist.py
```diff
@@ -99,11 +99,12 @@
 )
 
 
+# TODO, add a filter for meta, stage.
 GET_ANALYSES_QUERY = gql(
     """
         query AnalysesQuery($metamist_proj: String!, $analysis_type: String!, $analysis_status: AnalysisStatus!) {
             project(name: $metamist_proj) {
-                analyses (active: {eq: true}, type: {eq: $analysis_type}, status: {eq: $analysis_status}) {
+                analyses (active: {eq: true}, type: {eq: $analysis_type}, status: {eq: $analysis_status}, meta: {stage: {eq: $stage_name}}) {
```
dancoates, line 107: will need to add `$stage_name` as a variable to the query ie.

```
query AnalysesQuery($metamist_proj: String!, $analysis_type: String!, $analysis_status: AnalysisStatus!, $stage_name: String) ...

```

dancoates: Also, annoyingly this meta filter doesn't seem to be available directly on analyses in a project, you have to query like this, with sequencingGroups inbetween :/

```
query AnalysesQuery($stage_name: String) {
  project(name: "fewgenomes") {
    sequencingGroups {
      analyses(meta: {stage: {eq: $stage_name}}) {
        id
        type
        meta
        output
        status
        sequencingGroups {
          id
        }
      }
    }
  }
}
```

otherwise you get the error:

```
{
  "data": null,
  "errors": [
    {
      "message": "dict can not be used as parameter",
      "locations": [
        {
          "line": 3,
          "column": 7
        }
      ],
      "path": [
        "project",
        "analyses"
      ]
    }
  ]
}
```

cpg_workflows/inputs.py
```diff
@@ -277,11 +295,13 @@ def _populate_analysis(cohort: Cohort) -> None:
             dataset.get_sequencing_group_ids(),
             analysis_type=AnalysisType.GVCF,
             dataset=dataset.name,
+            meta=get_gvcf_meta(),
```
dancoates, line 298: nitpick, but these get_X_meta functions are called a bunch of times in this method, might be better to store the return value as a variable.

cpg_workflows/metamist.py
```diff
@@ -435,12 +436,17 @@ def get_analyses_by_sgid(
...
+        stage_name_key = f"{analysis_type.value.lower()}_stage_name"
+        stage_name = meta.get(stage_name_key) if meta else None
+
         analyses = query(
             GET_ANALYSES_QUERY,
             variables={
                 'metamist_proj': metamist_proj,
                 'analysis_type': analysis_type.value,
                 'analysis_status': analysis_status.name,
+                'stage_name': stage_name,
```
dancoates, line 449: I've tested this out, and when you pass `None` as a variable to the meta filter it has the same effect as if the stage name filter were not present - which I think is the intention, ie. if the stage name is not in the config then we shouldn't filter by stage name. But - this feels a bit flaky as generally I'd expect passing `None` as the value for an `eq` filter would return items where that field's value equals `None`, it almost feels like a bug in metamist. 

I think it would be more robust to not include the stage name filter in the query if the stage name value is `None`.  Would need to change `GET_ANALYSES_QUERY` to be a function that accepts a boolean like `include_stage_filter` or something to do this though.

dancoates (changes requested):

cpg_workflows/metamist.py
MattWellie: Could be better to query for everything and post-filter down to a specific type? 

A post-query filter method could be more reactive to config parameters, while a query is more hard coded. I'm picturing a query block in config like:

```yaml
[gvcf_meta_filters]
stage = "PipelineStage"
genome_reference = "gs://.../..."
```

so you can be as granular as you want without requiring increasingly complicated queries

cpg_workflows/inputs.py
vivbak: classic me behaviour, will fix

vivbak:
Note: If this is not immediately blocking, I'd be keen to consider a better solution for CPG Flow instead of merging this one. Will await clarification from @Alexander-Stuckey

cpg_workflows/metamist.py
dancoates: Yeah that sounds reasonable, query building in graphql is pretty messy

Alexander-Stuckey: I like the idea of being able to specify some query filters in the config

vivbak: Yup, big agree with the above. I'll leave the query alone and do the filtering in post.

cpg_workflows/metamist.py
vivbak: Good point! Fortunately no longer required given the refactor from other feedback above :)

vivbak:
That's ready for a re-review now

vivbak:
@Alexander-Stuckey or @michael-harper, when this gets approved would you be able to run a test in `test` with your data to check all is well before we merge.

Alexander-Stuckey:
> Note: If this is not immediately blocking, I'd be keen to consider a better solution for CPG Flow instead of merging this one. Will await clarification from @Alexander-Stuckey

I think that this is a blocker for @michael-harper , who might want to select some older data in bioheart for browser prep testing.
It's no longer a blocker for me, as I need to register all the reheadered CRAMs for tob, so those will become the default ones selected for realignment.

vivbak:
Gotcha, in that case @michael-harper if you'd like this PR in I'd be keen to get your help testing it!

dancoates (approved):
Looking good, just one tiny comment

cpg_workflows/metamist.py
```diff
@@ -760,3 +764,8 @@ def parse_reads(  # pylint: disable=too-many-return-statements
             )
 
         return fastq_pairs
+
+
+def filter_analyses_by_meta(analyses: list[dict], meta_filter: dict) -> list[dict]:
+    """Filter analyses by matching key-value pairs in the meta field."""
+    return [a for a in analyses if a.get("meta") and all(a["meta"].get(k) == v for k, v in meta_filter.items())]
```
jmarshall, line 771: How did the double quotes get past the linter-whiner…

cpg_workflows/inputs.py
```diff
@@ -292,6 +326,8 @@ def _populate_analysis(cohort: Cohort) -> None:
                     analysis.output,
                 )
                 sequencing_group.gvcf = GvcfPath(path=analysis.output)
+                if get_gvcf_meta():
```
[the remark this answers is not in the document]
jmarshall: Nice, I was going to suggest `:=` but that's even better.

cpg_workflows/metamist.py
dancoates: looks like we have it turned off https://github.com/populationgenomics/production-pipelines/blob/main/pyproject.toml#L3C1-L3C26

vivbak: whooops

jmarshall (approved):
Looks like a good approach to me too, just trivia comments.

vivbak:
Thanks @dancoates & @jmarshall! Have fixed all of those. Just awaiting an e2e test before we merge this one :)

michael-harper:
Will put together a test now @vivbak !

vivbak:
Latest test, thanks @michael-harper has failed https://batch.hail.populationgenomics.org.au/batches/593862

We'll continue to investigate why this seemingly isn't populating appropriately. 

**Do not merge the PR**

cpg_workflows/metamist.py
MattWellie: This is a pretty frustrating one.. AFAIK this is because we're skipping this rule with Black, and also skipping this rule ([Q000](https://github.com/populationgenomics/production-pipelines/blob/main/pyproject.toml#L66)) with Ruff. The `quote-style = "single"` rule is set in ruff.format, but we don't run ruff-format in CI, only ruff-lint.

Basically, I'm not sure this repo has had its grand reckoning PR when incorporating ruff (instead of/in addition to black), so there's still a lot of broken linting rules. Best to just look past some of these and solve when code is migrated out

vivbak:
This looks like it's working to me, https://batch.hail.populationgenomics.org.au/batches/593904/jobs/1 🥳 
But I'll wait for confirmation from @michael-harper, because I think the outputs didn't quite align with what he was expecting.

michael-harper:
Output looks good to me. Getting an error in a downstream stage that I think is unrelated so happy to go ahead and merge 🙂 Thanks @vivbak !

PR #1189, opened 2025-04-03
Register reheadered nagim crams

Alexander-Stuckey:
Register the reheadered NAGIM CRAMs in Metamist and add a comment to the sequencing group that the reheadered CRAM is now the latest analysis, up until the time that the CRAMs are realigned with Dragen 3.7.8.

The following example analysis entry is created

```
{'meta': {'archive': 'False',
          'project': 'tob-wgs',
          'sequence_type': 'genome',
          'sequencing_type': 'genome',
          'size': 22866202841,
          'source': 'Reheadered NAGIM from '
                    'gs://.../xxx.cram',
          'staging': 'False',
          'upload': 'False'},
 'output': 'gs://.../reheadered/xxx.cram',
 'sequencing_group_ids': ['xxx'],
 'status': 'completed',
 'type': 'cram'}
```

### DIFF
--- a/scripts/register_reheadered_nagim_crams.py
+++ b/scripts/register_reheadered_nagim_crams.py
@@ -122,14 +122,12 @@ def main(
     gcs_prefix: str,
     dry_run: bool,
 ):
-    # config = get_config(True)
+    config = get_config(True)
 
-    # dataset = config['workflow']['dataset']
-    # if config['workflow']['access_level'] == 'test' and not dataset.endswith('-test'):
-    #     dataset = f'{dataset}-test'
+    dataset = config['workflow']['dataset']
+    if config['workflow']['access_level'] == 'test' and not dataset.endswith('-test'):
+        dataset = f'{dataset}-test'
     coloredlogs.install(level=logging.INFO)
-
-    dataset = 'tob-wgs-test'
 
     project_analyses = query(FIND_ANALYSES, {'project': dataset})['project']['analyses']
     analysis_by_path = {analysis['output']: analysis for analysis in project_analyses if 'nagim' in analysis['output']}
--- a/scripts/register_reheadered_nagim_crams.py
+++ b/scripts/register_reheadered_nagim_crams.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+
+"""
+Register reheadered NAGIM CRAM files into Metamist
+
+Typical usage:
+
+analysis-runner ... scripts/register_reheadered_nagim_crams.py
+  --gcs-prefix gs://.../reheadered/
+  --dry-run (optional)
+"""
+
+import datetime
+import logging
+import subprocess
+from typing import Any
+
+import click
+import coloredlogs
+
+from cpg_utils import to_path
+from cpg_utils.config import get_config
+from metamist import models
+from metamist.apis import AnalysisApi
+from metamist.graphql import gql, query
+
+FIND_ANALYSES = gql(
+    """
+query AnalysesQuery($project: String!) {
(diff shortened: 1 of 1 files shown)


scripts/register_reheadered_nagim_crams.py
```diff
@@ -0,0 +1,167 @@
...
+
+FIND_ANALYSES = gql(
+    """
+query AnalysesQuery($project: String!) {
+  project(name: $project) {
+    analyses(type: {eq: "cram"}, active: {eq: true}) {
+      id
+      meta
+      output
+      sequencingGroups {
```
dancoates, line 35: should this be checking for active sequencing groups? It's possible to have an archived SG attached to a active analysis

scripts/register_reheadered_nagim_crams.py
```diff
@@ -0,0 +1,167 @@
...
+""",
+)
+
+FIND_SAMPLES = gql(
+    """
+query SamplesQuery($project: String!) {
+  project(name: $project) {
+    samples {
+      id
+      sequencingGroups {
```
dancoates, line 55: Same here, should this be checking for active sgs?

Alexander-Stuckey: Have filtered for active sequencing groups now

scripts/register_reheadered_nagim_crams.py
Alexander-Stuckey: This is handled in the python code now, since we can't do it in GQL

scripts/register_reheadered_nagim_crams.py
```diff
@@ -0,0 +1,171 @@
...
+
+    for fname in cram_list:
+        path = to_path(fname)
+        reheadered_path = path.parent / 'reheadered' / path.name
+        if str(reheadered_path) not in flist:
+            continue
+        try:
+            real_sequnging_group: str = sample_by_id[analysis_by_path[fname]['sequencingGroups'][0]['sample']['id']]
+        except KeyError:
+            logging.info(f'The sample {analysis_by_path[fname]["sequencingGroups"][0]["sample"]["id"]} is not active.')
```
dancoates, line 158: This looks like it will probably work, but the KeyError could also be because the filename is not found in the `analysis_by_path` list. Which if that is the case will throw again here as it is accessing the same path. If you're confident that all the files will exist in the `analysis_by_path` object then it is probably ok for a one-off script.

scripts/register_reheadered_nagim_crams.py
```diff
@@ -0,0 +1,171 @@
...
+
+    logging.info(f'Total reheadered NAGIM CRAM file entries found: {len(cram_list)}')
+
+    for fname in cram_list:
+        path = to_path(fname)
+        reheadered_path = path.parent / 'reheadered' / path.name
+        if str(reheadered_path) not in flist:
+            continue
+        try:
+            real_sequnging_group: str = sample_by_id[analysis_by_path[fname]['sequencingGroups'][0]['sample']['id']]
```
dancoates, line 156: just a typo here, probably worth fixing

dancoates (approved):
I think this looks good, for the metamist querying at least - aside from a couple of minor comments – but I'm not super knowledgeable on these sort of scripts so would be good to wait for another review or two before merging.

scripts/register_reheadered_nagim_crams.py
Alexander-Stuckey: I'm going to fix this anyway, otherwise it'll bug me 😆

PR #1190, opened 2025-04-04
only get samples with sequencing groups and analyses

Alexander-Stuckey:
Only get samples that actually have a sequencing group and an analysis

### DIFF
--- a/scripts/register_reheadered_nagim_crams.py
+++ b/scripts/register_reheadered_nagim_crams.py
@@ -136,7 +136,7 @@ def main(
     sample_by_id = {
         sample['id']: sample['sequencingGroups'][0]['id']
         for sample in project_samples
-        if len(sample['sequencingGroups']) != 0 and len(sample['sequencingGroups'][0]['analyses']) != 0
+        if sample['sequencingGroups'] and sample['sequencingGroups'][0]['analyses']
     }
 
     cram_list = analysis_by_path.keys()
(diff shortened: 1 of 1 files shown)


scripts/register_reheadered_nagim_crams.py
```diff
@@ -132,7 +132,12 @@ def main(
     project_analyses = query(FIND_ANALYSES, {'project': dataset})['project']['analyses']
     analysis_by_path = {analysis['output']: analysis for analysis in project_analyses if 'nagim' in analysis['output']}
     project_samples = query(FIND_SAMPLES, {'project': dataset})['project']['samples']
-    sample_by_id = {sample['id']: sample['sequencingGroups'][0]['id'] for sample in project_samples}
+
+    sample_by_id = {
+        sample['id']: sample['sequencingGroups'][0]['id']
+        for sample in project_samples
+        if len(sample['sequencingGroups']) != 0 and len(sample['sequencingGroups'][0]['analyses']) != 0
```
MattWellie, line 139:
```suggestion
        if sample['sequencingGroups'] and sample['sequencingGroups'][0]['analyses']
```

MattWellie (approved):
I'm pretty confident if the first condition fails it won't test the second one... in which case this is fine

scripts/register_reheadered_nagim_crams.py
Alexander-Stuckey: Yeah pretty sure python evaluates left to right and stops at the first failure, so it should be fine.

PR #1191, opened 2025-04-07
Update from a much earlier GATK-SV commit

MattWellie:
The diff I'm using to determine differences in stage inputs

https://github.com/populationgenomics/gatk-sv/compare/6d6100082297898222dfb69fcf941d373d78eede..dc145a52f76a6f425ac3f481171040e78c0cfeea

- Some changes to GatherSampleEvidence, more changes required if we want to accommodate DRAGEN inputs
- Many changes to MakeCohortVcf, including needing a bunch more references available (and a fresh rsync from Broad)
- Updates the name of gnomAD attributes to V4 (will require corresponding change to Talos. Does it need more changes for Seqr?)
- Updates the docker images in config
- Removes the need for a 'use_for_all_workflows' config file, instead adding the relevant contents to either the single- or multi-sample workflow's main config (applied by default)

## Still Needs

- ~needs the docker images to be copied over from Broad~
- ~needs the config file to be updated to point to new docker images~


cassimons:
> more changes required if we want to accommodate DRAGEN 
We don't need this now but will likely want it eventually. It is up to you if it is easier to ignore for the moment.

MattWellie:
> > more changes required if we want to accommodate DRAGEN
> We don't need this now but will likely want it eventually. It is up to you if it is easier to ignore for the moment.

It might be easier to add this during this PR, unless there's a possibility of PA forking this off into a repository they own entirely. The changes are a boolean for whether to use the dragen features (in a couple of places), and alignment references so that the workflow can do local realignment with BWA MEM

EddieLF (approved):
LGTM

PR #1193, opened 2025-04-10
Truncate rsid and keep only first variant for colocated ones

EddieLF:
Before exporting to elasticsearch, we need to make sure the rsid field is a reasonable length. The rsid field is populated before splitting and normalisation, so it can be a huge string of colocated variants.

This change will split the rsid on the ';' character to ensure it only contains the first variant, if there are multiple. Then, it will truncate the rsid to 512 characters at most. 

This change ran successfully in test for 6 individuals (2x trio VCFs)
https://batch.hail.populationgenomics.org.au/batches/594145

However, I'm not sure if this will ensure uniqueness, which is required for elasticsearch / seqr integration.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.23
+current_version = 1.36.24
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


MattWellie (approved):
Looks good as a stopgap, keen to see if this creates any issues. I'd be happy rolling this earlier into AnnotateCohort, but this is a single dataset pipeline at the moment so same same

PR #1194, opened 2025-04-15
Use trusted publisher for packaging to PyPI

nevoodoo:
The current workflow uses orthodox username/password to authenticate with PyPI for packaging. This change switches to using OIDC via Trusted Publishers for PyPI.


vivbak (approved):

jmarshall:
[Best practice](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) is to separate the publishing job from the sdist-building job so that only the publishing job gets the credentials, but 🤷

jmarshall:
[Let's not merge this until the driver image has been rebuilt first!]

PR #1195, opened 2025-04-16
removed both calls to ht.show()

joshuamschmidt:
ht.show() appears to be susceptible to transient type errors, for who knows why reasons.....
see this [discussion](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1744713605005159).

As a stop gap to resolving the ultimate cause of this error, this PR removes the two calls to ht.show() in `ancestry.py`

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -68,7 +68,6 @@ def add_background(
                 logging.info(f'Adding metadata table to background dataset: {path}')
                 sample_qc_background = hl.read_table(path)
                 metadata_tables.append(sample_qc_background)
-                logging.info(f'metadata table appended to metadata_tables: {sample_qc_background.show()}')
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
Looks good to me,

If it's considered a priority I'd recommend booking in some time to review this script in general. There's a few idiosyncrasies like many calls to `can_reuse` being combined with `overwrite=True`... [this will never do anything](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/utils.py#L250-L263). Outside the scope of this change, but there's definitely some oddities in here

joshuamschmidt:
> If it's considered a priority I'd recommend booking in some time to review this script in general. There's a few idiosyncrasies like many calls to `can_reuse` being combined with `overwrite=True`... [this will never do anything](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/utils.py#L250-L263). Outside the scope of this change, but there's definitely some oddities in here

Thanks for flagging. Haven't yet had time to review the code - just relying on what was previously developed to push through browser launch.

PR #1198, opened 2025-04-21
Frequencies and browser prep refactor

katiedelange:
Key changes:
- Refactor frequencies to group related functions together (filtering, variant QC, frequency estimation, histogram creation)
- Adding checkpoints to force periodic evaluation
- Including functionality to compute allele frequencies by subset
- Remove the need to calculate AC_raw for stratified allele frequencies, as we don't need this annotation downstream and it makes computation more complex
- Defining a default empty table to return from the browser prep function when no frequency data is passed for a given datatype (exome or genome).
- Adding variant filter flags (VQSR, InbreedingCoefficient, AC0)
- Simplifying the computation of co-located variants

### DIFF
--- a/cpg_workflows/large_cohort/browser_prepare.py
+++ b/cpg_workflows/large_cohort/browser_prepare.py
@@ -133,7 +133,7 @@ array<
 
 
 def _nullify_nan(value):
-    return hl.if_else(hl.is_nan(value), hl.null(value.dtype), value)
+    return hl.if_else(hl.is_nan(value), hl.missing(value.dtype), value)
 
 
 def _normalized_contig(contig: hl.expr.StringExpression) -> hl.expr.StringExpression:
--- a/cpg_workflows/large_cohort/browser_prepare.py
+++ b/cpg_workflows/large_cohort/browser_prepare.py
@@ -1,23 +1,146 @@
 import itertools
+import logging
 
 import hail as hl
 
+from gnomad.utils.filtering import add_filters_expr
 
-def nullify_nan(value):
+EMPTY_TABLE_CONFIG = """
+array<
+    struct{
+        locus: locus<GRCh38>,
+        alleles: array<str>,
+        variant_id: str,
+        rsids: str,
+        type: struct {
+            colocated_variants: struct {
+                all: array<str>
+            },
+            subsets: set<str>,
+            flags: set<str>,
+            freq: struct {
+                all: struct {
+                    ac: int32,
+                    an: int32,
+                    hemizygote_count: int32,
+                    homozygote_count: int32,
+                    ancestry_groups: array<struct {
+                        id: str,
+                        ac: int32,
+                        an: int32,
+                        hemizygote_count: int32,
+                        homozygote_count: int32
+                    }>
+                }
+            },
+            fafmax: struct {
+                faf95_max: float64,
+                faf95_max_gen_anc: str,
+                faf99_max: float64,
+                faf99_max_gen_anc: str
+            },
+            age_distribution: struct {
+                het: struct {
+                    bin_edges: array<float64>,
+                    bin_freq: array<int64>,
+                    n_smaller: int64,
+                    n_larger: int64
+                },
+                hom: struct {
+                    bin_edges: array<float64>,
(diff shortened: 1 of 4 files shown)


cpg_workflows/large_cohort/frequencies.py
```diff
@@ -67,78 +57,84 @@ def frequency_annotations(
...
+    logging.info('Annotating variant QC metrics...')
+    logging.info('Base variant QC...')
     mt = hl.variant_qc(mt)
 
-    logging.info('Calculating InbreedingCoeff...')
-    # NOTE: This is not the ideal location to calculate this, but added here
-    # to avoid another densify.
-    # The algorithm assumes all samples are unrelated:
+    logging.info('Allele-specific statistics...')
+    mt = mt.annotate_rows(info=site_only_ht[mt.row_key].info)
```
michael-harper, line 100: Just noting here that `mt` is split but the `site_only_ht` is not. Which I believe means the `info` row field in `mt` is filled with NA's at sites that have been split.

michael-harper (commented):
Ran changes through test dataset and looks good on my end. Noting that the test data I'm using is only ten samples and does not have the diversity we would like but structurally everything looks good!

cpg_workflows/large_cohort/browser_prepare.py
```diff
@@ -36,44 +159,51 @@ def variant_id(locus: hl.expr.LocusExpression, alleles: hl.expr.ArrayExpression,
...
+def _freq_index_key(subset=None, pop=None, sex=None, raw=False):
+    parts = [s for s in [subset, pop, sex] if s is not None]
+    parts.append("raw" if raw else "adj")
+    return "_".join(parts)
 
-    ds = ds.select_globals()
 
-    subsets = set(m.get("subset", None) for m in g.freq_meta)
+def _freq(ds, *args, **kwargs):
+    return ds.freq[g.freq_index_dict[_freq_index_key(*args, **kwargs)]]
```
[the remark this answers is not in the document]
katiedelange: Great catch, I missed that when transcribing across!

cpg_workflows/large_cohort/browser_prepare.py
```diff
@@ -36,44 +159,51 @@ def variant_id(locus: hl.expr.LocusExpression, alleles: hl.expr.ArrayExpression,
...
 
-    ######################
-    # Colocated variants #
-    ######################
+def prepare_gnomad_v4_variants_helper(ds_path: str, exome_or_genome: str) -> hl.Table:
+
+    if ds_path is None:
+        ds = hl.Table.parallelize(hl.literal([], "".join(EMPTY_TABLE_CONFIG.split()))).key_by('locus', 'alleles')
+        ds = ds.rename({'type': exome_or_genome})
+        return ds
```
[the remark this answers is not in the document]
katiedelange: Yep, exactly - so we can set up the first genome-only cohort without having to wait for the exome data to be ready.  If you call `prepare_v4_variants(exome_ds_path=None, genome_ds_path="path/to/genome/frequencies") it generates an empty data struct for the exome component and the browser release table is still functional.

cpg_workflows/large_cohort/frequencies.py
katiedelange: Ooh, ok, that's a very important detail I'd missed - while ultimately I want to transition everything to the VQSR output and avoid using `site_only_ht` entirely, for now we should apply the split to `site_only_ht` before we annotate. Are you able to look into this? Otherwise I might get a chance tomorrow afternoon.

katiedelange: The investigation suggests this is not straightforward - split_multis does not correctly split the allele specific information. As site_only_vcf is only a placeholder to allow testing until we get a very similar issue resolved for the VQSR stage, I suggest we leave this incorrect for now and proceed with pressure testing the browser loading. We will then replace this with VQSR before generating the release dataset.

michael-harper (approved):
Happy to continue and merge this PR to pressure test the browser loading knowing that we will need to fix the splitting of AS fields down the line.

PR #1201, opened 2025-04-23
Remove additional pathing name inside MakeSiteOnly

michael-harper:
Already passing the full path to desired hail table output. Removing unnecessary addition of name to path.


Alexander-Stuckey (approved):
Looks good to me

PR #1202, opened 2025-04-28
Adding default config for new prepare_browser_table workflow

michael-harper:
Every workflow listed in `main.py` requires it's own default config file. When adding `prepare_browser_table` as a separate workflow, cpg_workflows requires the existence of `/production-pipelines/configs/defaults/prepare_browser_table.toml`. This PR adds this file.


Alexander-Stuckey (approved):
Looks good to me

PR #1205, opened 2025-05-06
LoadVqsr Fix

michael-harper:
## Summary

The `LoadVqsr` stage processes a VQSR-annotated, sites-only VCF. However, this VCF contains fields that are incompatible with Hail’s VCF parser, requiring certain fields to be dropped during import. This workaround introduces complexity in the downstream Frequencies stage, which relies on annotations lost during parsing.

## Field Compatibility Issues

Some INFO fields in the VCF are incorrectly described in the header. For example, `SB` is declared as a Float, but the values are actually comma-separated lists of integers (e.g., `SB=6,11,2,0`). Hail fails to parse these entries, raising:
```
java.lang.NumberFormatException: For input string: "6,11,2,0"
```
To prevent parsing errors, we explicitly drop the `SB` field before writing the Hail Table, ensuring the parser never attempts to interpret it:
```
if 'SB' in ht.info:
    ht = ht.annotate(info=ht.info.drop('SB'))
```
We also drop all AS_* annotations and InbreedingCoeff, as these fields can cause errors during multiallelic splitting (`hl.split_multi_hts`). For example, if `AS_FilterStatus` has fewer values than the number of alternate alleles, Hail raises an array index out of bounds error:
```
HailException: array index out of bounds: index=1, length=1
```
To avoid this, we drop problematic fields before splitting:
```
ht = ht.annotate(info=ht.info.drop(*[f for f in ht.info if f.startswith('AS_')]))
```

## Downstream Impact

Since these fields are essential for downstream processing (e.g. frequency annotations), we must re-add them in later stages using the original site-only VCF, which retains the correct formatting.

However, the current `LoadVqsr` logic uses `hl.split_multi_hts`, which does not correctly split all INFO fields. For instance, while the locus and alleles fields are split as expected, `AC` and `AC_raw` remain as arrays of length two even after splitting a biallelic variant — indicating improper allele-specific splitting.

## Additional Note on AS_SB_TABLE

The `MakeSiteOnlyVcf` step includes a call to `adjust_vcf_incompatible_types()`, which stringifies `AS_SB_TABLE`, further breaking compatibility with Hail’s parser. A potential solution is to checkpoint the correctly formatted site-only table before this function runs, so that `LoadVqsr` can use the preserved version for import.

## Fixes

- To prevent parsing errors due to missing values in AS_* fields, we now pass `array_elements_required=False` to the parser, allowing INFO fields to contain missing values.
- Extract the header from VQSR vcf file and update the `SB` field to be non-float of any length and supply this header-only file to `import_vcf` 
- To avoid issues caused by the stringified `AS_SB_TABLE` field, we checkpoint the site-only Hail Table within `MakeSiteOnlyVcf` before this field is modified. During `LoadVqsr`, we restore the original (unstringified) version from this checkpoint prior to splitting.
- Finally, we apply multiallelic splitting using utility functions from the `gnomad_methods` repository to ensure that all fields — including AS_* annotations — are split correctly and consistently.


michael-harper:
@MattWellie this is breaking here: https://github.com/populationgenomics/production-pipelines/blob/83ce8d40376a56c1a15dfd4c4083c866bce6abc3/cpg_workflows/query_modules/seqr_loader.py#L101

I'm getting the AS_SB_TABLE field from the `MakeSiteOnlyVcf` stage before we run `adjust_vcf_incompatible_types` that stringifies this field. This was done to avoid dealing with strings and possible missed edge cases un-stringifying them

cpg_workflows/stages/large_cohort.py
```diff
@@ -242,12 +242,32 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+            inferred_pop=cohort.analysis_dataset.prefix()
+            / get_workflow().name
+            / ancestry_version
+            / 'ancestry'
+            / 'inferred_pop.ht',
+            sample_qc_ht=cohort.analysis_dataset.prefix()
+            / get_workflow().name
+            / ancestry_version
+            / 'ancestry'
+            / 'sample_qc_ht.ht',
```
katiedelange, line 270: These Ancestry stage changes are solving a separate problem and probably shouldn't be part of this PR?

michael-harper: Totally agree - this was just to help with testing. Have reverted back

cpg_workflows/stages/large_cohort.py
Alexander-Stuckey, line 1: Happy with the proposed change to use the fixed header in loadVQSR

katiedelange (commented):
I'm happy that this resolves our issues with load_vqsr, great work Harper. However, we need RD eyes on the breaking seqr_loader change before this can be approved.

MattWellie:
Happy to approve this in general, but wanted it to mostly come from PA. There's been a bit of a hybrid discussion on slack so I'll update here - seqr_loader is something of a legacy pipeline at this point, we now use `rd_combiner`. The equivalent step of that workflow does exactly as you suggest, the private `load_vqsr` here [here](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/scripts/annotate_cohort_small_vars.py#L58).

I'm comfortable that if we need to reinstate seqr_loader (and indeed if we port it to cpg-flow at all) we'll be able to pull in a working copy of this method from the rd_combiner script, or git history.

michael-harper:
Thanks @MattWellie and everyone for your reviews. It sounds like we are okay to move forward with the proposed changes - @Alexander-Stuckey / @katiedelange  I've re-requested a review for approval

Alexander-Stuckey (approved):
Looks good to me

katiedelange (approved):

PR #1208, opened 2025-05-12
Stripy config

EddieLF:
Adds a new argument to the stripy stage and job, `stripy_config`. This is a dict that can be modified when invoking the workflow via analysis-runner, and is used to update the `config.json` file that is pulled with the stripy image (see https://gitlab.com/andreassh/stripy-pipeline/-/blob/main/config.json?ref_type=heads)

Defaults have been added to the `defaults.toml` that replace the use of `sed` to modify the json file, which was done in the job as a quick and dirty method of updating. 

This change will allow us to completely customise the stripy config.json through the config toml used in the analysis-runner submission. 

e.g. If we want to limit the number of nucleotides extracted before and after a locus, we can set the config as
```toml
[stripy.config]
fregion_length_bp = 60
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.25
+current_version = 1.36.26
 commit = True
 tag = False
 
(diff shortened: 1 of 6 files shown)


EddieLF:
This PR was blocked by work in the images repo because it needed the stripy image to have jq installed. Resolved by https://github.com/populationgenomics/images/pull/223

MattWellie (approved):

PR #1210, opened 2025-05-13
Rename InbreedingCoeff metric

katiedelange:
Update the inbreeding coefficient metric name from "InbreedingCoeff" to "inbreeding_coeff" for compatibility with the browser.


cpg_workflows/large_cohort/browser_prepare.py
```diff
@@ -361,7 +361,7 @@ def prepare_gnomad_v4_variants_helper(ds_path: str | None, exome_or_genome: str)
 
     inbreeding_coeff_cutoff = -0.3
     filters = {
-        "InbreedingCoeff": ds.InbreedingCoeff[0] < inbreeding_coeff_cutoff,
```
michael-harper, line 364: I was going to ask if we need to change the key here as well? But taking a quick look at the browser code [here](https://github.com/broadinstitute/gnomad-browser/blob/5696124ff1a13a7cc02b9e34f90bffe4bbcac471/browser/src/QCFilter.tsx#L6-L13) it looks like it is expecting "InbreedingCoeff" in the filter field

michael-harper (approved):
This looks like it will do the job. One comment provided that suggests this PR is good to go!

PR #1212, opened 2025-05-15
Prepend global variants with sequencing type

katiedelange:
Prepend all global variables in the variants table with 'exome_' or 'genome_' as appropriate.


michael-harper (approved):
Investigated with notebook. Looks good!

PR #1214, opened 2025-05-15
Cohort level crammultiqc

Alexander-Stuckey:
Basically a duplicate of the `CramMultiQC` stage, but operates on a cohort instead of a dataset. This lets someone create a single MultiQC report for a cohort than contains sequencing groups from multiple datasets.

### DIFF
--- a/cpg_workflows/stages/cram_qc.py
+++ b/cpg_workflows/stages/cram_qc.py
@@ -16,9 +16,10 @@ from cpg_workflows.jobs.picard import picard_collect_metrics, picard_hs_metrics,
 from cpg_workflows.jobs.samtools import samtools_stats
 from cpg_workflows.jobs.verifybamid import verifybamid
 from cpg_workflows.stages.align import Align
-from cpg_workflows.targets import Dataset, SequencingGroup
+from cpg_workflows.targets import Cohort, Dataset, SequencingGroup
 from cpg_workflows.utils import exists
 from cpg_workflows.workflow import (
+    CohortStage,
     DatasetStage,
     SequencingGroupStage,
     StageInput,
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -97,6 +97,7 @@
         AncestryPlots,
         GvcfMultiQC,
         CramMultiQC,
+        CohortCramMultiQC,
```
MattWellie, line 100: do you want both?

cpg_workflows/stages/cram_qc.py
```diff
@@ -305,3 +306,100 @@ def queue_jobs(self, dataset: Dataset, inputs: StageInput) -> StageOutput | None
...
+        """
+        if config_retrieve(['workflow', 'skip_qc'], False):
+            return {}
+
+        # get the unique hash for these Sequencing Groups
+        sg_hash = cohort.get_alignment_inputs_hash()
+        return {
+            'html': cohort.analysis_dataset.web_prefix() / 'qc' / 'cram' / sg_hash / 'multiqc.html',
+            'json': cohort.analysis_dataset.prefix() / 'qc' / 'cram' / sg_hash / 'multiqc_data.json',
+            'checks': cohort.analysis_dataset.prefix() / 'qc' / 'cram' / sg_hash / '.checks',
```
MattWellie, line 329: This is currently the exact same output path as the non-Cohort version. I'd suggest changing one or the other, else this stage will have to be force-run, and the outputs won't be easy to disentangle

main.py
Alexander-Stuckey: Yeah probably. The cohort level one is a special case.

cpg_workflows/stages/cram_qc.py
Alexander-Stuckey: Whoops, fixed

MattWellie (approved):

PR #1218, opened 2025-05-21
Use allele-specific vqsr filters

katiedelange:
Refine the variant filtering to:
- Bring through the AS_lowqual filter into the final VQSR and frequencies hail tables
- Filter variants that fail on this AS_lowqual flag, or that fail on the allele-specific (as opposed to site-specific) VQSR filters

With VQSR now working correctly to bring in the allele-specific statistics, we can now also drop the dependency on the sites-only hail table.


michael-harper (approved):
Looks good!

PR #1220, opened 2025-05-21
Oneshot vds patch

MattWellie:
A fix for https://centrepopgen.slack.com/archives/C018KFBCR1C/p1747790657229869

- targets a VDS to modify, reads the reference_data MT
- reads a reference MT to template from
- adjusts the target VDS's reference_data MT to match the entry schema of the template
- write the resulting MT to a temporary location
- overwrite the original MT


EddieLF (approved):
Nice

PR #1225, opened 2025-05-21
Withdrawals in combiner

Alexander-Stuckey:
Implement a method to withdrawal samples in the combiner stage.

This is not the best implementation, as it creates the full VDS first, and then removes samples from it. However, this lets us keep the ability to merge multiple VDS together, which would otherwise become very complicated.

Also refactor the code into a more cpg_flow style, separating the logic out of the stage and into the job files.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.31
+current_version = 1.36.32
 commit = True
 tag = False
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,8 @@
 [bumpversion]
-current_version = 1.36.29
+current_version = 1.36.30
 commit = True
 tag = False
+allow_dirty = True
 
 [bumpversion:file:setup.py]
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -2,7 +2,6 @@
 current_version = 1.36.30
 commit = True
 tag = False
-allow_dirty = True
 
 [bumpversion:file:setup.py]
 
--- a/cpg_workflows/large_cohort/combiner.py
+++ b/cpg_workflows/large_cohort/combiner.py
@@ -76,12 +76,12 @@ def combiner(cohort: Cohort, output_vds_path: str, save_path: str) -> PythonJob:
     gvcf_external_header: str | None = None
     sequencing_group_names: list[str] | None = None
 
-    if combiner_config.get('vds_analysis_ids', None) is not None:
-        for vds_id in combiner_config['vds_analysis_ids']:
-            tmp_query_res, tmp_sg_ids_in_vds = get_vds_ids_output(vds_id)
-            vds_paths.append(tmp_query_res)
-            sg_ids_in_vds = sg_ids_in_vds + tmp_sg_ids_in_vds
-            sgs_for_withdrawal = [sg for sg in sg_ids_in_vds if sg not in cohort_sg_ids]
+    for vds_id in config_retrieve(['combiner', 'vds_analysis_ids'], []):
+        tmp_query_res, tmp_sg_ids_in_vds = get_vds_ids_output(vds_id)
+        vds_paths.append(tmp_query_res)
+        sg_ids_in_vds = sg_ids_in_vds + tmp_sg_ids_in_vds
+    
+    sgs_for_withdrawal = [sg for sg in sg_ids_in_vds if sg not in cohort_sg_ids]
 
     if combiner_config.get('merge_only_vds', False) is not True:
         # Get SG IDs from the cohort object itself, rather than call Metamist.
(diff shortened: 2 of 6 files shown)


cpg_workflows/large_cohort/combiner.py
```diff
@@ -9,23 +9,123 @@
...
+    cohort_sg_ids: list[str] = cohort.get_sequencing_group_ids(only_active=True)
+    gvcf_external_header: str | None = None
+    sequencing_group_names: list[str] | None = None
+
+    if combiner_config.get('vds_analysis_ids', None) is not None:
+        for vds_id in combiner_config['vds_analysis_ids']:
+            tmp_query_res, tmp_sg_ids_in_vds = get_vds_ids_output(vds_id)
+            vds_paths.append(tmp_query_res)
+            sg_ids_in_vds = sg_ids_in_vds + tmp_sg_ids_in_vds
+            sgs_for_withdrawal = [sg for sg in sg_ids_in_vds if sg not in cohort_sg_ids]
```
MattWellie, line 84:
```suggestion
    for vds_id in config_retrieve(['combiner', 'vds_analysis_ids'], []):
        tmp_query_res, tmp_sg_ids_in_vds = get_vds_ids_output(vds_id)
        vds_paths.append(tmp_query_res)
        sg_ids_in_vds = sg_ids_in_vds + tmp_sg_ids_in_vds
    
    sgs_for_withdrawal = [sg for sg in sg_ids_in_vds if sg not in cohort_sg_ids]
```

I'm continuing on my mission to evangelise for `config_retrieve` everywhere. I find the full hierarchy of keys being present in each call easier than remembering which subset of the overall config `combiner_config` represents.

In this situation `sgs_for_withdrawal` is overwritten with each loop anyway, so moving it outside the loop makes sense. If you go this way you can also remove the empty `sgs_for_withdrawal` initialisation above

All personal flavour, feel free to ignore this

Alexander-Stuckey: Not overwriting each time is good, also happy to agree with `config_retrieve`.

cpg_workflows/large_cohort/combiner.py
```diff
@@ -9,23 +9,123 @@
...
+
+    if combiner_config.get('merge_only_vds', False) is not True:
```
MattWellie, line 86:
```suggestion
    if not config_retrieve(['combiner', 'merge_only_vds'], False):
```

cpg_workflows/large_cohort/combiner.py
```diff
@@ -9,23 +9,123 @@
...
+        # Get VDS IDs first and filter out from this list
+        new_sg_gvcfs = [str(sg.gvcf) for sg in cohort_sgs if sg.gvcf is not None and sg.id not in sg_ids_in_vds]
+        if new_sg_gvcfs:
+            gvcf_external_header = new_sg_gvcfs[0]
+            sequencing_group_names = [
+                str(sg.id) for sg in cohort_sgs if sg.gvcf is not None and sg.id not in sg_ids_in_vds
+            ]
+
+    # if new_sg_gvcfs and len(new_sg_gvcfs) == 0 and len(vds_paths) <= 1:
+    #     raise Exception
```
MattWellie, line 97:
```suggestion
    if not (new_sg_gvcfs or vds_paths):
        raise Exception
```

This still feels like a valid exception, otherwise you continue to the combiner but it won't be doing any combining?

.bumpversion.cfg
```diff
@@ -1,7 +1,8 @@
 [bumpversion]
-current_version = 1.36.29
+current_version = 1.36.30
 commit = True
 tag = False
+allow_dirty = True
```
MattWellie, line 5: 😉

Alexander-Stuckey: haha I didn't realise that got committed. It's because there are a couple of tracked files in the repo that I don't want to commit, but wanted to bump the version. I'll remove this.

cpg_workflows/large_cohort/combiner.py
Alexander-Stuckey: Yeah so this will actually throw an exception if you want to only remove samples. Let me update this to make sure it checks that there are no SGs to remove as well.

cpg_workflows/large_cohort/combiner.py
```diff
@@ -54,19 +154,9 @@ def run(
     # set up a quick logger inside the job
     logging.basicConfig(level=logging.INFO)
 
-    if not can_reuse(to_path(output_vds_path)):
+    if not can_reuse(to_path(output_vds_path)) or sgs_for_withdrawal:
```
MattWellie, line 157: I think at the moment this block re-runs the combiner, even if the output_vds_path exists?

Alexander-Stuckey: Does it? My brain is really slow at the moment, so would like some help

MattWellie: I think so, there's an `if` further down that I think can be used to solve this

Alexander-Stuckey: Actually I should remove `sgs_for_withdrawal`, as if you are getting back the VDS that you input then something is incorrect with the config

MattWellie (approved):
LGTMatt. Props for bothering to do the code restructure pre-migration, I'm kicking the can on that one until CPG-Flow

PR #1226, opened 2025-05-22
Somalierpedigree output location

### LINKED ISSUE #974: SomalierPedigree stage outputs overwrite old outputs
## Description

In the `SomalierPedigree` stage `expected_outputs`, [the html output](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/cram_qc.py#L163) does not make use of any sequencing groups / cohort hash, or sequencing type string.

## Impact

This means that when this stage is run for a given dataset, the html report will be saved to the exact same location every time, regardless of the input sequencing groups, cohorts, or sequencing type. **Old html reports will therefore always be replaced by newer ones for the same dataset.**

## Proposed fix

- Update the html output path so that the report is saved to a location containing an input hash. `dataset.alignment_inputs_hash()` can be used, like the rest of the expected outputs from this stage.
- The output paths for all expected outputs from this stage should also be updated to distinguish between reports created for cohorts or sequencing groups of different sequencing types. i.e. incorporate 'genome' or 'exome' into the path prefix, like is done for other stages.

EddieLF:
Resolves https://github.com/populationgenomics/production-pipelines/issues/974

Instead of saving to:
```
gs://cpg-dataset-main-web/cram-somalier-pedigree.html
gs://cpg-dataset-main-web/exome/cram-somalier-pedigree.html
```

It will now save to:
```
gs://cpg-dataset-main-web/somalier/cram/<sg_hash>/cram-somalier-pedigree.html
gs://cpg-dataset-main-web/exome/somalier/cram/<sg_hash>/cram-somalier-pedigree.html
```

Which is in line with how we save other DatasetStage / CohortStage results. This will prevent the report html being overwritten with every run.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.30
+current_version = 1.36.31
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.36.30
+  VERSION: 1.36.31
 
 jobs:
   docker:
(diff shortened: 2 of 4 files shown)


MattWellie (approved):
yisssss, I think I've made this change before as part of my crusade against the QC stages, but never followed it through

PR #1227, opened 2025-05-22
Think of a good title

MattWellie:
I don't know if this is the right thing

I think the current code in `withdrawals-in-combiner` might fall over in a couple of scenarios:
- `output_vds_path` exists, but we want to withdraw samples from it (VDSs can't be overwritten)
- if there's no combining to do, but there are samples to remove, the combiner will run again anyway, then we remove SGs from the result

I don't know if this is a full fix for any of this


Alexander-Stuckey:
>     * `output_vds_path` exists, but we want to withdraw samples from it (VDSs can't be overwritten)
I'm not sure that this can happen. Removal of samples is triggered by checking the contents of the VDS against the samples in the cohort, which means that if there are samples to withdraw, the cohort is different and has a different name. Cohort name forms part of the VDS output name, so this _should_ always be fine.


>     * if there's no combining to do, but there are samples to remove, the combiner will run again anyway, then we remove SGs from the result
Yeah this should be fixed.

MattWellie:
Agreed it probably can't mechanically happen. The original point this branched off had the combiner running if `not output_vds exists, or sgs_to_remove` but that's been patched now.

MattWellie:
> if there are samples to withdraw, the cohort is different and has a different name. Cohort name forms part of the VDS output name, so this should always be fine.

This is what tripped us up - the Cohorts should be immutable, but currently if samples are withdrawn they're deleted from the Cohorts that contain them. The same Cohort ID/Name doesn't always represent the same group of SGs, even though that's exactly their purpose 😮‍💨

cpg_workflows/large_cohort/combiner.py
```diff
@@ -146,39 +156,48 @@ def _run(
         vds_paths (list[str] | None): list of paths to VDSs
         specific_intervals (list[str] | None): list of intervals to use for the combiner, if using non-standard
     """
+
+    if vds_paths is None:
+        vds_paths = []
+
+    if sgs_for_withdrawal and tmp_prefix_for_withdrawals is None:
```
Alexander-Stuckey, line 163: This should never happen in practice, since `tmp_prefix_for_withdrawals` is always set on line 67, but worth having the check anyway in case something strange happens.

cpg_workflows/large_cohort/combiner.py
```diff
@@ -146,39 +156,48 @@ def _run(
...
     # set up a quick logger inside the job
     logging.basicConfig(level=logging.INFO)
 
-    if not can_reuse(to_path(output_vds_path)):
-        init_batch(worker_memory='highmem', driver_memory='highmem', driver_cores=4)
+    # init batch early, this method won't run if output_vds_path exists, so we have some work to do
+    init_batch(worker_memory='highmem', driver_memory='highmem', driver_cores=4)
 
+    # do we need to do any combining?
+    combining_to_do: bool = bool(gvcf_paths or len(vds_paths) > 1)
```
Alexander-Stuckey, line 177: I like it

Alexander-Stuckey (approved):
looks good to me

cpg_workflows/large_cohort/combiner.py
MattWellie: Ah, I think I just put this in as housekeeping. The arguments didn't have default values, so one of the tests was failing. If they default to None the test shuts up, but they look more optional than they really are

PR #1228, opened 2025-05-22
Lrs checkpoint mt immediately

EddieLF:
Adds a checkpointing step to the LRS annotation pipeline, immediately after the merged VCF has been imported. 

This is to address OOM issues when the number of partitions of the MatrixTable is evaluated on the following line. See [this Slack thread](https://centrepopgen.slack.com/archives/C082NA58R7S/p1747872704048229) for a discussion.


MattWellie (approved):
Here's hopin'

PR #1229, opened 2025-05-22
Lrs checkpoint mt immediately2

EddieLF:
Follow up from https://github.com/populationgenomics/production-pipelines/pull/1228

Use `mt.checkpoint` rather than `cpg_workflows.utils.checkpoint_hail` method - which calls `mt.n_partitions()` before checkpointing (exactly what we wanted to avoid)

### DIFF
--- a/cpg_workflows/query_modules/seqr_loader.py
+++ b/cpg_workflows/query_modules/seqr_loader.py
@@ -48,7 +48,7 @@ def annotate_cohort(
         force_bgz=True,
         array_elements_required=False,
     )
-    mt.checkpoint(output=str(checkpoint_prefix) + 'mt-imported.mt', overwrite=True)
+    mt.checkpoint(output=str(checkpoint_prefix / 'mt-imported.mt'), overwrite=True)
     logging.info(f'Imported VCF {vcf_path} as {mt.n_partitions()} partitions')
 
     # Annotate VEP. Do it before splitting multi, because we run VEP on unsplit VCF,
--- a/cpg_workflows/query_modules/seqr_loader.py
+++ b/cpg_workflows/query_modules/seqr_loader.py
@@ -48,7 +48,7 @@ def annotate_cohort(
         force_bgz=True,
         array_elements_required=False,
     )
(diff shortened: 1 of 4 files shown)


cpg_workflows/query_modules/seqr_loader.py
```diff
@@ -48,7 +48,7 @@ def annotate_cohort(
         force_bgz=True,
         array_elements_required=False,
     )
-    mt = checkpoint_hail(mt, 'mt-imported-vcf.mt', checkpoint_prefix)
+    mt.checkpoint(output=str(checkpoint_prefix) + 'mt-imported.mt', overwrite=True)
```
MattWellie, line 51:
```suggestion
    mt.checkpoint(output=str(checkpoint_prefix / 'mt-imported.mt'), overwrite=True)
```

this will work but the current checkpoint path will be a bit dodgy looking

MattWellie (approved):

cpg_workflows/query_modules/seqr_loader.py
EddieLF: Unfortunately linter doesn't like this 
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

cpg_workflows/scripts/mt_to_es_without_dataproc.py:259: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
cpg_workflows/query_modules/seqr_loader.py:51: error: Unsupported left operand
type for / ("str")  [operator]
        mt.checkpoint(output=str(checkpoint_prefix / 'mt-imported.mt'), ov...
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cpg_workflows/query_modules/seqr_loader.py:51: error: Unsupported left operand
type for / ("None")  [operator]
        mt.checkpoint(output=str(checkpoint_prefix / 'mt-imported.mt'), ov...
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cpg_workflows/query_modules/seqr_loader.py:51: note: Left operand is of type "str | None"
```

PR #1230, opened 2025-05-22
Highmem workers annotatecohort

EddieLF:
After https://github.com/populationgenomics/production-pipelines/pull/1228 and https://github.com/populationgenomics/production-pipelines/pull/1229 failed to resolve the OOM issues in the annotate cohort QoB execution for the long reads pipeline, we should look into to increasing the memory used by the workers. 

According to the QoB playbook [here](https://github.com/populationgenomics/team-docs/blob/main/playbooks/hail_query_on_batch.md#out-of-memory), you can pass the arg `worker_memory='highmem'` to `init_batch()` and then any QoB workers used will be allocated 8GiB memory instead of 4. I'm assuming this argument works in the same way with the `cpg_utils.get_batch()` batch instantiation. 

With this PR, we will be able toggle highmem workers through a config option. Ugly implementation for now, will make it nicer in CPG flow.

- Also removes unused bool arg to the annotate_cohort job `long_read` - relic of an earlier pipeline iteration

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.36.34
+current_version = 1.36.35
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.36.34
+  VERSION: 1.36.35
 
 jobs:
   docker:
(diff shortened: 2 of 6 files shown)


MattWellie:
Your best approach will be using the `init_batch_args` argument to modify this - because the `annotate_cohort` process is in a totally separate job, you need to get these updates into the `init_batch` call inside the Job.

Example https://github.com/populationgenomics/production-pipelines/blob/d4f82a04ef800ca4e705aabce09347c49252a2d2/cpg_workflows/stages/large_cohort.py#L74-L97 

cpg-utils reference: https://github.com/populationgenomics/cpg-utils/blob/cf249f226fe54c8b7560cb01ae69b4bc8acb26ab/cpg_utils/hail_batch.py#L649-L662

EddieLF:
@MattWellie Excellent suggestion and linked example. I've copied the setup from the sampleqc code, except I'm pulling the config dict from `workflow.annotate_cohort` for a bit more granularity.

MattWellie (approved):

PR #1235, opened 2025-05-28
Adding ability to assign worker cores to batch

michael-harper:
Allowing finer control over both memory and storage allotted to worker nodes during Combiner batches


Alexander-Stuckey (approved):
Looks good to me

PR #1239, opened 2025-06-02
Revert addition of worker_cores arg in combiner init_batch

EddieLF:
Reverts https://github.com/populationgenomics/production-pipelines/pull/1235/files

Removes the combiner.worker_cores config arg to init_batch in combiner.py, because it's not used and causes error on startup. More detail in @MattWellie's comment below.

~Sets the worker cores in the rd_combiner stage to 1, which is [the hail default](https://github.com/populationgenomics/hail/blob/main/hail/python/hail/context.py#L319).~

~Without the worker cores in the default config, the combiner will fail to start unless the user adds it to their custom config included in their analysis-runner submission.~

e.g. https://batch.hail.populationgenomics.org.au/batches/616323/jobs/1
```
Traceback (most recent call last):
  File "<string>", line 32, in <module>
  File "/cpg_workflows/jobs/rd_combiner/combiner.py", line 55, in run
    worker_cores=config_retrieve(['combiner', 'worker_cores']),
  File "/usr/local/lib/python3.10/site-packages/cpg_utils/config.py", line 253, in config_retrieve
    raise ConfigError(message)
cpg_utils.config.ConfigError: Key "worker_cores" not found in {'force_new_combiner': False,  ...
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.2
+current_version = 1.37.3
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 env:
-  VERSION: 1.37.2
+  VERSION: 1.37.3
 
 jobs:
   docker:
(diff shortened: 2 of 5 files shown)


MattWellie (approved):
Equally happy with dropping this - it was added in error, and AFAIK didn't solve the problem it was aiming to solve. We already have the ability to set worker memory, which Hail matches to cores using it's special little equation, so I'm not sure it helps us

PR #1241, opened 2025-06-06
User defined VDS path

michael-harper:
This PR adds support for specifying a custom VDS path in the Combiner stage via the `combiner.vds_path` config option. If a user provides this path, it will be used directly; otherwise, the default path logic is applied as before.

Details
- Updated `Combiner.expected_outputs()` to check for a `vds_path` in the `combiner` config.
- If `vds_path` is provided, it is used as the output path for the VDS.
- If not provided, the original logic for constructing the VDS path based on cohort and version is used.
- This change allows users to reference VDS files with custom or legacy path structures, improving flexibility and reproducibility.

## Motivation
Previously, the VDS output path was always constructed based on the current cohort and version naming conventions. This made it difficult to reference VDS files generated with older or custom path structures. By allowing users to specify a direct path, we avoid issues with pathing updates and make it easier to reuse or reference previous VDS outputs.


katiedelange (approved):
LGTM

PR #1245, opened 2025-06-09
VDS re-code: 133 in 134 out

MattWellie:
Due to Hail 134 issues, any pre-134 VDSs being used in combining must first be written back out in 134 format, then combined. 

There's a range of possible actions here, from re-combining everything on 134, to overwriting all extant 133 data as 134 and then combining on top. 

This straddles the middle -
- reads in 133 data, and writes to a new location using 134. Importantly the existing 133 data is unmodified, so the analysis entry integrity is maintained. Just in case... y'know, more surprises turn up with this new data
- the combiner process is edited so that it can combine data using paths in config instead of analysis entries. RD never used the existing analysis-ID option anyway, and this method prevents the need to register this duplicate data in metamist (along with all the contained sample IDs) in order to make use of it.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.3
+current_version = 1.37.4
 commit = True
 tag = False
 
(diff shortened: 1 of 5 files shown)


Alexander-Stuckey (approved):
Looks good to me

PR #1247, opened 2025-06-10
Add new STRipy loci to defaults.toml

EddieLF:
Adds in two new STRipy loci to the default target loci, based on conversation [here](https://centrepopgen.slack.com/archives/C085XN3U58S/p1745886106989599).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.4
+current_version = 1.37.5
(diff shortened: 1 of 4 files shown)


cassimons:
@rociorius can you confirm that you are OK for these two loci to be in the default list?

rociorius:
yes [ABCD3](https://stripy.org/database/ABCD3) [THAP11](https://stripy.org/database/THAP11) should be ok to include in default list

PR #1248, opened 2025-06-10
Prevent running of migrated workflows in this codebase

MattWellie:
Purely out of force of habit I just ran the production-pipelines version of a migrated workflow. Not only did this run old code, it also might contain some bugs which were found and patched during migration, without the changes being backported to this codebase (because why).

It makes sense to disable this possibility, without having to delete all the code in this repository. Adding an extensible list of migrated workflows, an informative error message, and leaving the underlying code intact, seems like a quick and easy compromise

... I think the Dragen workflow might also be fully migrated and in use, but that had a different runner script. Feel free to extend this initial list if there are other successful migrations to disable.


Alexander-Stuckey (approved):
Looks good to me

PR #1251, opened 2025-06-11
Stripy loci by dataset

EddieLF:
### Minor change

- Removes `PRNP` from the default target loci based on [this conversation](https://centrepopgen.slack.com/archives/C085XN3U58S/p1745886106989599).

### Major changes

- Adds new function to get stripy config overrides for sequencing groups in specific datasets
  - For the target loci, this can be additional target loci for a dataset, or specific loci only for a dataset
  - `[stripy.validation.target_loci] = "ABCD1"` will add `ABCD1` to the list of target loci, alongside the defaults
  - `[stripy.validation.target_loci] = "ABCD1"` & `[stripy.validation.target_loci_override] = true` will mean `ABCD1` is the only target loci used
- Also adds this dataset specific functionality to the options `stripy.custom_loci_path` and `stripy.config` (which modifies the `config.json` inputs - not to be confused with the `[stripy]` config options in the analysis-runner toml)
- Create the output files if they don't exist, to prevent output writing errors like [this](https://batch.hail.populationgenomics.org.au/batches/618374/jobs/1) (when there are no loci found, the log and json won't be created) - but we still want the job to succeed so the html output is written to the bucket.

### Future scope (not in this PR)

- Overwrite the target_loci and other stripy configs at a sequencing group level rather than dataset level
- e.g. `[stripy.CPG12345.target_loci] = "ABCD1"` will add the locus `ABCD1` to the stripy target_loci for CPG12345 only


cpg_workflows/stages/stripy.py
```diff
@@ -21,6 +21,35 @@
...
+    """
+    Retrieve the STRipy config for a given key. If a dataset is provided, it will first
+    check for a dataset-specific config before falling back to the global config.
+
+    **NOT to be confused with the stripy config.json file used by the stripy job**
+    """
+    if dataset and config_retrieve(['stripy', dataset, key], default=None) is not None:
+        return config_retrieve(['stripy', dataset, key], default=None)
+    else:
+        return config_retrieve(['stripy', key], default=None)
```
MattWellie, line 34:
```suggestion
    if dataset and (result := config_retrieve(['stripy', dataset, key], default=None)) is not None:
        return result
    return config_retrieve(['stripy', key], default=None)
```

Just a couple of nits here - 
- there's a silly linting rule somewhere about not needing an `else` after a `return`
- instead of executing `config_retrieve(['stripy', dataset, key], default=None)` twice if successful you can bank the result using a walrus operator

MattWellie (approved):
one pretty minor nit, but otherwise looks good. It's increased complexity, but the self-documenting config entry is good

cpg_workflows/stages/stripy.py
EddieLF: Thanks Matt, good suggestion. I agree we're increasing complexity but also providing a lot more flexibility, so each dataset can freely specify their own loci.

PR #1254, opened 2025-06-12
Vcf from mt fix

EddieLF:
The expected outputs from `AnnotateDatasetSmallVariantsWithHailQuery` are not a dict, whilst in the `seqr_loader` pipeline (which I copied and pasted the mt to VCF stage from) they are. This change removes the dict key from the `inputs.as_path` call to prevent the error seen [here](https://batch.hail.populationgenomics.org.au/batches/618427/jobs/1).

```
File "/production-pipelines/cpg_workflows/workflow.py", line 154, in _get
    raise ValueError(f'{self.stage}: {self.data} is not a dictionary, can\'t get "{key}"')
ValueError: AnnotateDatasetSmallVariantsWithHailQuery <- [AnnotateCohortSmallVariantsWithHailQuery, SubsetMatrixTableToDatasetUsingHailQuery]: gs://cpg-heartkids-main/mt/AnnotateDatasetSmallVariantsWithHailQuery/5fae2753fb994fee344bbfac04aed6e66b3e3a_5804-heartkids.mt is not a dictionary, can't get "mt"
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.7
+current_version = 1.37.8
 commit = True
 tag = False
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.8
+current_version = 1.37.9
 commit = True
 tag = False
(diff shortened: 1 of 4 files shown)


MattWellie (approved):
I should have spotted that 🤦

PR #1257, opened 2025-06-17
Write unique CNV & SV VCF Path

MattWellie:
Major oversight from me - the CNV & SV VCFs which are now the starting point for Talos were being written to a really generic path (i.e. `gs://dataset-bucket/annotated_svs.vcf.bgz`). This contained nothing identifying to a specific run/cohort/multicohort, so this stage has only ever been run once.

This change inserts a fully qualified Path (including dataset, multicohort hash, workflow name, stage name), so it's guaranteed to run with each gCNV/GATK-SV pipeline run

I already fixed this in the CPG-Flow migration of GATK-SV, and just added it to the [gCNV migration](https://github.com/populationgenomics/cpg-flow-gcnv/pull/2)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.37.9
+current_version = 1.37.10
 commit = True
 tag = False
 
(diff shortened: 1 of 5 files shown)


cassimons (approved):

Alexander-Stuckey (approved):
Looks good to me

PR #1259, opened 2025-06-18
allow passing in a VDS via config

Alexander-Stuckey:
Pass in a VDS via config, in order to escape from the Hail 134 combiner woes (i.e. we have converted the 133 VDS to 134, but not registered it in Metamist since it'll be transient).


MattWellie (approved):
Are you going to be setting this as a list in config, or listifying a string?

Alexander-Stuckey:
> Are you going to be setting this as a list in config, or listifying a string?

List in config

PR #1260, opened 2025-06-18
Rename incompatible sampleqc fields

joshuamschmidt:
We got an error running the `ancestry` stage [related](https://batch.hail.populationgenomics.org.au/batches/617448/jobs/2)  to dropping columns that were not in one of the `sample_qc.ht`, due to changes in the `sample_qc.ht` schema in different cohorts.

These changes could be:

- Same data source, different row.field name
- Presence absence of row.field
- Asking to drop a row when it is not present

This PR makes two substantive changes as a catch all for all 3:

1. Adding a function, `sync_sampleqc_hts`, and logic, to explicitly drop ow fields missing from the join intersect set of row fields.

I tested this in a notebook:
```
import hail as hl
hgdp_qc   = hl.read_table("gs://cpg-hgdp-main-analysis/large_cohort/v1-0/sample_qc.ht")
ourdna_qc = hl.read_table("gs://cpg-ourdna-main-analysis/large_cohort/c837a1558f7f0c4540f9c68f99e3fa5b2d89ad_101/sample_qc.ht")

ourdna_qc.row.dtype.fields == hgdp_qc.row.dtype.fields
# False
ourdna_qc, hgdp_qc = sync_sampleqc_hts(ourdna_qc, hgdp_qc)
ourdna_qc.row.dtype.fields == hgdp_qc.row.dtype.fields
# True
```

2. When passing `drop_columns`, we first iterate of the row.fields to see if the request columns are still present in the merged `sample_qc.ht`.

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -32,8 +32,8 @@ def sync_sampleqc_hts(dataset_ht: hl.Table, reference_ht: hl.Table) -> tuple[hl.
     if dataset_ht.row.dtype.fields == reference_ht.row.dtype.fields:
         return dataset_ht, reference_ht
 
-    dataset_rows = [i[0] for i in dataset_ht.row.items()]
-    reference_rows = [i[0] for i in reference_ht.row.items()]
+    dataset_rows = [f for f in dataset_ht.row.dtype.fields]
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
This looks good to me. As a general statement the Ancestry stage is disgusting and has had config parameter after config parameter used to help analysts achieve what they want. The mental gymnastics required to step through this stage is too big.

michael-harper (approved):

PR #1261, opened 2025-06-18
Allow writing multiple files into a single analysis record

EddieLF:
- Switch to using `outputs` in the calls to `create_analysis` to allow for multiple output files per analysis record in Metamist (see the expected syntax discussed below)
- Treat a `Stage` class outputs like a dictionary most of the time for better compatibility with this 
- Write multiple of a stage's `expected_outputs` to a single analysis record, for stages specified through the config option `workflow.merge_analyses_stages = [ 'Stage1', 'Stage2', ... ]` 

I have tested this update with a few stages (`SomalierPedigree`, `CramMultiQC`, `GvcfMultiQC`) and it seems to work fairly well. 

https://batch.hail.populationgenomics.org.au/batches/618931/jobs/1
https://batch.hail.populationgenomics.org.au/batches/618932

In this run, I used the config option `workflow.merge_analyses_stages = [ "CramMultiQC", "GvcfMultiQC",]` to write multiple files to a single record for the two MultiQC stages, but not for `SomalierPedigree`. 

This makes the MultiQC stage write a single analysis record with both outputs `html` and `json`. While the Somalier stage creates four separate analysis records, one for each of it's analysis keys.

--- 

### TODO

There's still a need to refactor some other uses of `update_analysis_meta` across a few different stages to make it work with dict outputs. 
More testing is needed
Future expansion of this might be to also use the meta dict for each output, as well as the secondary files. 

---

### Syntax for writing analysis outputs

For stages who's expected outputs are just a string, they get coaxed into a dictionary like `{ 'output' : 'gs://bucket/file.html' }` as they are passed through the pipelines, but they are given to the Metamist API as the string alone.
```
outputs = 'gs://bucket/file.html'

analysis_api.create_analysis(
  Analysis(outputs=outputs)
)

# Writes analysis records like
Analysis {
  "id": 315006,
  "output": "gs://bucket/file.html"
  "outputs": {
      "path": "gs://bucket/file.html",
      "file_checksum": "QWy1CQ==",
      "size": 7386642,
      "secondary_files": {}
    }
  }
}
```

For stages with multiple output files that we want to write to the same analysis, we can use the `outputs` like so:
```
outputs = {
  'html': { 'basename': 'gs://bucket/file.html' },  # 
  'json': { 'basename': 'gs://bucket/file.json' }   # specific syntax using 'basename'
}

analysis_api.create_analysis(
  Analysis(outputs=outputs)
)

# Writes analysis records like
Analysis {
  "id": 315007,
  "outputs": {
    "html": {
      "path": "gs://bucket/file.html",
      "file_checksum": "QWy1CQ==",
      "size": 7386642,
      "secondary_files": {}
    },
    "json": {
      "path": "gs://bucket/file.json",
      "file_checksum": "Ax4nCN==",
      "size": 86253421,
      "secondary_files": {}
    }
  }
}
```
Output names as the keys to a second dict containing the path as the dict value for the key `'basename'`. This specific syntax will trigger the db to query GCS for each blobs metadata, and it will fill in each analysis file output record with this additional information.

### DIFF
--- a/cpg_workflows/metamist.py
+++ b/cpg_workflows/metamist.py
@@ -471,7 +471,7 @@ class Metamist:
 
     def create_analysis(
         self,
-        output: Path | str,
+        outputs: dict[str, str | Path],
         type_: str | AnalysisType,
         status: str | AnalysisStatus,
         sequencing_group_ids: list[str],
@@ -488,10 +488,16 @@ class Metamist:
         if isinstance(status, AnalysisStatus):
             status = status.value
 
+        if len(outputs) == 1:
+            # If output is a single file, just use the string path
+            outputs = {'basename': str(next(iter(outputs.values())))}
+        else:
+            outputs = {k: {'basename': str(v)} for k, v in outputs.items()}
+
         am = models.Analysis(
             type=type_,
             status=models.AnalysisStatus(status),
-            output=str(output),
+            outputs=outputs,
             sequencing_group_ids=list(sequencing_group_ids),
             meta=meta or {},
         )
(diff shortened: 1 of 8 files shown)


MattWellie:
I really like this proposed functionality, but there's too much entrenched code in prod-pipes and now all the CPG-Flow repos... 

A fully backwards-compatible change in behaviour was always going to be difficult, and an interface change would require more buy-in than I think this got. After a year of getting daily notifications I think I'm going to unsubscribe, and close 😢

This `output -> outputs` shift, and secondary files & types will be supported in NextFlow, but not in CPG-Flow.

PR #1262, opened 2025-06-18
added steps to preserve cols in the datset sample_qc.ht which are abs…

joshuamschmidt:
A better attempt than PR [1260](https://github.com/populationgenomics/production-pipelines/pull/1260) at fixing the issue that `large_cohort ancestry` will drop columns in the cohort/dataset s`ample_qc.ht` that are absent from the background (hgdp and 1kg) `sample_qc.ht`.
This can lead to errors when code expects those columns to exist e.g. `get_config()['large_cohort']['pca_background'].get('drop_columns')`

### DIFF
--- a/cpg_workflows/large_cohort/ancestry_pca.py
+++ b/cpg_workflows/large_cohort/ancestry_pca.py
@@ -70,6 +70,8 @@ def add_background(
                 metadata_tables.append(sample_qc_background)
             metadata_tables = hl.Table.union(*metadata_tables, unify=allow_missing_columns)
             metadata_tables = reorder_columns(metadata_tables, sample_qc_ht)
+            metadata_tables = hl.Table.union(*[metadata_tables, sample_qc_ht], unify=allow_missing_columns)
+
             background_mt = background_mt.annotate_cols(**metadata_tables[background_mt.col_key])
             if populations_to_filter:
                 logging.info(f'Filtering background samples by {populations_to_filter}')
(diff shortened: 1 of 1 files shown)


cpg_workflows/large_cohort/ancestry_pca.py
```diff
@@ -70,6 +70,8 @@ def add_background(
                 metadata_tables.append(sample_qc_background)
             metadata_tables = hl.Table.union(*metadata_tables, unify=allow_missing_columns)
             metadata_tables = reorder_columns(metadata_tables, sample_qc_ht)
+            metadata_tables = hl.Table.union(*[metadata_tables, sample_qc_ht], unify=allow_missing_columns)
```
michael-harper, line 73: Pair programmed inside a jupyter notebook with @joshuamschmidt.

michael-harper (commented):
This will synchronise the columns between the background sample_qc hail tables and the dataset currently being analysed sample_qc hail table.

There is the issue of the `drop_columns` later on in the script. At this stage the user will have to know what columns are present inside the background VDS/MT/HT as well as the datasets VDS/MT/HT before we provide a column to drop. Because if the column has been removed during `reorder_columns()` then `ht.drop()` will throw an error.

I have added more informative error message so the user will know what fields are available to be dropped. I think we would still want the stage to fail if a column that we want dropped is not actually able to be dropped.

@joshuamschmidt what are your thoughts?

joshuamschmidt (commented):
LGTM

michael-harper:
Agreed. There is a `can_reuse()` in the code before densification to avoid unnecessary compute, though it is in the `tmp` bucket that will delete after a week or so.

```
can_reuse(background_mt_checkpoint_path):
    logging.info(f'Reusing densified background mt from {background_mt_checkpoint_path}')
    background_mt = hl.read_matrix_table(str(background_mt_checkpoint_path))
```

michael-harper (approved):

PR #1264, opened 2025-06-25
Generalise inputs and outputs

joshuamschmidt:
This PR expands functionality of `large_cohort/sites_table_job.py`:

1. ht without `info.AS_FilterStatus` can be used as `--vqsr-table-path` - so long as it has `info.filters`
2. re-implemented ability to sub-sample rows before LD pruning.
3. RE 2 - no longer use const  `NUM_ROWS_BEFORE_LD_PRUNE`, but have options `--subsample` (`bool`) and `--subsample-n` (`int`)


michael-harper (approved):
Code reviewed with @joshuamschmidt, LGTUs

michael-harper (commented):
Code reviewed with @joshuamschmidt, LGTUs

PR #1265, opened 2025-06-25
Script to compile samtools stats for genome SGs

EddieLF:
Script to extract summary stats from multiple samtools-stats files and write the extracted information into a flat file.

**This script**:
- Queries metamist for short-read, genome sequencing groups from a list of datasets
  - Datasets can be specified on the command line, set to all seqr datasets, or all "my datasets"
- Constructs the path to each of the sequencing group samtools-stats file, like:
  - `gs://cpg-dataset-main/qc/samtools_stats/SG_ID.samtools-stats`
- Reads each of these stats files and extracts the summary values (rows that begin with `'SN'`)
- Writes the stats to a tsv file, with headers: `'dataset', 'sg_id', **stats`

The script saves the results to a tsv file, either local or in a bucket.

If the script is run again with the same output file path, it will read back the existing output file, and exclude any sequencing groups whose results already exist. So only new sequencing groups will be added to the flat file.

Output format (truncated, there are 39 stat columns total)
```
dataset        	 sg_id    	 raw total sequences	 filtered sequences	 sequences 	 is sorted	 1st fragments	 last fragments	 reads mapped
dataset1       	 CPG123456	           803580730	                  0	  803580730	         1	     401790365	      401790365	    802471510
dataset1       	 CPG234567	          1263709892	                  0	 1263709892	         1	     631854946	      631854946	   1223342278
dataset2       	 CPG345678	           768780058	                  0	  768780058	         1	     384390029	      384390029	    692234340
```

Example output for a few datasets here:
```
gs://cpg-seqr-test-analysis/samtools_stats/cardinal_samtools_stats_2025-06-25.tsv
```

### DIFF
--- a/scripts/compile_samtools_stats.py
+++ b/scripts/compile_samtools_stats.py
@@ -195,43 +195,42 @@ def get_existing_data(
     """
     Read an existing output file and return its contents as a list of lines.
     """
+    existing_data: list[str] = []
     exists = to_path(output_file).exists()
-    if exists:
-        logger.info(f'Output file {output_file} already exists. Appending to it.')
-        # Download the existing file and read its contents, then append the new stats to it and write it back
-        if output_file.startswith('gs://'):
-            # If the output file is in GCS, download it to a temporary local file
-            storage_client = storage.Client()
-            bucket = storage_client.bucket(bucket_name)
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
All looks reasonable, the pipeline doesn't track these corresponding metrics files in Metamist afaik, so we should handle that in the migration to make this type of operation simpler

PR #1267, opened 2025-07-01
Single sample hap.py

katiedelange:
Adds a script to perform hap.py validation on a single NA12878 sample gVCF.


cpg_workflows/scripts/NA12878_validation.py
```diff
@@ -0,0 +1,106 @@
...
+    vcf_path: str,
+    output_prefix: str,
+    region: str | None = None,
+):
+    """Run hap.py on a single-sample gVCF (NA12878) using Truth data from config."""
+
+    batch_instance = get_batch()
+
+    # region: run BCFtools to filter the gVCF
+    bcftools_job = batch_instance.new_job(f'Run BCFtools on {vcf_path!s}')
```
Alexander-Stuckey, line 23: `!s` casts the variable to a string, but `vcf_path` is already a string. So it's probably not needed. And so on for other instances.

cpg_workflows/scripts/NA12878_validation.py
```diff
@@ -0,0 +1,106 @@
...
+    bcftools_job.image(config.config_retrieve(['images', 'bcftools_120'])).memory('10Gi').storage('100Gi')
```
Alexander-Stuckey, line 24: Do we actually need `100Gi` of storage? Seems a lot for two GVCFs.

Alexander-Stuckey (approved):
Couple minor points. looks good.

PR #1268, opened 2025-07-03
remove call to _count_rows after sub-sampling

joshuamschmidt:
calls to `ht.count_rows` can be expensive. 
while we need to count n rows before subsampling, calling it afterwards to show for logging that we indeed end up with the n rows we required is unnecessary.


silkm (approved):
LGTM

PR #1269, opened 2025-07-07
Add ability to adjust batch resources for Coverage pipeline

michael-harper:
Implementing same logic to adjust batch worker/driver resources as in SampleQC Stage.

Worker and driver resources can be adjusted via the `[workflow]` config parameter. When using `query_command()` these parameters need to be passed via the `init_batch_args` for the resource adjustments to propagate through to the worker/driver nodes.


katiedelange (approved):
LGTM

PR #1270, opened 2025-07-08
Prep gvcf for happy

katiedelange:
Replace the bcftools rinse with pre.py conversion of the gVCF to a VCF before inputting to hap.py, enabling more thorough handling of <NON_REF> genotypes. 

Test run here: https://batch.hail.populationgenomics.org.au/batches/627499


MattWellie (approved):
Delicious

PR #1271, opened 2025-07-08
Coverage reuse

michael-harper:
If a stage has a dictionary as its `expected_outputs` (as `GenerateCoverageTable` does) and if, in a previous run, one or several of those outputs failed to be generated (due to batch reasons), then the workflow logic is: when submitting a follow-up run it will detect that at least one expected output doesn't exist and will submit jobs to generate **all** outputs again.

This is not desired behaviour. To avoid having all intermediate split coverage tables be generated if just one fails this PR implements the `can_reuse()` function that checks the individual coverage tables path and returns early if that path already exists


katiedelange (approved):
LGTM

PR #1272, opened 2025-07-09
Remove init_batch within coverage script

michael-harper:
Remove init_batch() from within the script as it is overriding the init_batch being called by query_command() that contains the init_batch resource overrides


joshuamschmidt (approved):
seems good to me

PR #1278, opened 2025-07-14
Shard coverage calculation per sample

michael-harper:
Costs have blown out with larger number of samples when calculating coverage. Adding the ability to shard coverage calculation per config-specified sample number.


Alexander-Stuckey (approved):
Looks good to me

PR #1279, opened 2025-07-15
Allowing Coverage table generation for scatter_count=1

michael-harper:
`get_intervals()` function has a special case where if scatter_count = 1 it just reads in the interval_list as a resource file instead of saving to a path. This unfortunately means the path the stage expects will not exist.


Alexander-Stuckey (approved):
Looks good to me

PR #1283, opened 2025-07-17
checkpoint after filtering and repartitioning vds

Alexander-Stuckey:
Repartition the VDS after filtering to exome intervals.
Checkpoint the VDS after repartitioning.
Remove an additional logging line that caused much lag in the logs, due to length.
Slightly adjust sample filtering to allow for running on subsets of samples or the entire dataset without having to know how large the dataset it.


michael-harper (approved):
LGTM

PR #1286, opened 2025-07-21
Handle multiallelics correctly when creating a dense subset

katiedelange:
Avoid splitting multiallelics before filtering to the pre-determined QC sites, which include only biallelic SNPs. Splitting will enable us to match cleanly against the reference sites, but causes us to retain sites that are truly multiallelic in our cohort but using only the reference allele, giving the false appearance that they were biallelic.

Having filtered, repartition the matrix table using a configurable number of partitions (default of 10, as our default sites are exome-only).

### DIFF
--- a/configs/defaults/large_cohort.toml
+++ b/configs/defaults/large_cohort.toml
@@ -54,6 +54,7 @@ num_pcs = 4
 # demanding higher resources for all operations as standard. Examples below
 
 [large_cohort]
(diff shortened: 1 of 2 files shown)


Alexander-Stuckey (approved):
Looks good to me.

Alexander-Stuckey (approved):
Looks good to me

PR #1291, opened 2025-07-23
Reduce burden of filtering lcr and seg_dups in sampleqc

joshuamschmidt:
previously would loop over each filter and: 
 - read ht
 -  filter + checkpoint **each loop**
 - vds.variant_data.count()

This changes so that:
-  loop over filter hts, and union them.
- filter and checkpoint **once**
- just log that we are done filtering, and not count remaining variants (removed vds.variant_data.count())

Still unknown to me if this step could be improved (re-partitioning):
```
       tmp_variant_data = vds.variant_data.filter_rows(
            hl.is_defined(interval_table[vds.variant_data.locus]),
            keep=False,
        )
        vds = VariantDataset(reference_data=vds.reference_data, variant_data=tmp_variant_data).checkpoint(
            str(vds_tmp_path),
            overwrite=True,
```

### DIFF
--- a/cpg_workflows/large_cohort/sample_qc.py
+++ b/cpg_workflows/large_cohort/sample_qc.py
@@ -105,24 +105,29 @@ def impute_sex(
 
     # Pre-filter here and setting `variants_filter_lcr` and `variants_filter_segdup`
     # below to `False` to avoid the function calling gnomAD's `resources` module:
-    for name in ['lcr_intervals_ht', 'seg_dup_intervals_ht']:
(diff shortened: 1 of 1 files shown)


katiedelange:
The error was just suggesting that you couldn't run hl.Table.Union without a base table ('self') - i.e. it needed to be in the form [actual hail table].union(). So this change just takes the first table in the list as the base and then adds all the other ones in to that.

katiedelange (approved):
LGTM

PR #1292, opened 2025-07-24
Remove low quality samples from PC calculations

katiedelange:
Rather than needing to manually add the low quality samples to the relateds_to_drop hail table in order to exclude them from PCA, we automatically pull this from the sample QC hail table.

These samples will still get projected into PCA space and be present in the final sample_qc.ht output, but will not be used for PCA generation.


joshuamschmidt (approved):
LGTM

PR #1293, opened 2025-07-25
Allow certain datasets (e.g. genomes vs exomes) to be prioritised over others in relatedness filtering

katiedelange:
Optionally rank the samples based on a dataset ordering provided in the config (samples in datasets listed first in the list will be preferentially kept).


michael-harper (approved):
This looks good to me. Cool way of prioritising

PR #1295, opened 2025-07-30
Add optional disk size parameter to index_files script

EddieLF:
Adds a new CLI option `--disk-size`, used to specify the storage allocated to the indexing job


amyminiter (approved):
Thanks for updating this, Ed 🌟

PR #1296, opened 2025-07-31
Increase checkpointing in make site only vcf

katiedelange:
- Increase checkpoint in MakeSiteOnlyVcf
- Include AS_lowqual and lowqual in the 'info' struct so hl.export_vcf doesn't drop them


MattWellie (approved):
LGTMatt

PR #1298, opened 2025-08-01
Add binned summary hts

silkm:
PR to add functionality to generate `binned_summaries.ht` necessary for the following stage.

The functions are a hacky implementation of Katie's notebook functions, and definitely yet another target for future pipeline work.

### DIFF
--- a/cpg_workflows/stages/large_cohort.py
+++ b/cpg_workflows/stages/large_cohort.py
@@ -693,7 +693,7 @@ class VariantBinnedSummaries(CohortStage):
             if workflow_config.get(key):
                 init_batch_args[key] = workflow_config[key]
 
-        happy_vcf_path = config_retrieve(['large_cohort', 'happy_concordance_vcf'])
+        happy_vcf_path = config_retrieve(['large_cohort', 'happy_vcf_path'])
         n_bins = config_retrieve(['large_cohort', 'n_bins'], default=None)  # n = 100 defaults in the stage itself
         fam_stats_ht_path = config_retrieve(['large_cohort', 'fam_stats_ht_path'], default=None)
         truth_sample_concordance_path = config_retrieve(['large_cohort', 'truth_sample_concordance_path'], default=None)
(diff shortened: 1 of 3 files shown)


cpg_workflows/stages/large_cohort.py
```diff
@@ -660,10 +660,67 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+                init_batch_args[batch_key] = 'highmem'
+        # Cores parameter
+        for key in ['driver_cores', 'worker_cores']:
+            if workflow_config.get(key):
+                init_batch_args[key] = workflow_config[key]
+
+        happy_vcf_path = config_retrieve(['large_cohort', 'happy_vcf_path'])
+        n_bins = config_retrieve(['large_cohort', 'n_bins'], default=100)  # FIXME default also set in function
+        fam_stats_ht_path = config_retrieve(['large_cohort', 'fam_stats_ht_path'], default=None)
+        use_truth_sample_concordance = config_retrieve(['large_cohort', 'truth_sample_concordance_path'], default=True)
```
katiedelange, line 699: This bit doesn't quite make sense to me - should that config parameter be a boolean called `use_truth_sample_concordance` instead, or was the idea that you would construct a boolean based on if the happy_vcf_path was provided?

cpg_workflows/stages/large_cohort.py
```diff
@@ -660,10 +660,67 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+
+        j.command(
+            query_command(
+                variant_binned_summaries,
+                variant_binned_summaries.create_binned_summary.__name__,
+                str(inputs.as_path(cohort, LoadVqsr)),
+                # "gs://cpg-bioheart-test/large_cohort/browser-testing-data/vqsr.ht",
```
katiedelange, line 706:
```suggestion
```

cpg_workflows/stages/large_cohort.py
```diff
@@ -660,10 +660,67 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
+                str(happy_vcf_path),
+                str(self.expected_outputs(cohort)),
+                int(n_bins),
+                fam_stats_ht_path,
+                str(use_truth_sample_concordance),
```
katiedelange, line 711: Does converting to a string work here? When the function expects a bool?

katiedelange (commented):
A few minor comments around one parameter, but otherwise looks great!

cpg_workflows/stages/large_cohort.py
silkm: I'm not sure why this didn't raise a type error in the function 🤔 I've removed the type casting for the args to be consistent with how the other stages are set up

cpg_workflows/stages/large_cohort.py
silkm: Great catch! forgot to adjust the name here

PR #1302, opened 2025-08-04
Specify bin to use for score threshold selection

katiedelange:
Enable the bin id to use for score selection to be specified via config parameters.


michael-harper (approved):
LGTM

PR #1310, opened 2025-08-10
Reduce Default Resources for GenerateCoverage and GenerateAlleleNumber

michael-harper:
After some considerable refactoring these stages no longer require such intensive resources. We are defaulting to using a single worker core and standard memory unless specifically requested


katiedelange (approved):
LGTM, thanks Harper

PR #1311, opened 2025-08-10
Removing MergeCoverageTables stage and its associated function from compute_stats_for_all_sites.py

michael-harper:
Stage no longer required after refactoring of coverage


katiedelange (approved):
LGTM

PR #1312, opened 2025-08-11
Adding joint_frequency table generation

michael-harper:
Adding the joint frequency table to the browser table. 

As discussed in [this Slack thread](https://centrepopgen.slack.com/archives/C02KQ4XMHQ9/p1754453580685289) where @joshuamschmidt  found a variant that is a singleton in exomes and not present in genomes and is reported as `No Variant` in WGS without any metrics, when the metrics should actually be Allele Count = 0, Allelle Number = X. This is a known issue where the site is likely well-covered in the other callset (WGS vs WES) but is homozygous ref for all individuals. This is important for clinical users to distinguish "confidently not present" from "site not called", and the proper solution is the joint table as implemented in v4. As such, this PR aims to generate the joint frequency table.

We first generate the joint frequency table in a new stage called `JointFrequencyTable`. This stage also outputs the `contingency_ht` and `chm_ht` which are used to annotate the final browser table inside the `joint` field.

The second step is to take the outputs of this stage and use them inside `PrepareBrowserTable`.

At this stage I am unable to test if this runs on batch due to lack of test data (I need a cohort of AGDD in test as well as test sequencing groups, stage outputs, etc.). However, this has worked great in a notebook.

Open question about the region flags. I have left it commented out as I am not sure if we have interval QC metrics available? @joshuamschmidt @katiedelange. I've left this portion of the code commented out.
```python
# intervals = [(c, calling_intervals(c, 0).ht()) for c in ["broad", "ukb"]]
#     intervals = [(f"{c}_capture", i) for c, i in intervals] + [
#         (f"{c}_calling", adjust_interval_padding(i, 150)) for c, i in intervals
#     ]

    region_expr = {
        # "fail_interval_qc": (~interval_qc_pass(all_platforms=True).ht()[ht.locus].pass_interval_qc),
        # **{f"outside_{c}_region": hl.is_missing(i[ht.locus]) for c, i in intervals},
        **{
            f"not_called_in_{d}": hl.is_missing(ht[f"{d}_freq"]) | hl.is_missing(ht[f"{d}_freq"][1].AN)
            for d in DATA_TYPES
        },
    }
```


Alexander-Stuckey (commented):
Happy with this, but will also let Katie give it a look over

cpg_workflows/large_cohort/joint_frequencies.py
```diff
@@ -0,0 +1,779 @@
...
+from gnomad.utils.release import make_freq_index_dict_from_meta
+
+logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
+logger = logging.getLogger("compute_combined_faf")
+logger.setLevel(logging.INFO)
+
+
+GEN_ANC_GROUPS_TO_REMOVE_FOR_GRPMAX = {"Unassigned"}
+DATA_TYPES = ["exomes", "genomes"]
+FAF_POPS = ['CSA', 'EAS', 'EUR']  # We can only perform this test for populations in both datasets
```
katiedelange, line 21: Can we either make this a config parameter or, ideally, compute it ourselves from the metadata on the tables?

cpg_workflows/large_cohort/joint_frequencies.py
```diff
@@ -0,0 +1,779 @@
...
+    genome_all_sites_ht = hl.read_table(str(genome_all_sites_path))
+    exome_all_sites_ht = hl.read_table(str(exome_all_sites_path))
+
+    # TODO (from gnomad): Need to resolve the type difference.
+    genome_freq_ht = genome_freq_ht.annotate(
+        freq=genome_freq_ht.freq.map(lambda x: x.annotate(homozygote_count=hl.int32(x.homozygote_count))),
+    )
+    exome_freq_ht = exome_freq_ht.annotate(
+        freq=exome_freq_ht.freq.map(lambda x: x.annotate(homozygote_count=hl.int32(x.homozygote_count))),
+    )
```
katiedelange, line 730: I'm pretty sure we don't have this type issue in this data. 

However, we do have an issue with the `age_hists`  not currently sitting within the `histograms` struct which we do need to fix before running `extract_freq_info` - see the notebook in [this Slack thread](https://centrepopgen.slack.com/archives/C08JL4YHPNE/p1754880142137519).

cpg_workflows/large_cohort/browser_prepare.py
```diff
@@ -938,9 +938,221 @@ def prepare_gnomad_v4_variants_helper(
...
+
+    # Add capture and calling intervals as region flags.
+    # intervals = [(c, calling_intervals(c, 0).ht()) for c in ["broad", "ukb"]]
+    # intervals = [(f"{c}_capture", i) for c, i in intervals] + [
+    #     (f"{c}_calling", adjust_interval_padding(i, 150)) for c, i in intervals
+    # ]
+
+    region_expr = {
+        #     "fail_interval_qc": (~interval_qc_pass(all_platforms=True).ht()[ht.locus].pass_interval_qc),
+        #     **{f"outside_{c}_region": hl.is_missing(i[ht.locus]) for c, i in intervals},
```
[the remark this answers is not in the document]
michael-harper: Is it safe to assume we'll always have more than one capture interval or should we always perform a length check?
```
capture_intervals_paths: list[str] = config_retrieve(['large_cohort', 'browser_prep', 'capture_intervals'])
# Merge capture regions
capture_tables = [hl.read_table(path) for path in capture_intervals_paths]
if len(capture_intervals_paths) > 1:
    # Deduplicate keys, keeping exactly one row for each unique key (key='interval').
    capture_intervals_ht = hl.Table.union(*capture_tables)
else:
    capture_intervals_ht = capture_tables[0]
```

cpg_workflows/large_cohort/joint_frequencies.py
michael-harper: Now getting from the joint hail table's `joint_faf_meta` struct
`faf_pops = list({d.get('pop') for d in ht.joint_faf_meta.collect()[0] if d.get('pop')})`

cpg_workflows/large_cohort/joint_frequencies.py
michael-harper: I have annotated the `age_hists` inside `histograms` during the `Frequencies` stage

katiedelange (approved):
LGTM, thanks Harper!

PR #1316, opened 2025-09-04
Getting intersection of faf populations instead of union

michael-harper:
[Slack thread here](https://centrepopgen.slack.com/archives/C08JL4YHPNE/p1757022919395919?thread_ts=1756936562.919359&cid=C08JL4YHPNE). We need to get the intersection of the faf populations instead of the union as we can only perform the Cochran-Mantel-Haenszel test for populations in both datasets


katiedelange (approved):
LGTM

PR #1320, opened 2025-09-29
Code to generate VCFs for download

michael-harper:


cpg_workflows/stages/large_cohort.py
```diff
@@ -989,3 +989,67 @@ def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
...
 
         return self.make_outputs(cohort, data=self.expected_outputs(cohort), jobs=[j])
+
+
+@stage()
+class PrepareBrowserVcfDataDownload(CohortStage):
+    def expected_outputs(self, cohort: Cohort) -> dict[str, Path]:
+        if browser_vcf_version := config_retrieve(
+            ['large_cohort', 'output_versions', 'preparebrowservcfdata'],
+            default=None,
```
Alexander-Stuckey, line 999: I don't think this should be optional, right?

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+ALLELE_TYPE_FIELDS = {
+    'exomes': ALLELE_TYPE_FIELDS_FLAT,
+    'genomes': ALLELE_TYPE_FIELDS_FLAT,
+    'joint': ALLELE_TYPE_FIELDS_FLAT,
+}
+
+HISTS = ['gq_hist_alt', 'gq_hist_all', 'dp_hist_alt', 'dp_hist_all', 'ab_hist_alt']
+
+# NOTE: ASSUMING VEP_CSQ_FIELDS HAVEN'T CHANGED IN VERSION 110
+CURRENT_VEP_VERSION = '105'  # it's actually 110
```
Alexander-Stuckey, line 207: May as well update this and also the dict entry then, rather than noting it in a comment ;)

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+]
+SNP_ALLELE_SPECIFIC_FEATURES = ALLELE_SPECIFIC_FEATURES + ['AS_MQ']
+INDEL_ALLELE_SPECIFIC_FEATURES = ALLELE_SPECIFIC_FEATURES
+####
+
+INFO_DICT = {
+    'FS': {
+        'Description': 'Phred-scaled p-value of Fisher"s exact test for strand bias',
+    },
+    'InbreedingCoeff': {
```
Alexander-Stuckey, line 248: We should pick one of these two entries and remove the other, just so we don't have extra header lines in the VCF that have no corresponding entries

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+            'Log-odds ratio of being a true variant versus being a false positive under'
+            ' the trained VQSR Gaussian mixture model'
+        ),
+    },
+    'culprit': {
+        'Description': 'Worst-performing annotation in the VQSR Gaussian mixture model',
+    },
+    'decoy': {'Description': 'Variant falls within a reference decoy region'},
+    'lcr': {'Description': 'Variant falls within a low complexity region'},
+    'nonpar': {
```
Alexander-Stuckey, line 306: likewise here

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+            ' autosomal or PAR regions or 10X coverage for non-PAR regions of'
+            ' chromosomes X and Y.'
+        ),
+    },
+    'outside_ukb_capture_region': {
+        'Description': 'Variant falls outside of UK Biobank exome capture regions.',
+    },
+    'outside_broad_capture_region': {
+        'Description': 'Variant falls outside of Broad exome capture regions.',
+    },
```
Alexander-Stuckey, line 325: These ones we can probably remove, right? Do we even have this information (or care?)

Alexander-Stuckey (changes requested):
Few comments, otherwise looks good.

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+
+logging.basicConfig(
+    format='%(asctime)s (%(name)s %(lineno)s): %(message)s',
+    datefmt='%m/%d/%Y %I:%M:%S %p',
+)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+joint_included = False
```
silkm, line 27: Can this be default-valued in the functions instead?

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
silkm: I think it will need to be updated since 110 uses alpha missense right? This string should be in the intermediate vcf produced during the annotation run I think. And seconding Alex's comment.

(Out of scope here but there are some VEP v110 hard coded items in the `vep.py` script also that would be good to add to the list of things to adjust later)

cpg_workflows/large_cohort/scripts/browser_vcf_release.py
```diff
@@ -0,0 +1,2396 @@
...
+        )
+
+    ht, new_row_annots = format_validated_ht_for_export(ht, data_type=data_type)
+
+    ordered_vcf_info_dict = {f: header_dict['info'][f] for f in list(ht.info) if f in header_dict['info']}
+    header_dict.update({'info': ordered_vcf_info_dict})
+
+    hl.export_vcf(
+        ht,
+        f'{contig}_{vcf_outpath}',
```
MattWellie, line 2392: This method allows `contig` to be None, so this output path should probably account for that

Alexander-Stuckey (changes requested):
Couple more minor comments

Alexander-Stuckey (approved):
Looks good to me

PR #1321, opened 2025-10-06
subset tob to chrM

Alexander-Stuckey:
Little script to subset the tob Dragen files to chrM and write them out.


michael-harper (approved):
LGTM!

PR #1325, opened 2025-10-08
Allowing for batch resource control 

michael-harper:
Have hit memory issues when exporting the joint table to vcf:
https://batch.hail.populationgenomics.org.au/batches/677964/jobs/1
https://batch.hail.populationgenomics.org.au/batches/677965/jobs/50


Alexander-Stuckey (approved):
Looks good to me

PR #1348, opened 2026-02-03
Compressing output

michael-harper:
Streaming reheader output to `bcftools view` so that it can properly be compressed and therefore indexed


Alexander-Stuckey (approved):
Looks good to me

PR #1349, opened 2026-02-03
EGA VCF Output Extension Fix

michael-harper:
Output extensions being messy using Batch


silkm (approved):
LGTM

jmarshall:
Is EGA actually asking for `.bgz`? Because `.gz` is more correct than `.bgz`.

michael-harper:
That's a good question - I'm not sure. I was just mimicking the outputs as is. @amyminiter are you aware of any requirements by EGA re: @jmarshall 's point above?

PR #1350, opened 2026-04-13
Script to release a VCF and sample mapping

MattWellie:
Minimised version of existing data export scripts - here this is a return to provider, so we only need the files, and the mapping of internal to external sample IDs


EddieLF (approved):
LGTM but we need to enable the shared project / release bucket for any projects that will use this. I'll put a PR in to do so

